PR #104, opened 2018-11-25
Avoid separate cache for integration tests

jonhoo:
Travis uses [several factors](https://docs.travis-ci.com/user/caching/#caches-and-build-matrices) to determine whether a given job shares a cache with a different job. In particular, it uses any environment variables specified in `.travis.yml`. Previously, integration tests set `INTEGRATION=true`, which meant that the integration tests did not share a cache with the non-integration tests, even though the compilation is exactly the same.

This patch fixes that by remvoing the `INTEGRATION` environment variable and instead using the globally available [`$TRAVIS_BUILD_STAGE_NAME`](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) to run the setup required for integration tests only in stages that run integration tests (namely integration and coverage). Now, the test and integration stages share all the parameters that Travis uses to determine cache identifiers, and so they&#39;ll share their cache!

Coverage still uses its own cache because it specifically needs to *not* cache the compiled crate, but *does* need to cache `cargo-tarpaulin`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ rust:
 if: type != push OR (tag IS blank AND branch = master)
 
 before_script:
- - if [[ "$INTEGRATION" == "true" ]]; then
+ - if [[ "$TRAVIS_BUILD_STAGE_NAME" == "Integration" || "$TRAVIS_BUILD_STAGE_NAME" == "Coverage" ]]; then
      docker pull greenmail/standalone:1.5.8 &&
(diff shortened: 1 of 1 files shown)


jonhoo:
Another reason for making this change is that it significantly reduces the "cost" (in terms of cache space) of also running integration tests on `stable`+`beta`+`nightly` (as opposed to running just one of them).

jonhoo:
Comparing https://travis-ci.org/jonhoo/rust-imap/builds/459218742 and https://travis-ci.org/jonhoo/rust-imap/builds/459494989, this halved runtime for the integration test stage (no recompile), and the final cache ends up ~100MB smaller (1474.91MB -> 1346.51MB).

PR #105, opened 2018-11-25
Use fast_finish to shave time off of CI

jonhoo:
It's worth noting that this makes two relatively important changes.

First, it no longer considers test status on Windows when deciding whether the build passes or fails. This is because Travis takes forever to spin up Windows VMs, and it holds up CI for a good ten minutes.

Second, it no longer considers coverage failing to run a failure of the build. This is because coverage just re-runs the test suite, which CI has already checked passes, but still takes ~16m without a warm cache (like for PRs). Interestingly enough, coverage was *already* ignored since it is using `rust: nightly`, but this change makes that more permanent by more explicitly matching against coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ rust:
   - stable
   - beta
   - nightly
+# We *must* specify these for allow_failures to work on the os and env keys
+# https://docs.travis-ci.com/user/customizing-the-build#rows-that-are-allowed-to-fail
+os: linux
+env:
 
 # always test things that aren't pushes (like PRs)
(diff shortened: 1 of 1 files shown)


jonhoo:
Note that this probably doesn't fully work as well as we want to yet because of https://github.com/travis-ci/travis-ci/issues/9677.

jonhoo:
This basically doesn't seem to work at the moment, so I'll hold off on merging. See https://github.com/travis-ci/travis-ci/issues/9677#issuecomment-441476502.

jonhoo:
This'll be obviated by the move to Azure Pipelines.

Issue #106, opened 2019-01-05
Err(Parse(Invalid(...)) ) on most methods to select/examine/fetch from gmail 


softprops:
First off, thank you for your work on this crate.

I'm having trouble getting almost all session methods to work with authenticating with an app password with gmail

```rust
let domain = "...";
let tls = TlsConnector::builder().build().unwrap();

let client = imap::connect((domain, 993), domain, &tls)
            .expect(&format!("failed to connect to {}", domain));

let mut session = client.login(username, app_password).expect("failed to authenticate with domain");

println!("{:#?}", session. select("INBOX").expect("failed to select inbox"));
```

In this example and many others I'll get `Parse(Invalid(...))` errors. When I view the bytes as a string in this case I see

```
 "* FLAGS (\\Answered \\Flagged \\Draft \\Deleted \\Seen $NotPhishing $Phishing OIB-Seen-INBOX OIB-Seen-OIB/Business OIB-Seen-OIB/Entertainment OIB-Seen-OIB/Finance OIB-Seen-OIB/Groups OIB-Seen-OIB/Home OIB-Seen-OIB/Jobs OIB-Seen-OIB/News OIB-Seen-OIB/Shopping OIB-Seen-OIB/Social Networking OIB-Seen-OIB/Travel OIB-Seen-[Gmail]/All Mail OIB-Seen-[Gmail]/Important OIB-Seen-[Gmail]/Spam OIB-Seen-[Gmail]/Trash)\r\n"
```

I'm not familiar with imap at the protocol level but am I missing something here?

jonhoo:
Hmm, that's very interesting... Thanks for reporting it!
Could you please post the full error + backtrace that you get?

dario23:
as far as i can see, imap-proto chokes on the ']' chars (in [its parser](https://github.com/djc/tokio-imap/blob/master/imap-proto/src/parser.rs#L145), `flag_char` references `atom` which is [everything that is not an atom special](https://github.com/djc/tokio-imap/blob/master/imap-proto/src/parser.rs#L34) and those include `b']'` in `resp_specials`.

not sure what the best path forward is, though i'd think towards having the parser be more liberal here.

jonhoo:
Ah, good catch! Let's see what @djc has to say here ^

softprops:
[...]

PR #107, opened 2019-01-05
add gmail app password example

softprops:
this adds a new example to help illustrate a reproducible case for https://github.com/jonhoo/rust-imap/issues/106 and an alternative to oauth2 tokens for users of this crate in the future

### DIFF
--- a/examples/gmail_app_password.rs
+++ b/examples/gmail_app_password.rs
@@ -1,0 +1,38 @@
+/// Example of using gmail authenticating using
+/// an [App Password](https://support.google.com/accounts/answer/185833)
+extern crate imap;
+extern crate native_tls;
+
+use native_tls::TlsConnector;
+
+fn main() {
+    let domain = "imap.gmail.com";
+    let port = 993;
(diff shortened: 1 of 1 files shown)


jonhoo:
Actually, on second thought, this is _super_ similar to the `basic.rs` example. I think maybe we just want to add a comment about app passwords in `basic.rs`. Maybe even rename it to `gmail.rs` + change the server address so it works "out of the box"?

jonhoo:
@softprops what do you think of my comment above?

softprops:
Im going to go ahead and close this, I was able to unblock myself to testing with another gmail account. feel free to close this

jonhoo:
Oh, that's too bad. What problem did you run into?

softprops:
This problem https://github.com/jonhoo/rust-imap/issues/106, something was up with my google account that caused this library to be unable to parse the imap responses. I timeboxed investigating that then ended up just using another gmail account.

jonhoo:
Ohh, I remember now. That's too bad — your updated example would be a good addition :)

PR #110, opened 2019-03-16
Handle unsolicited responses in noop.

Emm54321:
NOOP should handle unsolicited responses.


jonhoo:
Excellent, thank you!

PR #113, opened 2019-04-10
Add Fetch.internal_date method

### LINKED ISSUE #112: Add access to InternalDate in Fetch struct
I need to have access to InternalDate from the Fetch struct. Did you consider adding this feature or was it left apart intentionally ?

Thanks

AMDG2:
This method gives the internal date from the FETCH answer.

Fixes #112

### DIFF
--- a/src/types/fetch.rs
+++ b/src/types/fetch.rs
@@ -2,6 +2,9 @@ use super::{Flag, Seq, Uid};
 use imap_proto::types::{AttributeValue, Envelope, MessageSection, SectionPath};
 use chrono::{DateTime, FixedOffset};
 
+/// Format of Date and Time as defined RFC3501.
+/// See `date-time` element in [Formal Syntax](https://tools.ietf.org/html/rfc3501#section-9)
+/// chapter of this RFC.
 const DATE_TIME_FORMAT: &str = "%d-%b-%Y %H:%M:%S %z";
 
 /// An IMAP [`FETCH` response](https://tools.ietf.org/html/rfc3501#section-7.4.2) that contains
@@ -131,7 +134,7 @@ impl Fetch {
         self.fetch
             .iter()
             .filter_map(|av| match av {
-                AttributeValue::InternalDate(str) => Some(*str),
+                AttributeValue::InternalDate(date_time) => Some(*date_time),
                 _ => None,
             })
             .next()
--- a/src/types/fetch.rs
+++ b/src/types/fetch.rs
@@ -1,6 +1,6 @@
 use super::{Flag, Seq, Uid};
-use imap_proto::types::{AttributeValue, Envelope, MessageSection, SectionPath};
 use chrono::{DateTime, FixedOffset};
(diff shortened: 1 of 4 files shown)


jonhoo:
Ah, this is a great addition! I wonder if we should also parse the resulting date though? What do you think?

AMDG2:
Well, I implemented this features and 10 minutes after I was adding chrono as a dependency 😅 

I think it make sense, but do you want to parse it on your own, or to use chrono or another dependency ?

jonhoo:
I think adding `chrono` as a dependency makes a lot of sense. We might even want to then take advantage of that elsewhere too.

AMDG2:
Ok, I will do this then. Does it make sense to also have `internal_date_str()` to get the raw value ?

jonhoo:
Hmm, that's a good question. I think I'd favor just having the non-string version.

src/types/fetch.rs
```diff
@@ -119,4 +122,24 @@ impl Fetch {
...
+
+    /// Extract the `INTERNALDATE` of a `FETCH` response
+    ///
+    /// See [section 2.3.3 of RFC 3501](https://tools.ietf.org/html/rfc3501#section-2.3.3) for
+    /// details.
+    pub fn internal_date(&self) -> Option<DateTime<FixedOffset>> {
+        self.fetch
+            .iter()
+            .filter_map(|av| match av {
+                AttributeValue::InternalDate(str) => Some(*str),
```
jonhoo: I'm surprised that `str` is a valid variable identifier here since it's also a type. Maybe make it `s` or `date` instead so the syntax highlighter doesn't remain confused? Minor point though :)

src/types/fetch.rs
```diff
@@ -1,5 +1,8 @@
 use super::{Flag, Seq, Uid};
 use imap_proto::types::{AttributeValue, Envelope, MessageSection, SectionPath};
+use chrono::{DateTime, FixedOffset};
+
+const DATE_TIME_FORMAT: &str = "%d-%b-%Y %H:%M:%S %z";
```
jonhoo: Could you leave a comment pointing to where this format is standardized?

src/types/fetch.rs
AMDG2: Well I remember that I just copy/pasted from your code, but I don't find the same in yours. I will fix it.

src/types/fetch.rs
AMDG2: It was planned, but I was about to forget it !

AMDG2:
I integrated your suggestions, now I just need to write some tests, give me about one month again. I should find the time to do it !

AMDG2:
I added some tests, codecov should catch up once Travis is done. If the final coverage and tests are fine for you, you can merge. I can also rewrite the PR history to make it clean if you want.

jonhoo:
Excellent! I'll keep an eye on it. No need to rebase -- I'll squash for merging :)

PR #114, opened 2019-04-14
Implemented iteration over unsolicited responses during IDLE.

Emm54321:
The Handle returned by the idle() method now has iter() / iter_keepalive() / iter_timeout() methods that allow iterating over unsolicited responses until an errors occurs. The error is not reported to the user though. Not sure what would be a good way to do that.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/114)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,5 +1,6 @@
 use base64;
 use bufstream::BufStream;
+use enumset::EnumSet;
 use native_tls::{TlsConnector, TlsStream};
 use nom;
 use std::collections::HashSet;
@@ -13,6 +14,7 @@ use super::error::{Error, ParseError, Result, ValidateError};
 use super::extensions;
 use super::parse::*;
 use super::types::*;
+use super::unsolicited_responses::UnsolicitedResponseSender;
 
 static TAG_PREFIX: &'static str = "a";
 const INITIAL_TAG: u32 = 0;
@@ -48,7 +50,7 @@ fn validate_str(value: &str) -> Result<String> {
 #[derive(Debug)]
 pub struct Session<T: Read + Write> {
     conn: Connection<T>,
-    unsolicited_responses_tx: mpsc::Sender<UnsolicitedResponse>,
+    unsolicited_responses_tx: UnsolicitedResponseSender,
 
     /// Server responses that are not related to the current command. See also the note on
     /// [unilateral server responses in RFC 3501](https://tools.ietf.org/html/rfc3501#section-7).
@@ -376,8 +378,21 @@ impl<T: Read + Write> Session<T> {
         Session {
             conn,
             unsolicited_responses: rx,
-            unsolicited_responses_tx: tx,
+            unsolicited_responses_tx: UnsolicitedResponseSender::new(tx),
         }
+    }
+
+    /// Tells which unsolicited responses are required. Defaults to none.
+    ///
+    /// The server *is* allowed to unilaterally send things to the client for messages in
+    /// a selected mailbox whose status has changed. See the note on [unilateral server responses
+    /// in RFC 3501](https://tools.ietf.org/html/rfc3501#section-7). This function tells
+    /// which events you want to hear about.
+    ///
+    /// If you request unsolicited responses, you have to regularly check the
+    /// `unsolicited_responses` channel of the [`Session`](struct.Session.html) for new responses.
+    pub fn request_unsolicited_responses(&mut self, mask: EnumSet<UnsolicitedResponseCategory>) {
+        self.unsolicited_responses_tx.request(&self.unsolicited_responses, mask);
     }
 
     /// Selects a mailbox
--- a/src/client.rs
+++ b/src/client.rs
@@ -995,7 +995,8 @@ impl<T: Read + Write> Session<T> {
     ///
     /// See [`extensions::idle::Handle`] for details.
     pub fn idle(&mut self) -> Result<extensions::idle::Handle<T>> {
-        extensions::idle::Handle::make(self)
+        let sender = self.unsolicited_responses_tx.clone();
+        extensions::idle::Handle::make(self, sender)
     }
 
     /// The [`APPEND` command](https://tools.ietf.org/html/rfc3501#section-6.3.11) appends
--- a/src/extensions/idle.rs
+++ b/src/extensions/idle.rs
@@ -3,12 +3,12 @@
 
 use client::Session;
 use error::{Error, Result};
+use fallible_iterator::FallibleIterator;
 use native_tls::TlsStream;
 use std::io::{self, Read, Write};
 use std::net::TcpStream;
-use std::ops::{Deref, DerefMut};
 use std::sync::mpsc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use types::UnsolicitedResponse;
 use parse;
 
@@ -36,6 +36,7 @@ trait OnDrop<'a> {
 pub struct Handle<'a, T: Read + Write + 'a> {
     session: &'a mut Session<T>,
     unsolicited_responses_tx: mpsc::Sender<UnsolicitedResponse>,
+    start_time: Instant,
     keepalive: Duration,
     done: bool,
 }
@@ -47,12 +48,11 @@ pub struct Handle<'a, T: Read + Write + 'a> {
 pub struct IdleIterator<'a, T: Read + Write + 'a> {
     handle: Handle<'a, T>,
     buffer: Vec<u8>,
-    keepalive: bool,
 }
 
 impl<'a, T: Read + Write + 'a> IdleIterator<'a, T> {
-    fn new(handle: Handle<'a, T>, keepalive: bool) -> IdleIterator<'a, T> {
-        IdleIterator { handle, buffer: Vec::new(), keepalive }
+    fn new(handle: Handle<'a, T>) -> IdleIterator<'a, T> {
+        IdleIterator { handle, buffer: Vec::new() }
     }
 }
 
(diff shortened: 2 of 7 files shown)


jonhoo:
Oooh, this is awesome! I've wanted this for a while. Let me do a read-through and leave any notes I may have.

src/extensions/idle.rs
```diff
@@ -26,10 +35,50 @@ use std::time::Duration;
...
+impl<'a, T: Read + Write + 'a> IdleIterator<'a, T> {
+    fn new(handle: Handle<'a, T>, keepalive: bool) -> IdleIterator<'a, T> {
+        IdleIterator { handle, buffer: Vec::new(), keepalive }
+    }
+}
+
+/// 'TimeoutIdleIterator' allows a client to iterate over unsolicited responses during an IDLE operation.
+///
+/// As long as a [`TimeoutIdleIterator`] is active, the mailbox cannot be otherwise accessed.
+pub struct TimeoutIdleIterator<'a, T: SetReadTimeout + Read + Write + 'a> {
```
jonhoo: Why do we need a separate type for this?

src/client.rs
```diff
@@ -995,7 +995,8 @@ impl<T: Read + Write> Session<T> {
     ///
     /// See [`extensions::idle::Handle`] for details.
     pub fn idle(&mut self) -> Result<extensions::idle::Handle<T>> {
-        extensions::idle::Handle::make(self)
+        let sender = self.unsolicited_responses_tx.clone();
```
jonhoo: Re-using the `unsolicited_responses_tx` for this means that you'll also get any unsolicited messages from _before_ the IDLE command was issued. Do you think that may be problematic?

src/parse.rs
```diff
@@ -299,6 +299,36 @@ pub fn parse_ids(
...
 
+pub fn parse_idle(
+    mut lines: &[u8],
+    unsolicited: &mut mpsc::Sender<UnsolicitedResponse>,
+) -> Result<()> {
+    while !lines.is_empty() {
+        match imap_proto::parse_response(lines) {
+            Ok((rest, data)) => {
+                lines = rest;
+                if let Some(resp) = handle_unilateral(data, unsolicited) {
```
jonhoo: Are unilateral responses the only thing that can come from IDLE?

jonhoo:
I left some notes in-line, but overall this looks pretty good! I'm not sure what to do with errors though. Maybe [`FallibleIterator`](https://docs.rs/fallible-iterator/0.2.1/fallible_iterator/trait.FallibleIterator.html) is worth looking into?

src/extensions/idle.rs
Emm54321: Because the if a timeout is used, dropping the Iterator resets the timeout and this can't be done in IdleIterator if T isn't SetReadTimeout. So that requires different Drop implementations depending if T's traits. Or maybe it could be done in IdleIterator by registering a callback, but I'm not sure it's cleaner.

jonhoo: Ah, yes, that makes sense... Fair enough!

src/client.rs
Emm54321: Well, that way unsolicited responses that are sent between the previous command and the idle command aren't lost. And with the current implementation, the user should empty the unsolicited response queue after each command or it will grow indefinitely. Or maybe the library should clear the queue at the start of each command if the user has not done so.

jonhoo: I'm actually not sure what's best here. Forcing a clearing of the queue at every command seems like something the user might not be expecting. Maybe it'd make sense to purge it before you enter IDLE mode though?

src/parse.rs
Emm54321: IDLE can only send untagged responses, but there can be some that are not currently handled by the UnsolicitedResponse enum, like FETCH responses.

src/client.rs
Emm54321: Maybe. But currently if the user does nothing about unsolicited responses, the queue will keep growing forever.

Emm54321:
FallibleIterator is probably better indeed. I didn't know about it.

Emm54321:
Ok, I made a better implementation with FallibleIterators, and I fixed the timeout handling. I still don't empty the unsolicited response queue though. Maybe it's better to just add in the doc for idle() that if the user doesn't want old unhandled responses, he has to clear the queue before calling idle().

src/parse.rs
jonhoo: We probably want to handle these here too then, though in that case I wonder if we need a different channel since the types we send would be different?

src/client.rs
jonhoo: Hmm, yeah, I honestly don't know what to do here. Maybe we make the channel a bounded channel instead, and drop unsolicited responses if it's full? And then clear before entering IDLE.

src/parse.rs
Emm54321: I think a Fetch item should just be added to the UnsolicitedResponse enum.

jonhoo: Maybe, although that also seems weird if you can't generally receive unsolicitied `FETCH` response (are there other responses IDLE adds too?).

Emm54321: You can, see 7.4.2 of rfc 350: This response occurs as the result of a FETCH or
      STORE command, as well as by unilateral server decision (e.g.,
      flag updates).
There can also be OK/NO/BAD responses. E.g. Dovecot sends 'OK Still here' periodically, I currently filter OK out in parse_idle, but it could be returned to the user too. They may be sent to other commands too.

jonhoo: That sounds to me like we should really make the IDLE channel totally separate, and have it send only the `FETCH` and `STORE` responses. Unilateral responses would still go to the original unilateral channel. Maybe we make them both `crossbeam-channel` channels so that the user can select over both of them at the same time?

Emm54321: I don't think so. A FETCH response can really occur at any time just like an EXISTS, EXPUNGE or RECENT response. You'll get one with any command sent after a message has been marked as read for instance. You really want all these responses when you use idle(). Besides, if you use separate channels, you lose the chronology of those responses and thus can't maintain the id<->message association.

jonhoo: Sounds good to me! I guess we then need to update `handle_unilateral` to also handle those responses :)

Emm54321: Right, I'll give it a try tomorrow.

src/parse.rs
```diff
@@ -353,6 +345,53 @@ fn handle_unilateral<'a>(
...
         }
+        Response::Data { status: Status::Ok, code, information } => {
+            unsolicited
+                .send(UnsolicitedResponse::Ok {
+                    code: code.map(|c| c.into()),
+                    information: information.map(|s| s.to_string())
+                })
+                .unwrap();
+        }
+        Response::Data { status: Status::Bad, code, information } => {
```
jonhoo: Can these (Bad and No) really come as unsolicited messages?

src/types/mod.rs
```diff
@@ -204,6 +204,60 @@ pub use self::capabilities::Capabilities;
...
+
+// We need a ResponseCode that is not tied to a lifetime, to be used in UnsolicitedResponse.
+/// Response code that may be sent with OK/NO/BAD/BYE responses.
+/// See [RFC 3501](https://tools.ietf.org/html/rfc3501#section-3.1).
+#[derive(Debug, Eq, PartialEq)]
+pub enum ResponseCode {
+    //Alert: not parsed by imap-proto yet.
+    //BadCharset: not parsed by imap-proto yet.
+    //Capability: not parsed by imap-proto yet.
+    //Parse: not parsed by imap-proto yet.
```
jonhoo: Would be good to link to an imap-proto issue for this.

src/extensions/idle.rs
```diff
@@ -226,13 +226,9 @@ impl<'a, T: Read + Write + 'a> FallibleIterator for IdleIterator<'a, T> {
             if let Ok(u) = self.handle.session.unsolicited_responses.try_recv() {
                 return Ok(Some(u));
             }
-
+            self.buffer.clear();
```
jonhoo: This doesn't seem quite right? What if the first `readline` only gives you a partial response and you need to read more to get and parse the full response?

Emm54321: parse_idle() parses until the buffer is empty, so it shouldn't happen.

src/types/mod.rs
Emm54321: Yes, I'm working on a pull request for that.

src/extensions/idle.rs
jonhoo: The issue is if you get _half_ of a valid response, and you need to read in the other half before parsing can get the next item out.

Emm54321: Oh, right, there's a problem here.

src/client.rs
Emm54321: What about this:
- we add a request_unsolicited(category_flags) method to Session
- handle_unsolicited() enqueues only responses that are in the requested categories
- the category flags could default to be empty, so that users not explicitly asking for unsolicited responses wont waste memory with an infinitely growing queue, and we can assume that the user does handle responses he asks for.
- document the idle iterator as returning the not yet handled responses first, if any.
Alternatively, the default category_flags could be STATUS|RECENT|EXISTS|EXPUNGE if you prefer compatibility with the current behavior.

jonhoo: I like that plan! "Muting" all by default seems reasonable. I think finding a good name for the "unmasking" method is important, but beyond that this seems like a good strategy.

src/extensions/idle.rs
```diff
@@ -35,14 +35,17 @@ trait OnDrop<'a> {
...
-    unsolicited_responses_tx: mpsc::Sender<UnsolicitedResponse>,
+    unsolicited_responses_tx: UnsolicitedResponseSender,
     start_time: Instant,
     keepalive: Duration,
     done: bool,
 }
 
 
 /// 'IdleIterator' allows a client to iterate over unsolicited responses during an IDLE operation.
+/// Only the unsolicited responses requested by the [`Session::request`] method will be
```
jonhoo: I think you need to name the full method here?

Emm54321: Oops, indeed.

src/extensions/idle.rs
```diff
@@ -119,6 +164,30 @@ impl<'a, T: SetReadTimeout + Read + Write + 'a> Handle<'a, T> {
         self.keepalive = interval;
     }
 
+    /// Returns an iterator over unsolicited responses.
+    ///
+    /// Only the unsolicited responses requested by the [`Session::request`] method will be
```
jonhoo: Same here

jonhoo:
I'm going to close this in favor of #130.

PR #115, opened 2019-04-15
Idle fallible iterator

Emm54321:
- Implemented the IDLE iterator as a FallibleIterator, with proper error handling.
- Fixed timeout handling : the timeout should be handled as starting at the IDLE command, not restarted after each call to next(). This means that IdleIterator and TimeoutIdleIterator must have different next() implementations as the latter has to call set_read_timeout().


jonhoo:
Oh, just push this to #114 instead and we'll do all discussion there :)

PR #119, opened 2019-04-28
Fix quoting issue in LIST command

### LINKED ISSUE #118: Cannot list all names
Hey,

I'm having troubles issueing the following command:

```
LIST "" *
```

I assume it is because of `quote!()` here:

https://github.com/jonhoo/rust-imap/blob/ba09886e3b12867926093faf61711351ca5d2a59/src/client.rs#L898

I don't understand why this works because the RFC doesn't seem to mention it. But the above command is used successfully in other libraries to list all names. Even names above `INBOX`, such as draft, trash and so on.

See for example cpython:

https://github.com/python/cpython/blob/74125a60b7a477451ff2b8385bfbce3fdaee8dbc/Lib/imaplib.py#L580

This patch fixes the issue for me:

```diff
diff --git a/src/client.rs b/src/client.rs
index 0df57cb..0953f6d 100644
--- a/src/client.rs
+++ b/src/client.rs
@@ -895,7 +895,7 @@ impl<T: Read + Write> Session<T> {
     ) -> ZeroCopyResult<Vec<Name>> {
         self.run_command_and_read_response(&format!(
             "LIST {} {}",
-            quote!(reference_name.unwrap_or("\"\"")),
+            reference_name.unwrap_or("\"\""),
             mailbox_pattern.unwrap_or("\"\"")
         ))
         .and_then(|lines| parse_names(lines, &mut self.unsolicited_responses_tx))
```

I can then use the following code to fetch the list of all names:

```
map_session.list(None, Some("*"))
```

NickeZ:
Fixes #118

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -895,7 +895,7 @@ impl<T: Read + Write> Session<T> {
     ) -> ZeroCopyResult<Vec<Name>> {
         self.run_command_and_read_response(&format!(
             "LIST {} {}",
-            quote!(reference_name.unwrap_or("\"\"")),
+            quote!(reference_name.unwrap_or("")),
             mailbox_pattern.unwrap_or("\"\"")
         ))
         .and_then(|lines| parse_names(lines, &mut self.unsolicited_responses_tx))


jonhoo:
Released in 1.0.1 :rocket:

PR #120, opened 2019-05-03
Implemented new enumerated Capabilities from imap-proto crate.

lberezy:
This PR exposes the new capability of the underlying imap-proto crate (https://github.com/djc/tokio-imap/pull/47) to parse Capabilities into an enumerated type.

Note that this is not yet ready for merge as it depends on the master of imap-proto, not a currently released version.

### DIFF
--- a/src/types/capabilities.rs
+++ b/src/types/capabilities.rs
@@ -1,7 +1,9 @@
-use std::borrow::Borrow;
+use imap_proto::types::Capability;
 use std::collections::hash_set::Iter;
 use std::collections::HashSet;
-use std::hash::Hash;
+
+const IMAP4REV1_CAPABILITY: &str = "IMAP4rev1";
+const AUTH_CAPABILITY_PREFIX: &str = "AUTH=";
 
 /// From [section 7.2.1 of RFC 3501](https://tools.ietf.org/html/rfc3501#section-7.2.1).
 ///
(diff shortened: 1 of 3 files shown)


src/types/capabilities.rs
```diff
@@ -31,21 +33,28 @@ use std::hash::Hash;
...
 );
 
 impl Capabilities {
     /// Check if the server has the given capability.
-    pub fn has<S: ?Sized>(&self, s: &S) -> bool
-    where
-        for<'a> &'a str: Borrow<S>,
-        S: Hash + Eq,
-    {
+    pub fn has<'a>(&self, s: &Capability<'a>) -> bool {
```
jonhoo: This is a backwards incompatible change, which is a bit unfortunate.. Maybe keep `has` with the old signature and have it call `has_str` + add a deprecation notice to use `has_str` instead? And call the method that takes a `Capability` `contains`?

src/types/capabilities.rs
```diff
@@ -31,21 +33,28 @@ use std::hash::Hash;
...
         self.0.contains(s)
     }
 
+    /// Check if the server has the given capability via str.
+    pub fn has_str(&self, s: &str) -> bool {
```
jonhoo: I'd prefer for this to be generic over the type of string the same way the old one was.

src/types/capabilities.rs
```diff
@@ -31,21 +33,28 @@ use std::hash::Hash;
...
+        } else if s.starts_with(AUTH_CAPABILITY_PREFIX) {
+            self.has(&Capability::Auth(&s[AUTH_CAPABILITY_PREFIX.len()..]))
+        } else {
+            self.has(&Capability::Atom(s))
+        }
+    }
+
     /// Iterate over all the server's capabilities
-    pub fn iter(&self) -> Iter<&str> {
+    pub fn iter(&self) -> Iter<Capability> {
```
jonhoo: Hmm, this is also backwards incompatible.. I would pull the same trick as above here. Add `iter_str`, deprecate the old `iter` (but with the same signature) pointing people to `iter_str`, and then add the iterator over `Capability` under the name `all`? With the next major version bump we'll promote `contains` to `has` and `all` to `iter` (unless we prefer the new names?).

jonhoo:
Thank you! It's a little unfortunate that this is a breaking change as mentioned in the comments, but apart from that it looks pretty good!

link2xt:
> Note that this is not yet ready for merge as it depends on the master of imap-proto, not a currently released version.

`imap-proto` version 0.8.1 has been released recently, see https://github.com/djc/tokio-imap/issues/52

jonhoo:
A gentle ping on this @lberezy.

avitex:
@jonhoo i'll have look at this tomorrow :)

avitex:
@jonhoo There are three solutions I can see going forward.

As the upstream Capability type cuts off the `AUTH=` prefix, so dealing with just `&str` becomes a little more troublesome.

1. We create an owned Capability (requires copy of data as `&str` is contiguous) adding the prefix.
2. The upstream crate is changed to not cut the prefix (a breaking change).
3. This crate makes a breaking change.

All three are not desirable, that I understand. You've put good effort in making this crate with zero copy guarantees. I would make a wager you've already identified this dilemma, and would prefer upstream to change :stuck_out_tongue:

I'm happy to implement whichever case it be, but what in your eyes is the lesser evil?
I personally prefer a breaking change on this crate.

lberezy:
It might be painful, but I think I agree that [picking number 3](https://www.youtube.com/watch?v=9e-ZM-NXKOU) might be the best option.

jonhoo:
Hmm, yeah, good points. Okay, I guess it's time to do a breaking change. It's probably good to have this coincide with #130, which is not _quite_ a breaking change, but not far from it.

jonhoo:
I think the only thing missing here then is the one outstanding comment about making `has_str` generic over the string type the same way the old one was.

jonhoo:
Replaced by #133.

Issue #123, opened 2019-05-06
SSL/TLS dependency


lberezy:
In cases where the crate is being used only for insecure connections to IMAP server, there is no need to build with SSL/TLS support.

I think the correct way to handle this would be to put all SSL/TLS related features behind a "ssl" compile-time feature. This feature would be included in the default features so that forgoing TLS usage is opt-out. This is how other libraries dealing with this typically implement this.

I've got a PR in the works that does this.

jonhoo:
I'm somewhat disinclined to make SSL/TLS support optional, because I think it should be very rare these days that SSL/TLS is not what you want. Of course, there may be smaller exceptions, like if you already have a secured channel, but those are very niche. What is the use-case you have in mind for this?

lberezy:
I agree that TLS should be more than strongly recommended, however I think it is still important to give people the option not to pay for something they do not need, especially if this has no impact for the general case (feature on by default).

My scenario is as your said, one that is protected by other transport level protections. The other libraries I am using in my application all have opt-out SSL, but my final binary still depends on a bunch of SSL here. This complicates deployment dependencies and ease of cross-compilation (some SSL crates have special build requirements, like requiring a C compiler be installed).

brycefisher:
I'm running into the same issue here with cross compilation specifically.

I would propose supporting a pure Rust TLS crate in addition to native_tls as my preferred option to (1) keep TLS support and (2) ease cross complication significantly.

jonhoo:
@brycefisher I'd be happy to review a PR adding support for a pure Rust TLS crate!

brycefisher:
Ok! I'll see if I can put one together. I'm experimenting with rustls on raspberrypi now. If that works, I'll open a PR here.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

brycefisher:
So, without looking too deeply into the source of this crate, it appears that the openssl-sys crate is mandatory requirement. If I'm able to get `rustls` working, both myself and @lberezy would want to opt out of openssl-sys. I would probably do that by creating a cargo feature for openssl that is enabled by default, and then users wanting to cross-compile and rely on rustls would simply disable all features.

The main wrinkle I see with this plan is that the easiest way I can see to do this would allow users to opt out of TLS support altogether by simply using `connect_insecure` and disabling the default features. I only mention this because @jonhoo specifically stated a desire not to provide a means of disabling TLS completely.

[...]

PR #124, opened 2019-05-06
Placed SSL/TLS dependencies behind a default feature flag

lberezy:
#123 
This allows usage of the crate without a dependency on SSL/TLS.

By default, the "ssl" feature is included in the dependencies as to not cause any breaking changes.

Note that I have not run any of the functional tests in the test suite.

### DIFF
--- a/examples/gmail_oauth2.rs
+++ b/examples/gmail_oauth2.rs
@@ -1,7 +1,8 @@
 extern crate base64;
 extern crate imap;
+#[cfg(feature = "ssl")]
 extern crate native_tls;
-
+#[cfg(feature = "ssl")]
 use native_tls::TlsConnector;
 
 struct GmailOAuth2 {
@@ -20,6 +21,7 @@ impl imap::Authenticator for GmailOAuth2 {
     }
 }
 
+#[cfg(feature = "ssl")]
 fn main() {
     let gmail_auth = GmailOAuth2 {
         user: String::from("sombody@gmail.com"),
--- a/src/error.rs
+++ b/src/error.rs
@@ -10,7 +10,9 @@ use std::string::FromUtf8Error;
 use base64::DecodeError;
 use bufstream::IntoInnerError as BufError;
 use imap_proto::Response;
+#[cfg(feature = "ssl")]
 use native_tls::Error as TlsError;
+#[cfg(feature = "ssl")]
 use native_tls::HandshakeError as TlsHandshakeError;
 
 /// A convenience wrapper around `Result` for `imap::Error`.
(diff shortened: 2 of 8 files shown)


src/error.rs
```diff
@@ -22,8 +24,10 @@ pub enum Error {
     /// An `io::Error` that occurred while trying to read or write to a network stream.
     Io(IoError),
     /// An error from the `native_tls` library during the TLS handshake.
+    #[cfg(feature = "ssl")]
```
jonhoo: This (and some of the other changes) would make the `ssl` feature non-additive, which [causes problems](https://github.com/Geal/nom/issues/544). I'm not entirely sure that there's a way to work around this :/

examples/gmail_oauth2.rs
```diff
@@ -20,6 +21,7 @@ impl imap::Authenticator for GmailOAuth2 {
     }
 }
 
+#[cfg(feature = "ssl")]
```
jonhoo: It's probably better here to list the example in `Cargo.toml` with `required-features = ["ssl"]` (assuming that works).

tests/imap_integration.rs
```diff
@@ -1,18 +1,21 @@
 extern crate imap;
 extern crate lettre;
 extern crate lettre_email;
+#[cfg(feature = "ssl")]
```
jonhoo: I think it's probably fine for `native-tls` to always be included in `dev-dependencies`

examples/gmail_oauth2.rs
lberezy: I didn't know about `required-features`. Seems like that's a much better solution, thanks.

src/error.rs
lberezy: Yes this is a tricky one to solve cleanly, especially without any breaking changes (split the enum into subset/superset relationship).
Perhaps it is appropriate though to allow this as the current API of the crate makes it a conscious decision to create a secure or insecure session - it would not be possible anyway to switch from not having the ssl feature to having it without also having to adapt other parts of a user's code.

tests/imap_integration.rs
lberezy: Good pickup, that makes sense.

lberezy:
I'm also considering renaming the feature flag to "tls" as I think that's more appropriate in 2019.

src/error.rs
jonhoo: Unfortunately crate authors do not have the option of choosing. Since `cargo` assumes that features are additive, it will automatically pick the maximal set of features for any dependency. For example if `a` and `b` both depended on `imap`, one with the feature and one without, cargo would only compile `imap` once _with_ the feature and then compile both `a` and `b` against that, at which point `b` would fail to compile if it ever tried to match over the errors for example. That's far from ideal..

jonhoo:
Implemented in #140.

Issue #128, opened 2019-07-29
parse error in `noop` loop from "seen" event from server


norcalli:
Pretty easy to reproduce:

```rust
    loop {
        println!("tick");
        if let Ok(response) = imap_session.unsolicited_responses.try_recv() {
            println!("response: {:?}", response);
            match response {
                UnsolicitedResponse::Recent(id) | UnsolicitedResponse::Exists(id) => {
                    print_email(&mut imap_session, id)?;
                }
                other => {
                    println!("Other {:?}", other);
                }
            }
        }
        std::thread::sleep(std::time::Duration::from_secs(1));
        imap_session.noop()?;
    }
```

While in this loop, read an email and notice that the process exits with the following error:

`Error: Imap(Parse(Unexpected("Fetch(31724, [Uid(85608), Flags([\"\\\\Seen\"])])")))`

It seems like this is a bug in not adding it to the unsolicited events. This is with a Gmail server.

jonhoo:
Ah, interesting. I believe this is fixed by #114 by @Emm54321, specifically the changes to `handle_unilateral` around [here](https://github.com/jonhoo/rust-imap/pull/114/files#diff-258c0555e6f9c4bc4fba3a551d81638aR325). That adds a part that handles fetch replies as unilateral responses, which I believe should fix your issue. That PR isn't too far from landing I think, though if you are in a rush to get this fixed, you could just lift the changes related to unsolicited responses from that PR and file it as a separate PR :)

norcalli:
I'm also interested in the results of that PR since that is the smarter alternative to a noop loop when the server has IDLE. I believe that if I did take that part of the PR out, then I could use IDLE and still iterate over the unsolicited responses after IDLE ends, right? In that case, then I think that the iterator is merely a nice facade over a loop + IDLE + try_recv(). I'll try to implement that PR, then.

jonhoo:
Yes indeed!

norcalli:
PR up. Testing locally looks good so far.

PR #130, opened 2019-08-04
Implement unsolicited message handling while IDLEing

norcalli:
Adapted from https://github.com/jonhoo/rust-imap/pull/114

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/130)

### DIFF
--- a/src/extensions/idle.rs
+++ b/src/extensions/idle.rs
@@ -4,9 +4,11 @@
 use client::Session;
 use error::{Error, Result};
 use native_tls::TlsStream;
+use parse;
 use std::io::{self, Read, Write};
 use std::net::TcpStream;
 use std::time::Duration;
+use unsolicited_responses::UnsolicitedResponseSender;
 
 /// `Handle` allows a client to block waiting for changes to the remote mailbox.
 ///
@@ -26,7 +28,8 @@ use std::time::Duration;
 #[derive(Debug)]
 pub struct Handle<'a, T: Read + Write + 'a> {
     session: &'a mut Session<T>,
-    keepalive: Duration,
+    unsolicited_responses_tx: UnsolicitedResponseSender,
+    keepalive: Option<Duration>,
     done: bool,
 }
 
@@ -45,10 +48,14 @@ pub trait SetReadTimeout {
 }
 
 impl<'a, T: Read + Write + 'a> Handle<'a, T> {
-    pub(crate) fn make(session: &'a mut Session<T>) -> Result<Self> {
+    pub(crate) fn make(
+        session: &'a mut Session<T>,
+        unsolicited_responses_tx: UnsolicitedResponseSender,
+    ) -> Result<Self> {
         let mut h = Handle {
             session,
-            keepalive: Duration::from_secs(29 * 60),
+            unsolicited_responses_tx,
+            keepalive: None,
             done: false,
         };
         h.init()?;
@@ -91,17 +98,29 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
     ///
     /// This is necessary so that we can keep using the inner `Session` in `wait_keepalive`.
     fn wait_inner(&mut self) -> Result<()> {
-        let mut v = Vec::new();
-        match self.session.readline(&mut v).map(|_| ()) {
-            Err(Error::Io(ref e))
-                if e.kind() == io::ErrorKind::TimedOut || e.kind() == io::ErrorKind::WouldBlock =>
-            {
-                // we need to refresh the IDLE connection
-                self.terminate()?;
-                self.init()?;
-                self.wait_inner()
+        let mut buffer = Vec::new();
+        loop {
+            match self.session.readline(&mut buffer).map(|_| ()) {
+                Err(Error::Io(ref e))
+                    if e.kind() == io::ErrorKind::TimedOut
+                        || e.kind() == io::ErrorKind::WouldBlock =>
+                {
+                    if self.keepalive.is_some() {
+                        // we need to refresh the IDLE connection
+                        self.terminate()?;
+                        self.init()?;
+                    }
+                }
+                Err(err) => return Err(err),
+                Ok(_) => {
+                    // Unsolicited responses coming in from the server are not multi-line,
+                    // therefore, we don't need to worry about the remaining bytes since
+                    // they should always be terminated at a LF.
+                    let remaining = parse::parse_idle(&buffer, &mut self.unsolicited_responses_tx)?;
+                    assert_eq!(remaining.len(), 0);
+                    return Ok(());
+                }
             }
-            r => r,
         }
     }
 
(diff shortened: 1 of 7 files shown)


norcalli:
Fixes #128

norcalli:
I had to change around the logic for the "read_response" after "terminate" because there could've been multiple messages sent from the server, so we need to iterate over all of them, but I also need to know when to stop, and the only cue to stop is the `OK` we get from the server for the IDLE `DONE`.

norcalli:
I'm also not sure about making the unsolicited responses be empty by default, because I found it surprising. However, from an engineering perspective, it makes sense because if someone doesn't handle them then they build up in memory. This could be solved by a bounded channel with a sane default, as well, to impose an upper limit.

norcalli:
The test failure appears to be unrelated to any code I wrote.

norcalli:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[97, 49, 56]`,
 right: `[97, 49, 57]`', /home/ashkan/works/3rd/rust-imap/src/client.rs:1206:25
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

had an error related to match being one shorter than the client expected. I'm wondering if me doing the matching for the `ok` response in the handle is breaking some invariants.

norcalli:
FYI, I've been running this for the past week without any crashes or problems. I'm using it as my main ingestion pipeline for my mail now.

jonhoo:
I've been away on vacation for the past few weeks, so haven't had a chance to look. Will try to take a look once I get the time! Thanks for writing this up! :D

src/parse.rs
```diff
@@ -46,7 +46,7 @@ where
                     match map(resp)? {
                         MapOrNot::Map(t) => things.push(t),
                         MapOrNot::Not(resp) => match handle_unilateral(resp, unsolicited) {
-                            Some(Response::Fetch(..)) => continue,
+                            // Some(Response::Fetch(..)) => continue,
```
jonhoo: Why this change?

src/extensions/idle.rs
```diff
@@ -77,31 +84,53 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+                        || e.kind() == io::ErrorKind::WouldBlock =>
+                {
+                    if self.keepalive.is_some() {
+                        // we need to refresh the IDLE connection
+                        self.terminate(true)?;
+                        self.init()?;
+                    }
+                }
+                Err(err) => return Err(err),
+                Ok(_) => {
```
jonhoo: I don't understand this clause at all. Why do we need this nested loop construct? And why do we need to terminate the IDLE for each iteration?

src/parse.rs
```diff
@@ -202,7 +199,7 @@ pub fn parse_mailbox(
                 if let imap_proto::Status::Ok = status {
                 } else {
                     // how can this happen for a Response::Data?
-                    unreachable!();
+                    unreachable!("Received something other than OK from mailbox data");
```
jonhoo: This should probably include `status` for convenience.

src/parse.rs
```diff
@@ -297,29 +292,124 @@ pub fn parse_ids(
...
+                    ..
+                },
+            )) => {
+                match status {
+                    imap_proto::Status::Ok => return Ok(true),
+                    imap_proto::Status::Bad => {
+                        return Err(Error::Bad(information.unwrap_or("").to_owned()))
+                    }
+                    imap_proto::Status::No => {
+                        return Err(Error::Bad(information.unwrap_or("").to_owned()))
```
jonhoo: Don't we also have `Error::No`?

src/extensions/idle.rs
```diff
@@ -77,31 +84,53 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+                    let _ = parse::parse_idle(&buffer, &mut self.unsolicited_responses_tx)?;
```
jonhoo: Specifically, shouldn't this use the return value from `parse_idle` to know whether or not `DONE` was sent?

src/parse.rs
```diff
@@ -297,29 +292,124 @@ pub fn parse_ids(
...
+                    }
+                    _ => unreachable!(),
+                }
+            }
+            Ok((rest, data)) => {
+                lines = rest;
+                if let Some(resp) = handle_unilateral(data, unsolicited) {
```
jonhoo: We should probably rename `handle_unilateral` to `try_handle_unilateral` to make it clearer that `Some` is returned if it wasn't handled. That's orthogonal to your change though :)

src/types/mod.rs
```diff
@@ -204,6 +205,60 @@ pub use self::capabilities::Capabilities;
...
+    TryCreate,
+    /// Next unique identifier value.
+    UidNext(u32),
+    /// The unique identifier validity value.
+    UidValidity(u32),
+    /// First message without the \Seen flag set.
+    Unseen(u32),
+}
+
+impl<'a> From<imap_proto::types::ResponseCode<'a>> for ResponseCode {
```
jonhoo: I wonder if [`ToOwned`](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html) is the trait to implement here.

src/unsolicited_responses.rs
```diff
@@ -0,0 +1,37 @@
...
+        UnsolicitedResponseSender {
+            sender,
+            allow: EnumSet::empty(),
+        }
+    }
+
+    // Set the new filter mask, and remove unwanted responses from the current queue.
+    pub(crate) fn request(
+        &mut self,
+        receiver: &mpsc::Receiver<UnsolicitedResponse>,
```
jonhoo: I'm not sure I understand what this receiver is? Where does it come from?

src/unsolicited_responses.rs
```diff
@@ -0,0 +1,37 @@
...
+        mask: EnumSet<UnsolicitedResponseCategory>,
+    ) {
+        self.allow = mask;
+        for message in receiver.try_iter() {
```
jonhoo: This also seems pretty strange -- what about future messages? Are they never forwarded then?

jonhoo:
Hi @norcalli -- a gentle ping on this :)

norcalli:
@jonhoo hey sorry for the late reply, gonna take a look at this today.

hpk42:
@norcalli just so you i'd be interested in your PR landing but my Rust-Fu is not good enough yet to help much.

jonhoo:
@norcalli Another gentle ping on this :)

src/unsolicited_responses.rs
jonhoo: Long time no see :) Did you ever get around to doing this rewrite?

jonhoo:
@norcalli Another gentle ping on this :)

jonhoo:
@norcalli I'd like to try to push out 3.0.0 soon — want to give this another stab?

jonhoo:
Replaced by #186.

PR #134, opened 2019-09-02
Add BODYSTRUCTURE support

### LINKED ISSUE #116: Support BODYSTRUCTURE parsing for FETCH results
I think this might rely on the underlying parser crate supporting this (https://github.com/djc/imap-proto/issues/15), but it would be nice if this crate could expose a parsed BODYSTRUCTURE for messages (FETCH results).

This would be useful to allow the client to determine properties of a message in a bandwidth-constrained environment. Currently, the a message with large attachments must be fully downloaded and parsed just to determine that there are attachments.

avitex:
Exposes the body structure from a query.
Closes #116

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/134)

### DIFF
--- a/src/types/fetch.rs
+++ b/src/types/fetch.rs
@@ -1,6 +1,6 @@
 use super::{Flag, Seq, Uid};
 use chrono::{DateTime, FixedOffset};
-use imap_proto::types::{AttributeValue, Envelope, MessageSection, SectionPath};
+use imap_proto::types::{AttributeValue, Envelope, MessageSection, SectionPath, BodyStructure};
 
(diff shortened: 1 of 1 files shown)


avitex:
Sure can. Will add one to this branch tomorrow hopefully :)

avitex:
@jonhoo At what level would you like a test placed?

jonhoo:
Unit tests in `src/lib.rs` are fine. Of course, integration tests would be great, but they're also much more of a pain to write, and this is a relatively small feature anyway :)

avitex:
Looks like this is already in now :)

PR #135, opened 2019-09-03
Improve auth resp parsing

avitex:
- Regex is lazy init.
- Removes two allocations and a panic.
- Breaking change on DataNotUtf8 (wasn't used).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ imap-proto = "0.8"
 nom = "4.0"
 base64 = "0.10"
 chrono = "0.4"
+lazy_static = "1.4"
 
 [dev-dependencies]
 lettre = "0.9"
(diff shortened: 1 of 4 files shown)


avitex:
Travis is complaining on `rust 1.32`, is this version required over stable for any reason?

jonhoo:
@avitex I try to make it so that the library still works with older versions of rustc where possible, as some people are sadly stuck on older versions for one reason or another. It'd probably be okay to bump the min version slightly though. What version does the code compile with?

avitex:
Not sure, the older version just isn't doing the smarts with the lifetime, but I'll check tomorrow. If you end up bumping the version high enough, could perhaps go with the next rust edition ;)

Im off to sleep for now!

jonhoo:
Ohh, that's totally what's tripping us up. NLL only works on the 2018 edition in the older compilers. Let me bump to the new edition real quick and then this should work (you'll have to merge with `master` though).

avitex:
Whatever is easiest for you my friend. I can do a quick rebase no problem

jonhoo:
There, try syncing with the changes to `master` now!

avitex:
Should be all good to go now :)

jonhoo:
Published in `2.0.0` :tada:

PR #136, opened 2019-09-04
Make capabilities has_str case insensitive

avitex:
Related djc/tokio-imap/issues/54


jonhoo:
Looks good to me, thanks!

PR #138, opened 2019-09-17
(feat) Provide example integration with Rustls crate

brycefisher:
By checking this code in as an example, we can prevent bitrot as we continuously test this
in CI. It also demonstrates how to setup TLS without relying on platform specific dependencies
via native_tls crate.

First step of https://github.com/jonhoo/rust-imap/issues/123

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,12 +9,12 @@ stages:
  # This represents the minimum Rust version supported.
  # Tests are not run as tests may require newer versions of rust.
  - stage: msrv
-   displayName: "Minimum supported Rust version: 1.32.0"
+   displayName: "Minimum supported Rust version: 1.36.0"
    dependsOn: []
    jobs:
      - template: azure/cargo-check.yml@templates
        parameters:
-        rust: 1.32.0
+        rust: 1.36.0
  - stage: test
    displayName: Test suite
    dependsOn: check
--- a/examples/rustls.rs
+++ b/examples/rustls.rs
@@ -3,12 +3,10 @@ extern crate rustls_connector;
 
 use std::{env, error::Error, net::TcpStream};
 
-use dotenv::dotenv;
 use rustls_connector::RustlsConnector;
 
 fn main() -> Result<(), Box<dyn Error>> {
     // Read config from environment or .env file
-    dotenv().ok();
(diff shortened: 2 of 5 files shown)


azure-pipelines.yml
```diff
@@ -9,12 +9,12 @@ stages:
...
  # Tests are not run as tests may require newer versions of rust.
  - stage: msrv
-   displayName: "Minimum supported Rust version: 1.32.0"
+   displayName: "Minimum supported Rust version: 1.36.0"
    dependsOn: []
    jobs:
      - template: azure/cargo-check.yml@templates
        parameters:
-        rust: 1.32.0
+        rust: 1.36.0
```
jonhoo: Ouch, is this really necessary? What causes the need for this bump?

examples/rustls.rs
```diff
@@ -0,0 +1,65 @@
...
+extern crate rustls_connector;
+
+use std::{env, error::Error, net::TcpStream};
+
+use dotenv::dotenv;
+use rustls_connector::RustlsConnector;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // Read config from environment or .env file
+    dotenv().ok();
```
jonhoo: I think we probably don't need to include dotenv in this example.

brycefisher:
I know!!! I was disappointed by this too. The current version of Rustls is tested against 1.36.0 and one of its dependencies relies on the alloc crate (cant remember which one).

There's a few things we could do here:

 1. Somehow compile the rustls example code separately since everything else in the imap crate is 1.32.0 compatible (probably make a separate crate as a cargo workspace)
 2. Downgrade rustls until we find something that works at a lower version number.

I'd probably opt for 1
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

brycefisher:
Perfect! I'll take it out
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

jonhoo:
Hmm... Testing with a separate version just for one subcrate will be annoying, as will not having it show up as a regular example. I think I'm inclined to keep the MSRV version bump. It's too bad, but I don't think it's actually a problem since the bump won't apply to those who rely just on the `imap` crate (since `rusttls` is a `dev-dependency`).

PR #140, opened 2019-09-20
(feature) Allow opt-out from `native_tls` crate

brycefisher:
Ready for review! @jonhoo 

Establishes conditional compilation for all integration with the native_tls crate. Since native_tls has been deeply integrated into this crate for a long time, we want to maintain backwards compatibility by making this feature part of the default.

For a consumer of this crate to "opt-out", include this in cargo.toml:

```toml
[dependencies.imap]
version = "0.16.0"        # Replace this with the correct version
default-features = false
```

See the conversation for details on this approach:
https://github.com/jonhoo/rust-imap/issues/123

---

In order to discourage developers from avoiding TLS, we're also simultaneously removing the `connect_insecure` method. However, it is still possible to use an unsecured TCP connection for an imap session by using `imap::Client::new()` and providing a naked TcpStream.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,6 +17,7 @@ task:
   build_script:
     - . $HOME/.cargo/env
     - cargo build --all-targets --verbose
+    - cargo build --all-targets --verbose --no-default-features
   test_script:
     - . $HOME/.cargo/env
     - cargo test --examples
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,7 +17,6 @@ task:
   build_script:
     - . $HOME/.cargo/env
     - cargo build --all-targets --verbose
-    - cargo build --all-targets --verbose --no-default-features
   test_script:
     - . $HOME/.cargo/env
     - cargo test --examples
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,12 @@ maintenance = { status = "actively-developed" }
 is-it-maintained-issue-resolution = { repository = "jonhoo/rust-imap" }
 is-it-maintained-open-issues = { repository = "jonhoo/rust-imap" }
 
+[features]
+tls = ["native-tls"]
+default = ["tls"]
+
 [dependencies]
-native-tls = "0.2.2"
+native-tls = { version = "0.2.2", optional = true }
 regex = "1.0"
 bufstream = "0.1"
 imap-proto = "0.9.0"
@@ -35,3 +39,15 @@ lazy_static = "1.4"
 lettre = "0.9"
 lettre_email = "0.9"
 rustls-connector = "0.8.0"
+
+[[example]]
+name = "basic"
+required-features = ["default"]
+
+[[example]]
+name = "gmail_oauth2"
+required-features = ["default"]
+
+[[test]]
+name = "imap_integration"
+required-features = ["default"]
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,8 +41,6 @@ stages:
        displayName: Run doctests
      - script: cargo test --lib
        displayName: Run unit tests
-     - script: cargo build --all-targets --verbose --no-default-features
-       displayName: Compile without openssl
    - job: integration
      displayName: cargo test
      pool:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,6 +41,8 @@ stages:
        displayName: Run doctests
      - script: cargo test --lib
        displayName: Run unit tests
+     - script: cargo build --all-targets --verbose --no-default-features
+       displayName: Compile without openssl
    - job: integration
      displayName: cargo test
      pool:
--- a/src/client.rs
+++ b/src/client.rs
@@ -110,39 +110,6 @@ impl<T: Read + Write> DerefMut for Session<T> {
     }
 }
 
-/// Connect to a server using an insecure TCP connection.
-///
-/// The returned [`Client`] is unauthenticated; to access session-related methods (through
-/// [`Session`]), use [`Client::login`] or [`Client::authenticate`].
-///
-/// Consider using [`connect`] for a secured connection where possible.
-/// You can upgrade an insecure client to a secure one using [`Client::secure`].
-/// ```rust,no_run
-/// # extern crate native_tls;
-/// # extern crate imap;
-/// # use std::io;
-/// # use native_tls::TlsConnector;
-/// # fn main() {
-/// // a plain, unencrypted TCP connection
-/// let client = imap::connect_insecure(("imap.example.org", 143)).unwrap();
-///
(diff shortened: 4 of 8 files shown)


.cirrus.yml
```diff
@@ -17,6 +17,7 @@ task:
   build_script:
     - . $HOME/.cargo/env
     - cargo build --all-targets --verbose
+    - cargo build --all-targets --verbose --no-default-features
```
jonhoo: You don't need this -- it is already handled by the Azure Pipelines CI. Cirrus is just there to test on FreeBSD.

azure-pipelines.yml
```diff
@@ -41,6 +41,8 @@ stages:
        displayName: Run doctests
      - script: cargo test --lib
        displayName: Run unit tests
+     - script: cargo build --all-targets --verbose --no-default-features
```
jonhoo: This is also not necessary. The `check` step above will already do it :)

src/client.rs
```diff
@@ -1,5 +1,6 @@
 use base64;
 use bufstream::BufStream;
+#[cfg(feature = "default")]
```
jonhoo: This should use
```rust
#[cfg(feature = "native-tls")]
```

Cargo.toml
```diff
@@ -21,8 +21,11 @@ maintenance = { status = "actively-developed" }
 is-it-maintained-issue-resolution = { repository = "jonhoo/rust-imap" }
 is-it-maintained-open-issues = { repository = "jonhoo/rust-imap" }
 
+[features]
+default = ["native-tls"]
```
jonhoo: I think I would name the feature `tls` explicitly, like so:
```toml
[features]
tls = ["native-tls"]
default = ["tls"]
```
And then use
```rust
#[cfg(feature = "tls")]
```

src/error.rs
```diff
@@ -22,8 +25,10 @@ pub enum Error {
     /// An `io::Error` that occurred while trying to read or write to a network stream.
     Io(IoError),
     /// An error from the `native_tls` library during the TLS handshake.
+    #[cfg(feature = "default")]
```
jonhoo: So, this is actually pretty problematic, because it makes features non-additive. I think for this to be okay, we'll need https://github.com/rust-lang/rust/issues/44109, which hasn't landed yet. One possible solution would be to add a variant to the end of this `enum` like this:
```rust
#[doc(hidden)]
__Nonexhaustive,
```
That prevents people from matching on `Error`, and thus makes the features additive again.

jonhoo:
I also just saw https://github.com/jonhoo/rust-imap/issues/123#issuecomment-533771168, and I think `secure` _does_ still make sense. It should be generic though; currently it is under `impl Client<TcpStream>`, but I think it should be possible to `secure` any transport that is `Read + Write`.

jonhoo:
As for that test, I would re-write it to manually establish the `TcpStream`, then use `Client::new` + `secure`.

brycefisher:
Np! Will do tonight 
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

brycefisher:
I'll not super experienced with cargo features, but when writing tests and trying to prepare this PR, I found that compiling with different sets of features results in the compiler type-checking different sets of code. To ensure that both default and no default compiles, I _think_ it is necessary to at least run a cargo check this way.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

brycefisher:
I see! Ok, so we probably do want the check with --no-default-features in Azure pipelines, but its probably overkill here.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

brycefisher:
Right! Or rather feature = "tls" as you mentioned earlier
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

brycefisher:
Love it!
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

brycefisher:
I'm very happy to make that change, but I don't fully understand the problem. Does the link you shared explain this issue in more depth?
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

brycefisher:
That makes a ton of sense to me. I'll  see how far I can get on this!!
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

jonhoo:
So, with regards to Azure Pipelines, the current CI already checks with `--no-default-features`, so it should already include all the steps you had there :)

As for additive features, consider if one crate, A, has:
```toml
[dependencies]
imap = { version = "...", default-features = false }
```
And another, B, has:
```toml
[dependencies]
imap = "..." # so with the tls feature
```
If you have a crate C that depends on both A **and** B, then cargo will only compile imap once with all the features enabled. However, consider if A contained something like:
```rust
match imap::connect(...) {
  Ok(c) => ...,
  Err(imap::Error::Io(...)) => ...,
  Err(imap::Error::Bad(...)) => ...,
  Err(imap::Error::No(...)) => ...,
  ...
}
```
That compiles just fine if the `tls` feature is _not_ enabled, but would not compiled if the `tls` feature _is_ enabled (it would complain that you are missing the variants `TlsHandshake` and `Tls`). This is what we mean by features having to be "additive" — adding a feature can never _remove_ something or otherwise cause something that compiled to no longer compile, because cargo will frequently add features to combine dependencies.

By adding the proposed hidden variant:
```rust
#[doc(hidden)]
__Nonexhaustive,
```

Now it'd be pretty clear to whoever wrote crate A that they should never be trying to match on all the variants of `Error`, because new variants may be added.

brycefisher:
I see! Okay, got it.

brycefisher:
> secure does still make sense. It should be generic though; currently it is under impl Client<TcpStream>, but I think it should be possible to secure any transport that is Read + Write.

I'm happy to tackle this! However, I'm seeing what looks like a lot of refactoring work here to nudge the type signatures. In particular the `Error::TlsHandshake` also knows about the underlying type of the transport. Making this enum generic will probably make a lot of other things more generic, and may make the `Error` type itself a wee bit unergonomic to users of the library.

I spent a few minutes trying to think of how to change the `Error` type without being so generic. The only decent idea I came up with was trying to immediately pull out the values of the `TlsHandshakeError` as strings and storing those on the `Error::TlsHandshake`. It feels a bit unidiomatic, but on the other hand, the any `Error` trait object should be easy to pull strings from.

jonhoo:
Ah, that's a shame about `.secure`, but it's probably okay for now. It should be a backwards-compatible change to change it later anyway.

brycefisher:
Oh good catch! I wrote this code last night but somehow forgot to push it.

jonhoo:
Looks great, thank you!

jonhoo:
Published in `2.0.0` :tada:

PR #144, opened 2019-10-16
feat: expose read_greeting

dignifiedquire:
This is needed when manually using `Client::new()`.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1104,10 +1104,11 @@ impl<T: Read + Write> Session<T> {
 
 impl<T: Read + Write> Connection<T> {
     /// Read the greeting from the connection. Needs to be done after `connect`ing.
+    ///
+    /// Panics if called more than once on the same `Connection`.
     pub fn read_greeting(&mut self) -> Result<Vec<u8>> {
-        if self.greeting_read {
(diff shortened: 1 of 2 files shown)


jonhoo:
Hmm, it's a little weird for this to be `pub`, since it basically gives a direct way of calling `readline` on a connection at any time. I would prefer for there to at least be a runtime check that this is indeed only called when the connection has just been established.

dignifiedquire:
I added a tracking thing, to enforce it is only read once.

src/client.rs
```diff
@@ -1099,10 +1103,16 @@ impl<T: Read + Write> Session<T> {
 }
 
 impl<T: Read + Write> Connection<T> {
-    fn read_greeting(&mut self) -> Result<()> {
+    /// Read the greeting from the connection. Needs to be done after `connect`ing.
+    pub fn read_greeting(&mut self) -> Result<Vec<u8>> {
+        if self.greeting_read {
```
jonhoo: I think I would just make this `assert!(!self.greeting_read)`. This shouldn't generally be something that only happens _some_ of the time.

dignifiedquire: fixed

jonhoo:
Published in `2.0.0` :tada:

Issue #139, opened 2019-09-19
wait_timeout doesn't work as expected


kanekv:
Documentation suggests (and in fact it will be useful) that `wait_timeout` blocks until new messages appear in inbox or after specified timeout. In fact it calls `wait_inner`.
https://github.com/jonhoo/rust-imap/blob/master/src/extensions/idle.rs#L148

And `wait_inner` in fact catches timeouts and restart itself, so timeouts would end up not stopping it as expected.
https://github.com/jonhoo/rust-imap/blob/master/src/extensions/idle.rs#L97

Is it intentional behavior or you wanted opposite logic in there?
@jonhoo

jonhoo:
Ah, interesting, that is a good catch! You're right -- we would want `wait_inner` to not restart the connection on a timeout if `wait_timeout` was used (it should still do so in other cases where it gets `TimedOut` though. I'd be happy to review a PR!

dario23:
should this be closed, now that #159 is merged?

jonhoo:
@dario23 Ah, yes, good catch, thanks!

PR #148, opened 2019-10-23
add idle example and add one debug line to idle.rs

hpk42:
- new example/idle.rs
- this adds a debug line also highlighting #139 

this example is part of an attempt to debug an issue we have with "hanging idle/imap" threads and DeltaChat's usage of rust-imap.  On my linux machine if i disconnect my WiFi during the wait_keepalive() call of the example, and then reconnect after a while i even get a crash. https://gist.github.com/hpk42/2da33921476ddb57417ce2b501628799

however, this PR is also independently maybe interesting an an example.

### DIFF
--- a/examples/idle.rs
+++ b/examples/idle.rs
@@ -1,0 +1,70 @@
+extern crate imap;
+extern crate native_tls;
+use std::env;
+use std::time::Duration;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() != 4 {
+        eprintln!("need three arguments: imap-server login password");
+    } else {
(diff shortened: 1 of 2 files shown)


jonhoo:
Could you take a look at the test failures? I think at least part of them is failure to compile with `cargo check --no-default-features`.

hpk42:
closing this PR as work is now being done rather in the new (experimental) https://github.com/async-email/async-imap which is an async fork of rust-imap.  This has much clearer idle handling producing  timeout/manual-interrupt/newdata outcomes, i.e. one can interrupt it safely from another thread.

PR #149, opened 2019-10-29
Adding METADATA support

Sergey-Kitov:
Add support for METADATA IMAP extension, RFC-5464

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/149)

### DIFF
--- a/src/extensions/metadata.rs
+++ b/src/extensions/metadata.rs
@@ -1,0 +1,179 @@
+//! Adds support for the IMAP METADATA extension specificed in [RFC
+//! 5464](https://tools.ietf.org/html/rfc5464).
+
+use crate::client::*;
+use crate::error::{Error, ParseError, Result};
+use crate::parse::handle_unilateral;
+use crate::types::*;
+use imap_proto::types::{MailboxDatum, Metadata, Response};
+use std::io::{Read, Write};
+use std::sync::mpsc;
+
+trait CmdListItemFormat {
+    fn format_as_cmd_list_item(&self) -> String;
+}
+
+impl CmdListItemFormat for Metadata {
+    fn format_as_cmd_list_item(&self) -> String {
+        format!(
+            "{} {}",
+            validate_str(self.entry.as_str()).unwrap(),
+            self.value
+                .as_ref()
+                .map(|v| validate_str(v.as_str()).unwrap())
+                .unwrap_or("NIL".to_string())
+        )
+    }
+}
+
+/// Represents variants of DEPTH parameters for GETMETADATA command.
+///  "0" - no entries below the specified entry are returned
+///  "1" - only entries immediately below the specified entry are returned
+///  "infinity" -  all entries below the specified entry are returned
+/// See [RFC 5464, section 4.2.2](https://tools.ietf.org/html/rfc5464#section-4.2.2)
+#[derive(Debug, Copy, Clone)]
+pub enum MetadataDepth {
+    /// Depth 0 for get metadata
+    Zero,
+    /// Depth 1 for get metadata
+    One,
+    /// Depth infinity for get metadata
+    Inf,
+}
+
+impl Default for MetadataDepth {
+    fn default() -> Self {
+        Self::Zero
+    }
+}
+
+impl MetadataDepth {
+    fn depth_str<'a>(self) -> &'a str {
+        match self {
+            MetadataDepth::Zero => return "0",
+            MetadataDepth::One => return "1",
+            MetadataDepth::Inf => return "infinity",
+        }
+    }
+}
+
+fn parse_metadata<'a>(
+    mut lines: &'a [u8],
+    unsolicited: &'a mut mpsc::Sender<UnsolicitedResponse>,
+) -> Result<Vec<Metadata>> {
+    let mut res: Vec<Metadata> = Vec::new();
+    loop {
+        if lines.is_empty() {
+            break Ok(res);
+        }
+
+        match imap_proto::parse_response(lines) {
+            Ok((rest, resp)) => {
+                lines = rest;
+                match resp {
+                    Response::MailboxData(MailboxDatum::MetadataSolicited {
+                        mailbox: _,
+                        mut values,
+                    }) => {
+                        res.append(&mut values);
+                    }
+                    _ => {
+                        if let Some(unhandled) = handle_unilateral(resp, unsolicited) {
+                            break Err(unhandled.into());
+                        }
+                    }
+                }
+            }
+            Err(_) => {
+                return Err(Error::Parse(ParseError::Invalid(lines.to_vec())));
+            }
+        }
+    }
+}
+
+impl<T: Read + Write> Session<T> {
+    /// Sends GETMETADATA command of the METADATA extension to IMAP protocol
+    /// to the server and returns the list of entries and their values.
+    /// Server support for the extension is indicated by METADATA capability.
+    /// @param mbox mailbox name. When the mailbox name is the empty string, this command retrieves server annotations. When the mailbox name is not empty, this command retrieves annotations on the specified mailbox.
+    /// @param entries list of metadata entries to be retrieved.
+    /// @param depth GETMETADATA DEPTH option, specifies if children entries are to be retrieved as well.
+    /// @param maxside GETMETADATA MAXSIZE option. When the MAXSIZE option is specified with the GETMETADATA command, it restricts which entry values are returned by the server. Only entry values that are less than or equal in octet size to the specified MAXSIZE limit are returned.
+    /// See [RFC 5464, section 4.2](https://tools.ietf.org/html/rfc5464#section-4.2) for more details.
+    pub fn get_metadata(
+        &mut self,
+        mbox: impl AsRef<str>,
+        entries: &[impl AsRef<str>],
+        depth: MetadataDepth,
+        maxsize: Option<usize>,
+    ) -> Result<Vec<Metadata>> {
+        let v: Vec<String> = entries
+            .iter()
+            .map(|e| validate_str(e.as_ref()).unwrap())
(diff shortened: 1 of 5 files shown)


jonhoo:
Sorry that it has taken me forever to look at this! Will dive into it now :)

Cargo.toml
```diff
@@ -29,7 +29,7 @@ default = ["tls"]
 native-tls = { version = "0.2.2", optional = true }
 regex = "1.0"
 bufstream = "0.1"
-imap-proto = "0.9.0"
+imap-proto = { git = "https://github.com/djc/tokio-imap", branch = "master" }
```
jonhoo, line 32: Is this still necessary?

src/extensions/metadata.rs
```diff
@@ -0,0 +1,166 @@
...
+                .as_ref()
+                .map(|v| validate_str(v.as_str()).unwrap())
+                .unwrap_or("NIL".to_string())
+        )
+    }
+}
+
+/// Represents variants of DEPTH parameters for GETMETADATA command
+#[derive(Debug, Copy, Clone)]
+pub enum MetadataDepth {
```
jonhoo, line 30: Maybe implement `Default` for this too, and set it to `Zero` (the spec says [that's the default](https://tools.ietf.org/html/rfc5464#section-4.2.2)).

src/extensions/metadata.rs
```diff
@@ -0,0 +1,166 @@
...
+            validate_str(self.entry.as_str()).unwrap(),
+            self.value
+                .as_ref()
+                .map(|v| validate_str(v.as_str()).unwrap())
+                .unwrap_or("NIL".to_string())
+        )
+    }
+}
+
+/// Represents variants of DEPTH parameters for GETMETADATA command
```
jonhoo, line 28: It'd be great to include more information here. You can probably take most of the text from [the spec](https://tools.ietf.org/html/rfc5464#section-4.2.2).

src/extensions/metadata.rs
```diff
@@ -0,0 +1,166 @@
...
+            Ok((rest, resp)) => {
+                lines = rest;
+                match resp {
+                    Response::MailboxData(MailboxDatum::MetadataSolicited {
+                        mailbox: _,
+                        mut values,
+                    }) => {
+                        res.append(&mut values);
+                    }
+                    Response::MailboxData(MailboxDatum::MetadataUnsolicited {
```
jonhoo, line 69: This doesn't quite seem right. I think this method should look similar to the other parsers, where it _just_ handles the solicited responses (here `MailboxDatum::MetadataSolicited`), and then has a fallback clause to catch _any_ unilateral server response:
https://github.com/jonhoo/rust-imap/blob/fbdb5d3be4465d3b010369a9b98ad8b0631b16fa/src/parse.rs#L183-L188

The reason for this is that unilateral responses may come _at any time_, including in the middle of other commands. So even while you are running the `GETMETADATA` command, a unilateral `EXPUNGE` response may arrive.

src/extensions/metadata.rs
```diff
@@ -0,0 +1,166 @@
...
+            }
+            Err(_) => {
+                return Err(Error::Parse(ParseError::Invalid(lines.to_vec())));
+            }
+        }
+    }
+}
+
+/// Sends GETMETADATA command to the server and returns the list of entries and their values.
+pub fn get_metadata<'a, S: AsRef<str>, T: Read + Write>(
```
jonhoo, line 91: This, and `set_metadata` should be methods _on_ `Session`, not free-standing methods like this. I'd also like to see the documentation a lot more extensive (again copying from the RFC). The docs for both methods should also make it very clear that this is an extension that may not be available for the given server. Ideally it'd also include some sample code for how to determine whether it's supported by checking the server capabilities.

src/extensions/metadata.rs
```diff
@@ -0,0 +1,166 @@
...
+    session: &'a mut Session<T>,
+    mbox: S,
+    entries: &[S],
```
jonhoo, line 94:
```suggestion
    mbox: impl AsRef<str>,
    entries: &[impl AsRef<str>],
```

src/extensions/metadata.rs
jonhoo: It should also document what each of the arguments do.

src/lib.rs
```diff
@@ -89,6 +89,8 @@ pub use crate::client::*;
 pub mod error;
 
 pub mod extensions;
+// mod metadata;
+pub use extensions::metadata::{get_metadata, set_metadata};
```
jonhoo, line 93: This, I think, will go away.

src/extensions/metadata.rs
jonhoo: The code for handling `MetadataUnsolicited` should be moved into `handle_unilateral`.

src/types/capabilities.rs
```diff
@@ -33,32 +33,28 @@ const AUTH_CAPABILITY_PREFIX: &str = "AUTH=";
 pub struct Capabilities(
     // Note that this field isn't *actually* 'static.
     // Rather, it is tied to the lifetime of the `ZeroCopy` that contains this `Name`.
-    pub(crate) HashSet<Capability<'static>>,
+    pub(crate) HashSet<imap_proto::Capability<'static>>,
```
jonhoo, line 36: The changes in this file seem unrelated?

src/types/mod.rs
```diff
@@ -262,6 +262,15 @@ pub enum UnsolicitedResponse {
     /// sequence numbers 9, 8, 7, 6, and 5.
     // TODO: the spec doesn't seem to say anything about when these may be received as unsolicited?
     Expunge(u32),
+
+    /// An unsolicited METADATA response (https://tools.ietf.org/html/rfc5464#section-4.4.2)
+    /// that reports a change in a server or mailbox annotation.
+    Metadata {
```
jonhoo, line 268: This is, sadly, a breaking change. But it seems like we'll have to do one soon anyway. See also #163.

Sergey-Kitov: I am not sure about what I can do regarding this comment, and I have to say that our project has switched to another library, so I have processed the comments I could, but I can't spend more time on this, so you are free to use this code, but without support from our team. I apologize for any inconvenience.

jonhoo: So, the way I approached this with another recent change was to simply drop the unsolicited messages, rather than forward them, with the intent that on the next breaking change, we would add the enum variant and forward them then. The alternative is that we keep things the way they are now, and I only merge this when I make the next breaking release (which I suspect won't be in _too_ long).

I appreciate you updating the PR even though you have moved on from the library! As I don't need this myself either, I will probably then let the PR linger until someone else who needs it comes along and can push it over the finish line.

src/extensions/metadata.rs
```diff
@@ -0,0 +1,179 @@
...
+                .unwrap_or("NIL".to_string())
+        )
+    }
+}
+
+/// Represents variants of DEPTH parameters for GETMETADATA command.
+///  "0" - no entries below the specified entry are returned
+///  "1" - only entries immediately below the specified entry are returned
+///  "infinity" -  all entries below the specified entry are returned
+/// See [RFC 5464, section 4.2.2](https://tools.ietf.org/html/rfc5464#section-4.2.2)
```
jonhoo, line 33: This doc comment won't render nicely, since in Markdown linebreaks aren't significant. We'll probably want to make this a list (start each item with a `-`), and have an empty line above and below the list to make sure we get paragraph breaks. I also think it'd be great to include some concrete examples here of what paths would or would not be matched with different parameters (the same way the RFC does).

src/extensions/metadata.rs
```diff
@@ -0,0 +1,179 @@
...
+#[derive(Debug, Copy, Clone)]
+pub enum MetadataDepth {
+    /// Depth 0 for get metadata
+    Zero,
+    /// Depth 1 for get metadata
+    One,
+    /// Depth infinity for get metadata
+    Inf,
```
jonhoo, line 41: I think we should spell this out completely.

src/extensions/metadata.rs
```diff
@@ -0,0 +1,179 @@
...
+
+impl<T: Read + Write> Session<T> {
+    /// Sends GETMETADATA command of the METADATA extension to IMAP protocol
+    /// to the server and returns the list of entries and their values.
+    /// Server support for the extension is indicated by METADATA capability.
+    /// @param mbox mailbox name. When the mailbox name is the empty string, this command retrieves server annotations. When the mailbox name is not empty, this command retrieves annotations on the specified mailbox.
+    /// @param entries list of metadata entries to be retrieved.
+    /// @param depth GETMETADATA DEPTH option, specifies if children entries are to be retrieved as well.
+    /// @param maxside GETMETADATA MAXSIZE option. When the MAXSIZE option is specified with the GETMETADATA command, it restricts which entry values are returned by the server. Only entry values that are less than or equal in octet size to the specified MAXSIZE limit are returned.
+    /// See [RFC 5464, section 4.2](https://tools.ietf.org/html/rfc5464#section-4.2) for more details.
```
jonhoo, line 102: This too will render quite oddly with `rustdoc`. We'll probably want to write this a bit more like prose, in the style of other the other doc comments. An example code block to check for the metadata capability would also be great here.

jonhoo: And same for `set_metadata`.

jonhoo:
As discussed in https://github.com/jonhoo/rust-imap/pull/149#discussion_r427284775, the PR is probably going to sit idle until someone who needs it has the bandwidth to push it over the finish line. If that's you, drop me a line! I've left comments on what remains :)

jonhoo:
I'm going to merge this now for the 3.0.0 alpha and then to the remaining comments myself :)

PR #153, opened 2019-11-30
Add `imap::connect_starttls` convenience function

Celti:
Add a `imap::connect_starttls` convenience function, with the same
syntax as `imap::connect`, to make STARTTLS connections as easily as
IMAPS.

PR#140 removed `imap::connect_insecure` with the stated goal of
encouraging users to connect securely. With that change, users are
forced to construct their own `TcpStream`s when interacting with
non-IMAPS servers.

This change may perversely incentivise the use of insecure connections,
instead of discouraging them, as users may blindly copy-paste code
involving `Client::new` and `TcpStream`s so things "just work", without
the visual indicator of `_insecure` suggesting something is wrong.

Tangentially, [RFC 2595] encourages using STARTTLS instead of raw SSL
sockets. Ideally, we should support both options equally well.

[RFC 2595]: https://tools.ietf.org/html/rfc2595#section-7

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -154,6 +154,46 @@ pub fn connect<A: ToSocketAddrs, S: AsRef<str>>(
     }
 }
 
+/// Connect to a server and upgrade to a TLS-encrypted connection.
+///
+/// This is the [STARTTLS](https://tools.ietf.org/html/rfc2595) equivalent to [`connect`]. All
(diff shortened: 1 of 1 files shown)


jonhoo:
Perfect, thank you!

jonhoo:
Published in `2.0.0` :tada:

PR #158, opened 2020-03-29
Make Display implementation independent of Error::description

mordak:
Since Error::description was deprecated in rust 1.42, the use of description() causes warnings. I suspect that you don't want to delete the existing implementation of these deprecated functions just yet.

This PR updates the implementation of Display to not rely on description(), and marks the uses found in the existing description() implementation as allowed in order to suppress the warnings there.

You should be able to just delete the implementations of Error::description and Error::cause whenever you're comfortable doing that.

If you don't want to merge this that's okay with me. I just saw the warnings and thought I could help make them go away in a forward-looking but backwards-compatible way.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/158)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -94,15 +94,18 @@ impl fmt::Display for Error {
             #[cfg(feature = "tls")]
             Error::TlsHandshake(ref e) => fmt::Display::fmt(e, f),
             Error::Validate(ref e) => fmt::Display::fmt(e, f),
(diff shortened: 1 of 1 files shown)


jonhoo:
This looks great to me, thank you!
You can ignore the failure, it's just `rustfmt` not being available on `nightly`

PR #159, opened 2020-04-08
Support timing out wait on IDLE

mredaelli:
I also had the issue described [here](https://github.com/jonhoo/rust-imap/issues/139), so I did a quick patch that is certainly not elegant (and keep in mind this is my first rust ever), but seems to work for me.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/159)

### DIFF
--- a/src/extensions/idle.rs
+++ b/src/extensions/idle.rs
@@ -31,6 +31,15 @@ pub struct Handle<'a, T: Read + Write> {
     done: bool,
 }
 
+/// The result of a wait on a [`Handle`]
+#[derive(Debug, PartialEq, Eq)]
+pub enum WaitOutcome {
+    /// The wait timed out
+    TimedOut,
+    /// The mailbox was modified
+    MailboxChanged,
+}
+
 /// Must be implemented for a transport in order for a `Session` using that transport to support
 /// operations with timeouts.
 ///
@@ -91,34 +100,37 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
     /// Internal helper that doesn't consume self.
     ///
     /// This is necessary so that we can keep using the inner `Session` in `wait_keepalive`.
-    fn wait_inner(&mut self) -> Result<()> {
+    fn wait_inner(&mut self, reconnect: bool) -> Result<WaitOutcome> {
         let mut v = Vec::new();
         loop {
-            match self.session.readline(&mut v).map(|_| ()) {
+            let result = match self.session.readline(&mut v).map(|_| ()) {
                 Err(Error::Io(ref e))
                     if e.kind() == io::ErrorKind::TimedOut
                         || e.kind() == io::ErrorKind::WouldBlock =>
                 {
-                    // we need to refresh the IDLE connection
-                    self.terminate()?;
-                    self.init()?;
-                    return self.wait_inner();
+                    if reconnect {
+                        self.terminate()?;
+                        self.init()?;
+                        return self.wait_inner(reconnect);
+                    }
+                    Ok(WaitOutcome::TimedOut)
                 }
-                r => r,
+                Ok(()) => Ok(WaitOutcome::MailboxChanged),
+                Err(r) => Err(r),
             }?;
 
             // Handle Dovecot's imap_idle_notify_interval message
             if v.eq_ignore_ascii_case(b"* OK Still here\r\n") {
                 v.clear();
             } else {
-                break Ok(());
+                break Ok(result);
             }
         }
     }
 
     /// Block until the selected mailbox changes.
     pub fn wait(mut self) -> Result<()> {
-        self.wait_inner()
+        self.wait_inner(true).map(|_| ())
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/extensions/idle.rs
```diff
@@ -91,24 +100,28 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
-                // we need to refresh the IDLE connection
-                self.terminate()?;
-                self.init()?;
-                self.wait_inner()
+                if reconnect {
+                    self.terminate()?;
+                    self.init()?;
+                    self.wait_inner(reconnect)
+                } else {
+                    Ok(WaitOutcome::TimedOut)
```
jonhoo, line 114: Thanks for the patch -- it looks pretty reasonable to me! I think we can avoid changing the public API though (and thus avoid a major version bump). What if we just returned the underlying error when `reconnect == false`? That way, the caller can determined if a timeout occurred simply by inspecting the type of the error. We should of course document that in the documentation for `wait_timeout`. And that way, the return type for all the `wait` methods can stay the same as they were.

mredaelli: Makes sense, although using an error condition to signal the very purpose of the function feels a bit suboptimal.

Especially because clients would need to understand the semantics of the underlying session and (if my experiments are correct) would have to check (at least) for  `WouldBlock` (which sounds counter-intuitive and I don't know what that value means :) )

Also, it would be nice to integrate getting back *what* was changed from the `Handle` (there's another open issue for that, right?), in which case what I called `WaitOutcome` would be better suited to turn into something more "dataful".

So anyway, yes, maybe the trouble of a major version bump could wait for the integration of that part too?

Up to you :)

jonhoo: I think the `WouldBlock` error code should actually not be handled there at all, and `TimedOut` is the only one that "matters". But you're right, we could translate that particular error for `wait_timeout` specifically. I don't think we need to change the other methods though.

I agree with you that `WaitOutcome` would be a good addition eventually when we also add data, but probably not worth a breaking change until we do. For the time being, how about:

 - Revert the return type of all the `pub` methods
 - Mark `wait_timeout` as `#[deprecated]`
 - Add `pub fn wait_with_timeout`, which returns `Result<WaitOutcome>`

mredaelli: > I think the WouldBlock error code should actually not be handled there at all, and TimedOut is the only one that "matters"

That was my first attempt: a new branch in the `match`, on `Err(e) if e.kind() == io::ErrorKind::TimedOut`. But after the timeout the function did not return :/

> For the time being, how about:

That makes a lot of sense

mredaelli: Sorry it took me a while :P

Is this what you meant?

jonhoo: Yes, exactly!

src/extensions/idle.rs
```diff
@@ -91,24 +100,28 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
                 if e.kind() == io::ErrorKind::TimedOut || e.kind() == io::ErrorKind::WouldBlock =>
             {
-                // we need to refresh the IDLE connection
-                self.terminate()?;
-                self.init()?;
-                self.wait_inner()
+                if reconnect {
+                    self.terminate()?;
+                    self.init()?;
+                    self.wait_inner(reconnect)
```
jonhoo, line 112: This should _probably_ be a `return` instead. Just to try and nudge Rust to make it tail-call optimized :p

src/extensions/idle.rs
```diff
@@ -137,16 +150,26 @@ impl<'a, T: SetReadTimeout + Read + Write + 'a> Handle<'a, T> {
         // This still allows a client to receive immediate mailbox updates even
         // though it need only "poll" at half hour intervals.
         let keepalive = self.keepalive;
-        self.wait_timeout(keepalive)
+        self.timed_wait(keepalive, true).map(|_| ())
     }
 
     /// Block until the selected mailbox changes, or until the given amount of time has expired.
-    pub fn wait_timeout(mut self, timeout: Duration) -> Result<()> {
+    #[deprecated]
```
jonhoo, line 157: Let's also include `note = "use wait_with_timeout instead"`.

src/extensions/idle.rs
```diff
@@ -137,16 +150,26 @@ impl<'a, T: SetReadTimeout + Read + Write + 'a> Handle<'a, T> {
...
+    pub fn wait_timeout(self, timeout: Duration) -> Result<()> {
+        self.timed_wait(timeout, false).map(|_| ())
```
jonhoo, line 159: Let's have this call `wait_with_timeout` instead.

src/extensions/idle.rs
mredaelli: Oh? I had no idea! Nice :)

src/extensions/idle.rs
```diff
@@ -91,24 +100,28 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
             {
-                // we need to refresh the IDLE connection
-                self.terminate()?;
-                self.init()?;
-                self.wait_inner()
+                if reconnect {
+                    self.terminate()?;
+                    self.init()?;
+                    return self.wait_inner(reconnect);
+                } else {
```
jonhoo, line 113: You actually don't even need the `else` any more now that the statement above returns! Nice.

mredaelli:
Grrr I just noticed there's now a conflict. I'll have to understand what's going on to solve it.

src/extensions/idle.rs
```diff
@@ -31,6 +31,15 @@ pub struct Handle<'a, T: Read + Write> {
     done: bool,
 }
 
+/// The result of a wait on a `Handle`
```
jonhoo, line 34:
```suggestion
/// The result of a wait on a [`Handle`]
```
This makes it a clickable link!

mredaelli:
There we go: it still works for my use-case, but to be honest I'm not sure if it still covers the Dovecot one

jonhoo:
From my reading of it, it should! Thanks :D

PR #160, opened 2020-04-19
Handle Dovecot's periodic IDLE notification messages ("OK Still here").

mordak:
These messages show up at an interval controlled by Dovecot's imap_idle_notify_interval setting, which defaults to 2 minutes. These messages were causing the IDLE loop to exit prematurely because it looks like a notification response from the server. This means that IDLE connections to servers that do this are short lived.

With this change, these "OK Still here" messages instead keep the IDLE session alive indefinitely unless the socket timeout is less than the server notify interval. When something changes on the mailbox, the normal notification (FETCH, EXISTS, etc.) causes the IDLE loop to exit as before and return to the client.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/160)

### DIFF
--- a/src/extensions/idle.rs
+++ b/src/extensions/idle.rs
@@ -93,16 +93,26 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
     /// This is necessary so that we can keep using the inner `Session` in `wait_keepalive`.
     fn wait_inner(&mut self) -> Result<()> {
         let mut v = Vec::new();
-        match self.session.readline(&mut v).map(|_| ()) {
-            Err(Error::Io(ref e))
-                if e.kind() == io::ErrorKind::TimedOut || e.kind() == io::ErrorKind::WouldBlock =>
-            {
-                // we need to refresh the IDLE connection
-                self.terminate()?;
-                self.init()?;
-                self.wait_inner()
+        loop {
+            match self.session.readline(&mut v).map(|_| ()) {
+                Err(Error::Io(ref e))
+                    if e.kind() == io::ErrorKind::TimedOut
+                        || e.kind() == io::ErrorKind::WouldBlock =>
+                {
+                    // we need to refresh the IDLE connection
+                    self.terminate()?;
+                    self.init()?;
+                    return self.wait_inner();
+                }
+                r => r,
+            }?;
+
+            // Handle Dovecot's imap_idle_notify_interval message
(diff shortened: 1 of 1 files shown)


jonhoo:
Thank you for this! I only have a smaller nit, and am otherwise happy to merge :)

src/extensions/idle.rs
```diff
@@ -93,16 +93,26 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+                    // we need to refresh the IDLE connection
+                    self.terminate()?;
+                    self.init()?;
+                    self.wait_inner()
+                }
+                r => r,
+            };
+
+            // Handle Dovecot's imap_idle_notify_interval message
+            if res.is_ok() && v.eq_ignore_ascii_case(b"* OK Still here\r\n") {
```
jonhoo, line 111: I think it'd probably be better to use `?` on the `res` above. That way you know that you're in the `Ok` case, and wouldn't need to check for `is_ok()`. And then in the `else` case you'd just `break Ok(())`.

src/extensions/idle.rs
```diff
@@ -93,16 +93,26 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+        loop {
+            let res = match self.session.readline(&mut v).map(|_| ()) {
+                Err(Error::Io(ref e))
+                    if e.kind() == io::ErrorKind::TimedOut
+                        || e.kind() == io::ErrorKind::WouldBlock =>
+                {
+                    // we need to refresh the IDLE connection
+                    self.terminate()?;
+                    self.init()?;
+                    self.wait_inner()
```
jonhoo, line 105: I think this should probably be a `return self.wait_inner();`. At least then there's a hope that this might get tail-recursion optimized.

src/extensions/idle.rs
mordak: Oh yeah, that is better!

src/extensions/idle.rs
mordak: Thanks for pointing this out! I am changing the recursive path here.

mordak:
Thanks for reviewing! I updated with your feedback and pushed an update. If there's anything else, just let me know.

jonhoo:
This looks great, thanks! I assume you have also tested this patch in practice, and it works against Dovecot?

PR #165, opened 2020-05-19
Added: STARTTLS example

rvuong:
- [x] This PR adds an example file for STARTTLS connection.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/165)

### DIFF
--- a/examples/starttls.rs
+++ b/examples/starttls.rs
@@ -1,0 +1,68 @@
+/**
+ * Here's an example showing how to connect to the IMAP server with STARTTLS.
+ * The only difference with the `basic.rs` example is when using `imap::connect_starttls()` method
+ * instead of `imap::connect()` (l. 52)
+ *
+ * The following env vars are expected to be set:
+ * - IMAP_HOST
+ * - IMAP_USERNAME
+ * - IMAP_PASSWORD
+ * - IMAP_PORT (supposed to be 143)
+ */
+
+extern crate imap;
+extern crate native_tls;
+
+use native_tls::TlsConnector;
+use std::env;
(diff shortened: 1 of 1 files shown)


examples/starttls.rs
```diff
@@ -0,0 +1,68 @@
+/**
+ * Here's an example showing how to connect to the IMAP server with STARTTLS.
+ * The only difference with the `basic.rs` example is when using `imap::connect_starttls()` method
+ * instead of `imap::connect()` (l. 52)
```
jonhoo, line 4: Maybe add:
> and so you can connect on port 143 instead of 993 as you have to when using TLS the entire way.

jonhoo:
Ah, this is great, thank you!

It looks like [CI thinks](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1297&view=logs&j=a33723e4-26d4-5cee-460b-5f6fe51cc6d9&t=9d8409da-07aa-5b2d-2830-4ce5d7d1fab0&l=11) you need to run `cargo fmt`

examples/starttls.rs
rvuong: True

rvuong:
Sorry, I'm not familiar with `cargo test` yet. Would you explain me why this fails?
(tests on `master` also fail on my local environment)

jonhoo:
Ah, so, `cargo test` for this project assumes that you have an IMAP test server set up, so they probably won't pass locally.

The only test that failed on CI is for nightly, and that is due to a component not being available on the latest nightly. We can safely ignore that :)

PR #168, opened 2020-06-21
server greeting doc and timeout example

### LINKED ISSUE #167: imap::connect timeout
I have a specific server with a long connection time and would like to configure a timeout for that, specifically for the `imap::connect` function. 

From looking at the documentation and source code, my current understanding is that `imap::connect` uses [`TcpStream::connect`](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.connect). There is also an alternative [`TcpStream::connect_timeout`](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.connect_timeout). Would it be useful to either add a timeout function argument to `imap::connect` or to mimic `TcpStream` and an `imap::connect_timeout` alternative?

wookietreiber:
fixes #167

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/168)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -7,3 +7,4 @@ Examples:
   * basic - This is a very basic example of using the client.
   * gmail_oauth2 - This is an example using oauth2 for logging into gmail as a secure appplication.
   * rustls - This demonstrates how to use Rustls instead of Openssl for secure connections (helpful for cross compilation).
+  * timeout - This demonstrates how to use timeouts while connecting to an IMAP server.
--- a/examples/timeout.rs
+++ b/examples/timeout.rs
@@ -1,0 +1,92 @@
+extern crate imap;
+extern crate native_tls;
+
+use imap::Client;
+use native_tls::TlsConnector;
+use native_tls::TlsStream;
+use std::env;
+use std::error::Error;
+use std::fmt;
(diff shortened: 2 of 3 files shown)


jonhoo:
This is great, thank you! Although it looks like the tests are failing?

wookietreiber:
The docs were all messed up because I forgot to include all the ceremony for the doc tests to compile.

wookietreiber:
Now that those are gone, there seem to be some unrelated problems. Looks like something with macros. Don't know if I can do something about that, though, @jonhoo would you please take a look?

jonhoo:
Ah, interesting... I'm working to fix those in https://github.com/jonhoo/rust-imap/pull/169

jonhoo:
Could you try merging master now?

wookietreiber:
Rebased. Now, one of the FreeBSD tests segfaulted. It doesn't seem to show which one, unfortunately. It says: to rerun, pass `--lib` and I only changed docs and examples, which were successful.

@jonhoo Did the FreeBSD segfault also occur with the commits I rebased onto?

jonhoo:
The FreeBSD failure is, as far as I can tell, a separate issue. So I went ahead and merged, thanks!

PR #170, opened 2020-07-08
Re-export Result and Error types

timando:
Quite a few crate (serde_json, postgres, reqwest, std::io) re-export the error type from their error module, so I thought it would be good if this crate did as well.
I found myself trying to use imap::Error rather than imap::error::Error.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/170)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -87,7 +87,7 @@ mod client;
 pub use crate::client::*;
 
 pub mod error;
-pub use error::{Result,Error};
+pub use error::{Error, Result};
 
 pub mod extensions;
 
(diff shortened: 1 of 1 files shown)


timando:
one thing I'm not sure about is whether the re-exports should be `#[doc(inline)]` or not.

jonhoo:
This seems like a reasonable addition to me, thanks!
The test failure seems to be because you haven't run rustfmt over your code. Specifically:
```diff
-pub use error::{Result,Error};
+pub use error::{Error, Result};
```

PR #171, opened 2020-08-16
Add append_with_flags

mordak:
Allows a client to set flags on a new message as it is appended to a mailbox.

I have had this in my tree for some time in order to support setting flags on messages as they are added to a mailbox (eg. setting 'Seen' on messages that are appended to the Sent box by a mail client) and figured it would be useful to upstream.


mordak:
I am not sure what happened with the freebsd test. I have run the same tests as given in the .cirrus.yml on a FreeBSD 12.1 machine and they all pass. Happy to chase it down if it is continuously failing in your environment.

jonhoo:
Nice, let me take a look! You can ignore the FreeBSD test — it seems to be failing regularly.

src/client.rs
```diff
@@ -1078,10 +1078,29 @@ impl<T: Read + Write> Session<T> {
...
+
+    /// The [`APPEND` command](https://tools.ietf.org/html/rfc3501#section-6.3.11) can take
+    /// an optional FLAGS parameter to set the flags on the new message.
+    ///
+    /// > If a flag parenthesized list is specified, the flags SHOULD be set
+    /// > in the resulting message; otherwise, the flag list of the
+    /// > resulting message is set to empty by default.  In either case, the
+    /// > Recent flag is also set.
+    ///
+    /// Behaves like ['Session::append'], except takes a parenthesized list of flags.
```
jonhoo, line 1092: Can we just add the parentheses ourselves so the caller doesn't need to know about this detail?

src/client.rs
```diff
@@ -1078,10 +1078,29 @@ impl<T: Read + Write> Session<T> {
...
+    pub fn append_with_flags<S: AsRef<str>, B: AsRef<[u8]>, C: AsRef<str>>(
+        &mut self,
+        mailbox: S,
+        content: B,
+        flags: C,
```
jonhoo, line 1097: Instead of taking a single string, should we be taking a slice of `&str` instead so the caller does not need to know the exact format? Maybe we can even use the existing `Flag` type?

src/client.rs
mordak: Haha! Yes! The only reason I did it this way is for consistency with the rest of the API. For example, `store` takes a flag list that is a space separated list of flag strings in parentheses, `status` takes `data_items` that is a space separated list in parentheses, etc. I am happy to be inconsistent to save the caller from needing to know this detail if you're happy with it.

src/client.rs
mordak: Yes! I think the single string is a bit inconvenient, and a `Vec<Flag>` or `&[Flag]` would be better. What would you prefer? It is only this way for consistency with some of the other APIs, and I would be happy to change it to something nicer to use and harder to use wrong.

src/client.rs
jonhoo: Hmm, that is a good point.. From memory, that might have been done to support the list being empty (and therefore absent) without having a different call. But here that seems like something we probably wouldn't want to support since the "flagless" version already exists as its own method. I think not requiring the parentheses is fine here.

As for the comma-separated part, I suppose you're right that that would be very inconsistent with `store`. That is unfortunate indeed. But at the same time, I think it's valuable to start pushing in the right direction where we have the option to without a breaking change. So if you can make it work with a well-typed list, I think that's preferable :)

src/client.rs
jonhoo: `&[Flag]` I think is the way to go!

mordak:
Updated with taking `&[Flag]` instead of a string. Also ended up implementing Display for Flag as a side effect, which seems like a good knock-on effect. :-)

src/client.rs
```diff
@@ -1078,10 +1078,37 @@ impl<T: Read + Write> Session<T> {
...
+        flags: &[Flag<'_>],
+    ) -> Result<()> {
         let content = content.as_ref();
+        let flagstr = flags
+            .iter()
+            .filter(|f| **f != Flag::Recent)
+            .map(|f| f.to_string())
+            .collect::<Vec<String>>()
+            .join(" ");
+
```
jonhoo, line 1107: Should we check if the list is empty here, and if so call `append`?

src/types/mod.rs
```diff
@@ -169,6 +169,21 @@ impl Flag<'static> {
...
+impl<'a> fmt::Display for Flag<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match *self {
+            Flag::Seen => write!(f, "{}", "\\Seen"),
+            Flag::Answered => write!(f, "{}", "\\Answered"),
+            Flag::Flagged => write!(f, "{}", "\\Flagged"),
+            Flag::Deleted => write!(f, "{}", "\\Deleted"),
+            Flag::Draft => write!(f, "{}", "\\Draft"),
+            Flag::Recent => write!(f, "{}", "\\Recent"),
+            Flag::MayCreate => write!(f, "{}", "\\*"),
```
jonhoo, line 181: Is this really the right flag for `MayCreate`?!

jonhoo:
Nice! Left another few comments.

src/client.rs
mordak: I modified `append` to just call this function with an empty set of flags, so I think calling back into it if given an empty list would infinite loop?

If we take the `(` and `)` out of the format string given to APPEND and instead add them as part of `flagstr`, then we can make the empty list case omit the flags argument entirely, and then `append_with_flags`  will be exactly like what `append`  does now in the empty list case - the `flagstr` would just be `""` for the empty list, and `(Flag Flag2 Flag3)` for the not-empty case.

src/types/mod.rs
mordak: Yup. See PERMANENTFLAGS section here: https://tools.ietf.org/html/rfc3501#section-7.1

The `system` fn implementation for `Flag` here also reads this string as that flag:  https://github.com/jonhoo/rust-imap/blob/master/src/types/mod.rs#L166

I was also a bit surprised. :-)

src/client.rs
jonhoo: Oh, yeah, you're right, I misread!

jonhoo:
Ah, should have asked for this earlier — could you also add a simple integration test for this? Hopefully that shouldn't be _too_ hard to finagle under the current testing setup (it already starts an IMAP server for you).

mordak:
> Could you also add a simple integration test for this?

No problem. I thought you might ask. :-)

PR #172, opened 2020-11-11
Support HIGHESTMODSEQ and VANISHED

mordak:
This adds support for imap-proto `ResponseCode::HighestModSeq(u64)` and the `VANISHED` response described in RFC 7162 (QRESYNC). With these extensions, it is possible to synchronize a mailbox in a single round trip using `UID FETCH` and passing a modifier `(CHANGEDSINCE <MODSEQ> VANISHED)`. Only modifications and new messages since `<MODSEQ>` will be returned, and deleted messages will be returned in a `VANISHED` response on the `unsolicited_responses` channel.

Technically, if `QRESYNC` is enabled then `VANISHED` responses will be sent to `EXPUNGE` commands. Presently these will show up on the `unsolicited_responses` channel even though they are directly a response to a command. I wasn't sure what the best way to handle this would be in `parse_expunge()`, and commented to this effect in `parse_vanished_test()`, and the test is constructed in a way to verify that `parse_expunge()` succeeds under this condition, and the `VANISHED` response is available in the `unsolicited_responses` channel.

I also fixed clippy warnings. One of them I actually introduced (useless format string) - sorry. The other one is for the [manual_non_exhaustive lint](https://rust-lang.github.io/rust-clippy/master/index.html#manual_non_exhaustive). I separated them out into different commits so you can see them in isolation.

I am happy to revise this if you would like, particularly if you have a preference for what to do in `parse_expunge()`. I did experiment with expanding the `VANISHED` uid sequence-set out into a `Vec<u32>`, but it's kind of ugly, and also kind of dangerous for very large ranges, so I erred on the side of at least being consistent and having these responses always show up on the `unsolicited_responses` channel.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/172)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+ - VANISHED support in EXPUNGE responses and unsolicited responses (#172).
 
 ### Changed
+ - MSRV increased to 1.43 for nom6 and bitvec
+ - `expunge` and `uid_expunge` return `Result<Deleted>` instead of `Result<Vec<u32>>`.
 
 ### Removed
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -47,12 +47,12 @@ jobs:
  - job: msrv
    pool:
      vmImage: ubuntu-latest
-   displayName: "Minimum supported Rust version: 1.40.0"
+   displayName: "Minimum supported Rust version: 1.43.0"
    dependsOn: []
    steps:
      - template: install-rust.yml@templates
        parameters:
-         rust: 1.40.0 # static-assertions (1.37+) and base64 (1.40+)
+         rust: 1.43.0 # nom6 depends on bitvec (1.43+)
      - script: cargo check
        displayName: cargo check
  - job: integration
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -158,6 +158,10 @@ pub fn parse_expunge(
         }
     }
 
+    // If the server sends a VANISHED response then they must only send VANISHED
+    // in lieu of EXPUNGE responses for the rest of this connection, so it is
+    // always one or the other.
+    // https://tools.ietf.org/html/rfc7162#section-3.2.10
     if !vanished.is_empty() {
         Ok(Deleted::from_vanished(vanished))
     } else {
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -44,7 +44,7 @@ where
                 break Ok(things);
             }
 
-            match imap_proto::parse_response(lines) {
+            match imap_proto::parser::parse_response(lines) {
                 Ok((rest, resp)) => {
                     lines = rest;
 
@@ -143,7 +143,7 @@ pub fn parse_capabilities(
     let f = |mut lines| {
         let mut caps = HashSet::new();
         loop {
-            match imap_proto::parse_response(lines) {
+            match imap_proto::parser::parse_response(lines) {
                 Ok((rest, Response::Capabilities(c))) => {
                     lines = rest;
                     caps.extend(c);
@@ -179,7 +179,7 @@ pub fn parse_noop(
             break Ok(());
         }
 
-        match imap_proto::parse_response(lines) {
+        match imap_proto::parser::parse_response(lines) {
             Ok((rest, data)) => {
                 lines = rest;
                 if let Some(resp) = handle_unilateral(data, unsolicited) {
@@ -200,7 +200,7 @@ pub fn parse_mailbox(
     let mut mailbox = Mailbox::default();
 
     loop {
-        match imap_proto::parse_response(lines) {
+        match imap_proto::parser::parse_response(lines) {
             Ok((rest, Response::Data { status, code, .. })) => {
                 lines = rest;
 
@@ -212,6 +212,9 @@ pub fn parse_mailbox(
 
                 use imap_proto::ResponseCode;
                 match code {
+                    Some(ResponseCode::HighestModSeq(seq)) => {
+                        mailbox.highest_mod_seq = Some(seq);
+                    }
                     Some(ResponseCode::UidValidity(uid)) => {
                         mailbox.uid_validity = Some(uid);
                     }
@@ -254,7 +257,8 @@ pub fn parse_mailbox(
                     }
                     MailboxDatum::List { .. }
                     | MailboxDatum::MetadataSolicited { .. }
-                    | MailboxDatum::MetadataUnsolicited { .. } => {}
+                    | MailboxDatum::MetadataUnsolicited { .. }
+                    | MailboxDatum::Search { .. } => {}
                 }
             }
             Ok((rest, Response::Expunge(n))) => {
@@ -286,8 +290,8 @@ pub fn parse_ids(
             break Ok(ids);
         }
 
-        match imap_proto::parse_response(lines) {
-            Ok((rest, Response::IDs(c))) => {
+        match imap_proto::parser::parse_response(lines) {
+            Ok((rest, Response::MailboxData(MailboxDatum::Search(c)))) => {
                 lines = rest;
                 ids.extend(c);
             }
@@ -330,6 +334,11 @@ fn handle_unilateral<'a>(
         }
         Response::Expunge(n) => {
             unsolicited.send(UnsolicitedResponse::Expunge(n)).unwrap();
+        }
+        Response::Vanished { earlier, uids } => {
+            unsolicited
+                .send(UnsolicitedResponse::Vanished { earlier, uids })
+                .unwrap();
         }
         res => {
             return Some(res);
@@ -546,5 +555,60 @@ mod tests {
         assert!(recv.try_recv().is_err());
         let ids: HashSet<u32> = ids.iter().cloned().collect();
         assert_eq!(ids, HashSet::<u32>::new());
+    }
+
+    #[test]
+    fn parse_vanished_test() {
+        // VANISHED can appear if the user has enabled QRESYNC (RFC 7162), in response to
+        // SELECT/EXAMINE (QRESYNC); UID FETCH (VANISHED); or EXPUNGE commands. In the first
+        // two cases the VANISHED respone will be a different type than the expected response
+        // and so goes into the unsolicited respones channel. In the last case, VANISHED is
+        // explicitly a response to an EXPUNGE command, but the semantics of EXPUNGE (one ID
+        // per response, multiple responses) vs VANISHED (a sequence-set of UIDs in a single
+        // response) are different enough that is isn't obvious what parse_expunge() should do.
+        // If we do nothing special, then the VANISHED response ends up in the unsolicited
+        // responses channel, which is at least consistent with the other cases where VANISHED
+        // can show up.
+        let lines = b"* VANISHED 3\r\n";
+        let (mut send, recv) = mpsc::channel();
+        let resp = parse_expunge(lines.to_vec(), &mut send).unwrap();
+        assert!(resp.is_empty());
+
+        match recv.try_recv().unwrap() {
+            UnsolicitedResponse::Vanished { earlier, uids } => {
+                assert!(!earlier);
+                assert_eq!(uids.len(), 1);
+                assert_eq!(*uids[0].start(), 3);
+                assert_eq!(*uids[0].end(), 3);
+            }
+            what => panic!("Unexpected response in unsolicited responses: {:?}", what),
+        }
+        assert!(recv.try_recv().is_err());
+
+        // test VANISHED mixed with FETCH
+        let lines = b"* VANISHED (EARLIER) 3:8,12,50:60\r\n\
+                      * 49 FETCH (UID 117 FLAGS (\\Seen \\Answered) MODSEQ (90060115194045001))\r\n";
+
+        let fetches = parse_fetches(lines.to_vec(), &mut send).unwrap();
+        match recv.try_recv().unwrap() {
+            UnsolicitedResponse::Vanished { earlier, uids } => {
+                assert!(earlier);
+                assert_eq!(uids.len(), 3);
+                assert_eq!(*uids[0].start(), 3);
+                assert_eq!(*uids[0].end(), 8);
+                assert_eq!(*uids[1].start(), 12);
+                assert_eq!(*uids[1].end(), 12);
+                assert_eq!(*uids[2].start(), 50);
+                assert_eq!(*uids[2].end(), 60);
+            }
+            what => panic!("Unexpected response in unsolicited responses: {:?}", what),
+        }
+        assert!(recv.try_recv().is_err());
+        assert_eq!(fetches.len(), 1);
+        assert_eq!(fetches[0].message, 49);
+        assert_eq!(fetches[0].flags(), &[Flag::Seen, Flag::Answered]);
+        assert_eq!(fetches[0].uid, Some(117));
+        assert_eq!(fetches[0].body(), None);
+        assert_eq!(fetches[0].header(), None);
     }
 }
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -127,13 +127,42 @@ pub fn parse_fetches(
 pub fn parse_expunge(
     lines: Vec<u8>,
     unsolicited: &mut mpsc::Sender<UnsolicitedResponse>,
-) -> Result<Vec<u32>> {
-    let f = |resp| match resp {
-        Response::Expunge(id) => Ok(MapOrNot::Map(id)),
-        resp => Ok(MapOrNot::Not(resp)),
-    };
+) -> Result<Deleted> {
+    let mut lines: &[u8] = &lines;
+    let mut expunged = Vec::new();
+    let mut vanished = Vec::new();
 
-    unsafe { parse_many(lines, f, unsolicited).map(|ids| ids.take()) }
+    loop {
+        if lines.is_empty() {
+            break;
+        }
+
+        match imap_proto::parser::parse_response(lines) {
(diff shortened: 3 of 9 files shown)


jonhoo:
Thank you for this, those are good changes!

Sadly, these changes are also backwards-incompatible, and would require a major version bump (see #163 and https://github.com/jonhoo/rust-imap/pull/149#discussion_r426815938), so I may have to hold off on merging it for a little while. I'll leave some other comments in the meantime though!

As for `parse_expunge`, I'm not sure what the right way to go about it is either. I _think_ having the `VANISHED` responses go through unilateral responses is probably fine. It's not ideal as far as the API goes, but at least it makes it _possible_ for clients to do what they need to.

src/types/mailbox.rs
```diff
@@ -35,6 +35,10 @@ pub struct Mailbox {
     /// The unique identifier validity value.  See [`Uid`] for more details.  If this is missing,
     /// the server does not support unique identifiers.
     pub uid_validity: Option<u32>,
+
+    /// The highest mod sequence for this mailboxr. Used with
+    /// [Conditional STORE](https://tools.ietf.org/html/rfc4551#section-3.1.1).
+    pub highest_mod_seq: Option<u64>,
```
jonhoo, line 41: We should probably add `#[non_exhaustive]` to `Mailbox` as well so that we can make these kinds of changes in the future without a breaking change.

src/types/mailbox.rs
```diff
@@ -35,6 +35,10 @@ pub struct Mailbox {
     /// The unique identifier validity value.  See [`Uid`] for more details.  If this is missing,
     /// the server does not support unique identifiers.
     pub uid_validity: Option<u32>,
+
+    /// The highest mod sequence for this mailboxr. Used with
```
jonhoo, line 39:
```suggestion
    /// The highest mod sequence for this mailbox. Used with
```

src/types/mailbox.rs
mordak: Ah, that is a good idea! I added this change.

mordak:
> Sadly, these changes are also backwards-incompatible, and would require a major version bump (see #163 and [#149 (comment)](https://github.com/jonhoo/rust-imap/pull/149#discussion_r426815938)), so I may have to hold off on merging it for a little while. I'll leave some other comments in the meantime though!

Yup, no problem. If you are looking for an excuse to take the plunge on the major bump, then this is the only outstanding diff before I can publish [IMAP <-> Maildir sync tool](https://github.com/mordak/runt). It has been a private project for a long time though, so I am not in any rush. :-)

> As for `parse_expunge`, I'm not sure what the right way to go about it is either. I _think_ having the `VANISHED` responses go through unilateral responses is probably fine. It's not ideal as far as the API goes, but at least it makes it _possible_ for clients to do what they need to.

Yeah. If `parse_expunge` keeps the current signature, then it's expanding the `VANISHED` UID sequence-set out into a `Vec<u32>`. Otherwise, the signature changes to return an enum type with a `Vec<u32>` or `Vec<RangeInclusive>`; or the signature changes to return something else. I think the enum type is the most correct way to do it (they _are_ different things after all, `id` vs `uid` values, etc.), but it doesn't quite feel right? Of course, if you have a preference, then getting it in at the same time you're bumping the major anyway would be nice.

mordak:
> As for `parse_expunge`, I'm not sure what the right way to go about it is either. I _think_ having the `VANISHED` responses go through unilateral responses is probably fine. It's not ideal as far as the API goes.

Thinking about this some more, I think the nicest API for the client is for `client::expunge()` and `client::uid_expunge()` to return something that they can iterate easily. So if those returned (say) `Result<Expunged>` and `Expunged` was an enum (or something) that provided an iterator interface, then the client could do something nice like `for ids in client.expunge()? { ... }`. If the iterator just yielded `u32`s, then some clients might not even notice, since you can do the exact same thing with a `Vec`.

I am happy to work it out if you have a preference. Since I'm the one who wants `VANISHED` support, I should do the legwork to ensure it fits in the API nicely for you.

jonhoo:
> Yup, no problem. If you are looking for an excuse to take the plunge on the major bump, then this is the only outstanding diff before I can publish my [IMAP <-> Maildir sync tool](https://github.com/mordak/runt). It has been a private project for a long time though, so I am not in any rush. :-)

That's really cool! That definitely does give me a bit of a push :p

I like your idea of returning just "something iterable" instead — that does seem a lot nicer. Alternatively it could return a `Vec<enum>` where the `enum` is either a single id or a sequence-set?

mordak:
> I like your idea of returning just "something iterable" instead — that does seem a lot nicer. Alternatively it could return a `Vec<enum>` where the `enum` is either a single id or a sequence-set?

Yup, `Vec<enum>` would also work. I think the caller would have to `match` on the enum though to get at the values? The upside of this way is the caller will know if they're getting `ID` or `UID` values, because they would explicitly know if they're dealing with an `EXPUNGE` response (ids) or a `VANISHED` response (uids). I think there is value in making this distinction to the caller - `UID` and `ID` have different semantics (all the business around IDs decrementing between responses like in the example [in RFC4315](https://tools.ietf.org/html/rfc4315#section-2.1) is stuff the caller has to deal with).

I will mock something up and shoot you a diff. If you like it, I can merge it into this PR. If you don't, then we can hammer it out. :-)

mordak:
> I like your idea of returning just "something iterable" instead — that does seem a lot nicer. Alternatively it could return a `Vec<enum>` where the `enum` is either a single id or a sequence-set?

Pushed a commit with an implementation of a `Deleted` type that is returned from `parse_expunge()`.

The client can just match on the enum if they want to get at the underlying `Vec<Seq>` or `Vec<RangeInclusive<Uid>>`, but I also provided a few iterator conveniences and some examples for how they can be used. There's 3 ways to iterate:
- `Deleted::into_iter()`. So `for id in &deleted { } ` works and the client doesn't care what the numbers are (sequence numbers or UIDs). I figure this is probably compatible with what most users are probably already doing?
- `Deleted::seqs()`, if the client knows / expects to get an `EXPUNGE` response.
- `Deleted::uids()`, if the client knows / expects to get a `VANISHED` response.

Let me know what you think. I am happy to modify / rework if needed.

src/parse.rs
```diff
@@ -561,31 +590,22 @@ mod tests {
...
-                assert_eq!(*uids[0].start(), 3);
-                assert_eq!(*uids[0].end(), 3);
-            }
-            what => panic!("Unexpected response in unsolicited responses: {:?}", what),
-        }
+        // Should have one UID response
+        assert!(!resp.is_empty());
+        let mut uids = resp.uids().unwrap();
+        assert_eq!(Some(3), uids.next());
+        assert_eq!(None, uids.next());
```
jonhoo, line 603: Should this also assert that `seqs()` is empty?

src/types/deleted.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// returned in a `VANISHED` response
+    pub fn from_vanished(v: Vec<RangeInclusive<u32>>) -> Self {
+        Deleted::Vanished(v)
+    }
+
+    /// Return an iterator over message sequence numbers from an `EXPUNGE`
+    /// response. If the client is expecting sequence numbers this function
+    /// can be used to ensure only sequence numbers returned in an `EXPUNGE`
+    /// response are processed.
+    pub fn seqs(&'a self) -> Option<DeletedIterator<'a>> {
```
jonhoo, line 66: I think you could probably make this and `uids` much nicer by returning `impl ExactSizeIterator<Item = u32>`. That way you can get rid of all the various `struct` types and `impl Iterator for` blocks below, and just use combinators instead (`flat_map` will probably come in handy for `uids` for example).

I also don't think the `Option` adds much here. `ExactSizeIterator` is good enough since it also provides an `is_empty` method.

jonhoo:
This looks great — I think having the one shared type is a great improvement. I left a note on how you may be able to (significantly) simplify the implementation though!

src/parse.rs
mordak: Ah, yes!

src/types/deleted.rs
mordak: Oh, this is much better - thanks! I had not thought to do it this way.

I pushed a new version with `seqs()` and `uids()` returning `impl Iterator<Item = u32>`. `ExactSizeIterator` isn't implemented for `FlatMap` (which makes sense, I suppose), so I used `Iterator` instead. I could use `ExactSizeIterator` for `seqs()`, but wanted to keep them consistent.

As for the `Option` - I included it to let the caller distinguish the 'no values' from the 'didn't get that kind of response' cases, but wasn't sure if this was a distinction that added enough value to justify the `Option`, so am glad you have an opinion! I removed it in this version, though `Iterator` doesn't provide `is_empty()` like `ExactSizeIterator` does, but I am not entirely convinced this is needed? `Deleted` has `is_empty()` anyway. If you think this is important I can figure it out.

Thanks for the comments - I found all the structs a bit cumbersome, but didn't think to just return an `impl`. Being able to delete all that makes it much simpler / nicer.

mordak: Also, TIL `flat_map()` is a thing. That is much nicer than rolling it by hand!

src/parse.rs
```diff
@@ -127,13 +127,42 @@ pub fn parse_fetches(
...
+                }
+            }
+            _ => {
+                return Err(Error::Parse(ParseError::Invalid(lines.to_vec())));
+            }
+        }
+    }
 
-    unsafe { parse_many(lines, f, unsolicited).map(|ids| ids.take()) }
+    if !vanished.is_empty() {
```
jonhoo, line 161: Can there be both vanished and expunged results at once?

src/types/deleted.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// returned in a `VANISHED` response
+    pub fn from_vanished(v: Vec<RangeInclusive<u32>>) -> Self {
+        Deleted::Vanished(v)
+    }
+
+    /// Return an iterator over message sequence numbers from an `EXPUNGE`
+    /// response. If the client is expecting sequence numbers this function
+    /// can be used to ensure only sequence numbers returned in an `EXPUNGE`
+    /// response are processed.
+    pub fn seqs(&self) -> impl Iterator<Item = u32> + '_ {
```
jonhoo, line 64: This should probably use the `Seq` type alias to signify that it specifically returns sequence numbers.

src/types/deleted.rs
```diff
@@ -0,0 +1,180 @@
...
+            Deleted::Expunged(s) => s.iter(),
+            Deleted::Vanished(_) => [].iter(),
+        }
+        .copied()
+    }
+
+    /// Return an iterator over UIDs returned in a `VANISHED` response.
+    /// If the client is expecting UIDs this function can be used to ensure
+    /// only UIDs are processed.
+    pub fn uids(&self) -> impl Iterator<Item = u32> + '_ {
```
jonhoo, line 75: This should probably use the `Uid` type alias to signify that it specifically returns UIDs.

src/types/deleted.rs
```diff
@@ -0,0 +1,180 @@
...
+        match self {
+            Deleted::Expunged(_) => [].iter(),
+            Deleted::Vanished(s) => s.iter(),
+        }
+        .flat_map(|range| range.clone())
```
jonhoo, line 80: It doesn't seem quite right for this to need to clone the range. I think `iter().copied()` should do the trick?

jonhoo:
Ah, yes, this is looking much nicer, thank you! I left another couple of notes from looking through more carefully.

I'm not quite sure why the Minimum supported Rust version test is failing. Can you try bumping that to 1.41 here
https://github.com/jonhoo/rust-imap/blob/7868e312ff1349a667541b235053fc5c95c94854/azure-pipelines.yml#L50
and here
https://github.com/jonhoo/rust-imap/blob/7868e312ff1349a667541b235053fc5c95c94854/azure-pipelines.yml#L55

Also, while you're at it, could you please update the changelog to reflect both the Rust version change and the changes in this PR?

src/parse.rs
mordak: According to the spec, no. Once the client has enabled QRESYNC then the server must send only VANISHED responses in lieu of EXPUNGE responses.

https://tools.ietf.org/html/rfc7162#section-3.2.10

src/types/deleted.rs
mordak: You're right, fixed!

src/types/deleted.rs
mordak: Fixed!

src/types/deleted.rs
mordak: I actually fiddled with this a bunch and can't get it to work without the `clone()`. `RangeInclusve` is not `Copy`, and doesn't implement `iter()` because it _is_ the iterator. We can't just iterate the one in the `Vec` because it means mutating the `Vec` members in place (ie. [`next()` mutates the `RangeInclusive`](https://doc.rust-lang.org/src/core/iter/range.rs.html#649)).  So inside `flat_map()` we're expected to yield an iterator from the lambda, but because we cannot mutate the `RangeInclusive` in the `Vec` directly, we need a new iterator to return inside `flat_map()`, which means making a new one from the given `RangeInclusive`, which is a `clone()`. 

But I think this is really not a big deal. `RangeInclusive<u32>` is tiny (a pair of numbers and some state), so cloning it is cheap, and it means we do not mutate the members of the `Delete`, so we can iterate it more than once if we want, and I think that's the right behaviour to have.

mordak:
> Ah, yes, this is looking much nicer, thank you! I left another couple of notes from looking through more carefully.

Thanks! Your comments are really helpful.

> I'm not quite sure why the Minimum supported Rust version test is failing. Can you try bumping that to 1.41 here

Figured this out. It is because nom6 brings in a new version of bitvec, which has a minimum version of 1.44. I tested it with 1.43 and it works, so I set that as the minimum version. If you want to go to 1.44 then that's easy - it works with every version between 1.43 and the latest stable.

We actually do not _need_ the bump to nom6. I only did it because imap-proto did it, and it makes it possible to work on a local imap-proto while we're adding features to it and plumbing them through this crate. So when imap-proto went to nom6, I bumped this crate to nom6 so I could work on them in tandem (via `[patch.crates-io]`). I can revert it to nom5 again if you would prefer and I think this would let the minimum version stay at 1.40, but I am not sure if this is a priority for you. I totally get if it is, so just say so and I can figure it out.

> Also, while you're at it, could you please update the changelog to reflect both the Rust version change and the changes in this PR?

Done! Let me know if I messed it up or was too brief / verbose / anything. :-)

CHANGELOG.md
```diff
@@ -6,8 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+ - VANISHED support in EXPUNGE responses and unsolicited responses (#172).
 
 ### Changed
+ - MSRV increased
```
jonhoo, line 12: Let's say what it increased to :)

azure-pipelines.yml
```diff
@@ -47,12 +47,12 @@ jobs:
...
    pool:
      vmImage: ubuntu-latest
-   displayName: "Minimum supported Rust version: 1.40.0"
+   displayName: "Minimum supported Rust version: 1.43.0"
    dependsOn: []
    steps:
      - template: install-rust.yml@templates
        parameters:
-         rust: 1.40.0 # static-assertions (1.37+) and base64 (1.40+)
+         rust: 1.43.0 # static-assertions (1.37+) and base64 (1.40+)
```
jonhoo, line 55: Could you replace the comment here too now that the reason for the MSRV is different.

jonhoo:
I'm very happy to update to nom6, so please do keep that change!

The changes to the changelog look good -- I only left some minor notes and then I think this PR is basically ready :tada: Now I just need to get ready for a new major release :sweat_smile:

CHANGELOG.md
mordak: Done!

azure-pipelines.yml
mordak: No problem! :-)

mordak:
> The changes to the changelog look good -- I only left some minor notes and then I think this PR is basically ready 🎉 Now I just need to get ready for a new major release 😅

Great! If there is anything I can help with then let me know, or if you need additional changes here then just say the word.

Thanks for your time and effort helping me get this into shape! :-)

src/parse.rs
jonhoo: Ah, excellent! Could you please add a comment (with that link) pointing out that fact?

jonhoo:
Oh, just a note for the future: in general it's best to avoid force pushing during reviews, since GitHub has a hard time keeping track of what changed when you do that. It tends to be easier to do commit cleanup just at the very end instead :)

I missed one last thing that I left as a comment now, and then I think I'm done. Sorry to keep adding more nits — thank you for sticking with me through that!

src/parse.rs
mordak: Done!

mordak:
> Oh, just a note for the future: in general it's best to avoid force pushing during reviews, since GitHub has a hard time keeping track of what changed when you do that. It tends to be easier to do commit cleanup just at the very end instead :)

Oh, sorry. :( 

Thank you for saying this! I didn't know this  made reviews harder, and am used to force pushing to keep the commit history tidy. I didn't force the last one here. :-)

Incidentally, if you want some of the commits merged / shuffled / anything, just let me know and I can wrangle it.

> I missed one last thing that I left as a comment now ([#172 (comment)](https://github.com/jonhoo/rust-imap/pull/172#discussion_r537118865)), and then I think I'm done. Sorry to keep adding more nits — thank you for sticking with me through that!

No problem at all! It's your crate and your standards, and you're the one with the merge button. Don't feel bad for nits - if I don't fix them now you're just going to feel compelled to fix them later. :-)

jonhoo (approved):

PR #173, opened 2020-12-06
CI: Only install components when needed.

mordak:
Nightly pipeline is failing on missing rustfmt, but we don't actually need it except for the beta version.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/173)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,18 +6,23 @@ jobs:
        Linux-stable:
          vmImage: ubuntu-latest
          rust: stable
+         components: []
        Linux-beta:
          vmImage: ubuntu-latest
          rust: beta
+         components: [rustfmt, clippy]
        Linux-nightly:
          vmImage: ubuntu-latest
          rust: nightly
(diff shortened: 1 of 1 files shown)


mordak:
Feel free to ignore if you're not interested.

I noticed that the CI pipelines will fail when either clippy or rustfmt can't be installed on nightly, and this means you have to go look at why the pipeline failed on PRs and that's probably a bit annoying when it's just that rustfmt isn't available for nightly today. But the pipeline only actually uses clippy and rustfmt on the 'beta' toolchain, so missing those components on nightly is irrelevant.

So this moves the install of the rustfmt and clippy components into the pipeline step that needs them. I originally tried doing this by specifying the components with the matrix elements, but Azure variables can't be arrays so that doesn't work.

Anyway, not sure if this is something you want, but if so maybe it's handy for you.

jonhoo (approved):

jonhoo:
Oh, yes, this is great, thank you!

Issue #60, opened 2018-11-23
APPEND support is incomplete


jonhoo:
> At the moment `APPEND` only supports setting the body of the message. However, RFC3501 allows users to specify the flag set to be applied to a message, as well as the date the message was received. Quoting the specification:
> 
> ```
> 6.3.11. APPEND Command
> 
>    Arguments:  mailbox name
>                OPTIONAL flag parenthesized list
>                OPTIONAL date/time string
>                message literal
> ```
> 
> We could change the `append` method to:
> 
> ```rust
> use chrono::DateTime;
> use chrono::offset::Utc;
> 
> fn append(&mut self, folder: &str,
>           flags: Vec<String>,
>           datetime: DateTime<Utc>,
>           message: &[u8]) { /* ... */ } 
> ```

See the original issue here: https://github.com/mattnenterprise/rust-imap/issues/60

cyphar:
mattnenterprise/rust-imap#61 was my PoC for this.

mblarsen:
Would love to see the date feature implemented. Should I try to revise the old PR?

Any thoughts on the api? Keeping append as is and add append_with_flags_and_date to avoid breaking anything?

PR #174, opened 2020-12-12
feat: allow setting sent date on APPEND

mblarsen:
References issue #60

Followed syntax of (date-time token) http://www.faqs.org/rfcs/rfc3501.html

I followed the pattern of `append` calling `append_with_flags` calling `append_with_flags_and_date`.

- [x] existing tests passes
- [x] update fn docs

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/174)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1122,7 +1122,7 @@ impl<T: Read + Write> Session<T> {
         mailbox: S,
         content: B,
         flags: &[Flag<'_>],
-        date: Option<DateTime<FixedOffset>>,
+        date: impl Into<Option<DateTime<FixedOffset>>>,
     ) -> Result<()> {
         let content = content.as_ref();
         let flagstr = flags
@@ -1131,7 +1131,7 @@ impl<T: Read + Write> Session<T> {
             .map(|f| f.to_string())
             .collect::<Vec<String>>()
             .join(" ");
-        let datestr = match date {
+        let datestr = match date.into() {
             Some(date) => format!(" \"{}\"", date.format("%d-%h-%Y %T %z")),
             None => "".to_string(),
         };
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,5 +1,6 @@
 use base64;
 use bufstream::BufStream;
+use chrono::{DateTime, FixedOffset};
 #[cfg(feature = "tls")]
 use native_tls::{TlsConnector, TlsStream};
 use nom;
@@ -1121,7 +1122,7 @@ impl<T: Read + Write> Session<T> {
         mailbox: S,
         content: B,
         flags: &[Flag<'_>],
-        date: Option<chrono::NaiveDateTime>,
+        date: Option<DateTime<FixedOffset>>,
     ) -> Result<()> {
         let content = content.as_ref();
         let flagstr = flags
@@ -1131,7 +1132,7 @@ impl<T: Read + Write> Session<T> {
             .collect::<Vec<String>>()
             .join(" ");
         let datestr = match date {
-            Some(date) => format!(" \"{} +0000\"", date.format("%d-%h-%Y %T")),
+            Some(date) => format!(" \"{}\"", date.format("%d-%h-%Y %T %z")),
             None => "".to_string(),
         };
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -1097,6 +1097,26 @@ impl<T: Read + Write> Session<T> {
         content: B,
         flags: &[Flag<'_>],
     ) -> Result<()> {
+        self.append_with_flags_and_date(mailbox, content, flags, None)
+    }
+
+    /// The [`APPEND` command](https://tools.ietf.org/html/rfc3501#section-6.3.11) can take
+    /// an optional FLAGS parameter to set the flags on the new message.
+    ///
+    /// > If a flag parenthesized list is specified, the flags SHOULD be set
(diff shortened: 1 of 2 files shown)


mblarsen:
I tested the implementation on a little tool that migrates a massive mbox file through imap on fastmail.com. Emails appear with expected times.

There's no test for append so I was lazy and didn't add to it.

So yeah, ready for review @jonhoo :)

jonhoo:
Excellent, thank you! I left one small inline comment, but that's the only thing. It would be _very_ nice to have an append test, but I'm not going to require it :p

src/client.rs
```diff
@@ -1104,11 +1130,16 @@ impl<T: Read + Write> Session<T> {
             .map(|f| f.to_string())
             .collect::<Vec<String>>()
             .join(" ");
+        let datestr = match date {
+            Some(date) => format!(" \"{} +0000\"", date.format("%d-%h-%Y %T")),
```
jonhoo, line 1134: It seems odd to enforce a timezone of `+0000` here? I think I'd prefer for this method to take a normal `DateTime` (i.e., with timezone information), and then for that timezone to also be used in the datestr (in which case I think you can just use `date.format` and don't need the `format!` as well.

mblarsen: Yeah, it makes sense to switch to DateTime.

Regarding the use of `format!` is to swallow the extra space and double quotes in case of `None` for date. I know that they can be included in the date.format I just opted for clarity on what the date format is vs how it fits on the APPEND line.

mblarsen: I've updated it now with `DateTime<Utc>`. I'm really new to Rust so still getting the hang of the basic tool belt and language features. I'm guessing that `Utc` should not be hard-coded but specified with generics? E.g. 

```rust
    pub fn append_with_flags_and_date<S: AsRef<str>, B: AsRef<[u8]>, Z: TimeZone>(
        &mut self,
        mailbox: S,
        content: B,
        flags: &[Flag<'_>],
        date: Option<DateTime<Z>>,
    ) -> Result<()> {
```

However, this trickles up the chain to the other append methods in a way I don't quite understand yet:

image

mblarsen: I tried a bunch of variants as `Option<DateTime<TimeZone<offset = O>>>` (and one with `dyn` as well) but cannot seem to get it right.

mblarsen: Changed DateTime<Local> to DateTime<FixedOffset>

Local takes the timezone from the system. Utc sets to +0000 we seem to
be wanting one that remains fixed so that when formatted it is a
specified.

mblarsen:
> Excellent, thank you! I left one small inline comment, but that's the only thing. It would be _very_ nice to have an append test, but I'm not going to require it :p

Oh, my mistake hadn't seen the integration tests. I've added a test for the `_and_date` variant now. 

However, I had to disable it for now since the date returned from Greenmail is the time of the append not the date given. The date you specify with APPEND is the internal date as far as I know. I'm not sure how to best test this as the returned envelop would have the sent date not the internal date.

I guess I could send to emails first one without `_and_date` then one with that has a date in the past, then do a fetch of them. I would assume the latter would be the first if it works.

Any thoughts?

mblarsen:
Greenmail seem to be doing things by the book: 
https://github.com/greenmail-mail-test/greenmail/blob/567046758010524c7382393fc0f740b8b4073e08/greenmail-core/src/main/java/com/icegreen/greenmail/imap/commands/AppendCommand.java#L42-L67

src/client.rs
jonhoo: `date: Option<impl Into(DateTime<FixedOffset>)>` sounds great to me! Though I think you can move the `Into` one level out:
```rust
date: impl Into<Option<DateTime<FixedOffset>>>
```

jonhoo:
That's interesting -- if it doesn't modify the actual message date, how then is `append` with a date useful in practice? Can you extract it from the other fields of the message that aren't in the envelope?

Also, I had another thought over the weekend: I think this append API is getting pretty unwieldy. Given that we're going to be doing a breaking release soon anyway, I suggest we just make `append` be a single function that takes `impl Into<AppendOptions>` where
```rust
struct AppendOptions<'a> {
    flags: Option<&'a [Flag<'a>]>,
    date: Option<DateTime<FixedOffset>>,
}
```
And then have a couple of nice `Into` implementations for `AppendOption`. What do you think?

mblarsen:
> That's interesting -- if it doesn't modify the actual message date, how then is append with a date useful in practice? Can you extract it from the other fields of the message that aren't in the envelope?

Yes, that was it. `Fetch` has `internal_date` and with that in play the tests pass.

`date` updated to `Into<Option<...>>` as you suggested.

---

The `AppendOption` change makes sense to me. I can make that in a separate PR as it would break the current API as far as I can tell. 

You also mentioned (somewhere) that you take the path of a builder. Would the you then create an AppendCmd that implements `AppendOption` so that user can use the builder or `AppendOption` directly?

```
session.append(
  mbox,
  e.message_to_string().unwrap(),
  AppendCmd::new()
     .flag("\Seen")
     .internal_date(&date)
);
```

Or did you think to have a more generic command builder that was fed to something like `session.execute(..)`?

jonhoo (approved):

jonhoo:
Ah, yes, `AppendCmd` looks like a great idea! And a separate PR makes sense :)

PR #175, opened 2020-12-16
refactor: squash append command

mblarsen:
This is an attempt to simplify append by squashing the three variants into to one command + an options struct `AppendOptions`.

This PR also provides a command builder `AppendCmd` to simplify construction of the options struct as [suggested here](https://github.com/jonhoo/rust-imap/pull/174#issuecomment-745000597) and onwards:

```rust
session.append(
  mbox,
  e.message_to_string().unwrap(),
  AppendCmd::new()
     .flag(Flag::Seen)
     .internal_date(&date)
);
```

- [x] tests
- [x] update docs

---

Thoughts not related to what is implemented:

- If `AppendCmd` builds options maybe it should be called `AppendOptionsBuilder` (very long)

- and if so what `AppendCmd` look like:

```rust
struct AppendCmd {
  mbox: ..,
  content: ..,
  // impl
  pub fn run(&self) -> ...;
  pub fn with_options(&self) -> AppendOptionsBuilder {
     AppendOptionsBuilder {
         cmd: &self,
         flags: ...
         internal_date: ...
     }
     // impl
     fn run(&self) {
         self.cmd.run()
     }
  } 
}

session.append(mbox, e.message_to_string().unwrap()) |> AppendCmd
  .with_options() |> AppendOptionsBuilder
  .flag(Flag::Seen) |> AppendOptionsBuilder
  .internal_date(date) |> AppendOptionsBuilder
  .run()
```

Or maybe just combine them into to one. Compared to the existing api aside from the conveinince of builders there is an extra call to `run()`

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/175)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -85,6 +85,7 @@ pub struct Connection<T: Read + Write> {
 }
 
 /// A builder for the append command
+#[must_use]
 pub struct AppendCmd<'a, T: Read + Write> {
     session: &'a mut Session<T>,
     content: &'a [u8],
@@ -126,7 +127,6 @@ impl<'a, T: Read + Write> AppendCmd<'a, T> {
     }
 
     /// Run command
-    #[must_use = "always run a command once options are configured"]
     pub fn run(&mut self) -> Result<()> {
         let flagstr = self
             .flags
--- a/src/client.rs
+++ b/src/client.rs
@@ -93,6 +93,41 @@ pub struct AppendOptions<'a> {
     pub date: Option<DateTime<FixedOffset>>,
 }
 
+/// A builder for the append command
+#[derive(Default)]
+pub struct AppendCmd<'a> {
+    flags: Vec<&'a Flag<'a>>,
+    date: Option<DateTime<FixedOffset>>,
+}
+
+impl<'a> AppendCmd<'a> {
+    /// Create a new AppendCmd builder
+    pub fn create() -> Self {
+        Self::default()
+    }
+
+    /// Append a flag
+    pub fn flag(&mut self, flag: &'a Flag<'a>) -> &mut Self {
+        self.flags.push(flag);
+        self
+    }
+
+    /// Set the internal date
+    pub fn internal_date(&mut self, date: DateTime<FixedOffset>) -> &mut Self {
+        self.date = Some(date);
+        self
+    }
+}
+
+impl<'a> Into<AppendOptions<'a>> for AppendCmd<'a> {
+    fn into(self) -> AppendOptions<'a> {
+        AppendOptions {
+            flags: Some(&self.flags[..]),
+            date: self.date,
+        }
+    }
+}
+
 // `Deref` instances are so we can make use of the same underlying primitives in `Client` and
 // `Session`
 impl<T: Read + Write> Deref for Client<T> {
--- a/src/client.rs
+++ b/src/client.rs
@@ -94,7 +94,16 @@ pub struct AppendCmd<'a, T: Read + Write> {
 }
 
 impl<'a, T: Read + Write> AppendCmd<'a, T> {
-    /// Append a flag
+    /// The [`APPEND` command](https://tools.ietf.org/html/rfc3501#section-6.3.11) can take
+    /// an optional FLAGS parameter to set the flags on the new message.
+    ///
+    /// > If a flag parenthesized list is specified, the flags SHOULD be set
+    /// > in the resulting message; otherwise, the flag list of the
+    /// > resulting message is set to empty by default.  In either case, the
+    /// > Recent flag is also set.
+    ///
+    /// The [`\Recent` flag](https://tools.ietf.org/html/rfc3501#section-2.3.2) is not
+    /// allowed as an argument to `APPEND` and will be filtered out if present in `flags`.
     pub fn flag(&mut self, flag: Flag<'a>) -> &mut Self {
         self.flags.push(flag);
         self
@@ -106,14 +115,18 @@ impl<'a, T: Read + Write> AppendCmd<'a, T> {
         self
     }
 
-    /// Set the internal date
+    /// Pass a date in order to set the date that the message was originally sent.
+    ///
+    /// > If a date-time is specified, the internal date SHOULD be set in
+    /// > the resulting message; otherwise, the internal date of the
+    /// > resulting message is set to the current date and time by default.
     pub fn internal_date(&mut self, date: DateTime<FixedOffset>) -> &mut Self {
         self.date = Some(date);
         self
     }
 
-    /// Run command when set up
-    #[must_use]
+    /// Run command
+    #[must_use = "always run a command once options are configured"]
     pub fn run(&mut self) -> Result<()> {
         let flagstr = self
             .flags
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -93,6 +93,41 @@ pub struct AppendOptions<'a> {
...
+}
+
+impl<'a> Into<AppendOptions<'a>> for AppendCmd<'a> {
+    fn into(self) -> AppendOptions<'a> {
+        AppendOptions {
+            flags: Some(&self.flags[..]),
+            date: self.date,
+        }
+    }
+}
```
mblarsen, line 129: The PR adds both AppendOptions + AppendCmd. Leaving out the latter things are mostly ready. Tests are green and all.

I'm having an issue with AppendCmd and lifetimes with regards to the collected flags:

image

I'm still super comfortable with lifetimes in Rust. Perhaps you can give me some pointers @jonhoo ? I choose to use a Vec to store the flags for the ease of `push()`ing.

I tried adding separate lifespans `'a, 'b: 'a` but that quickly became a mess.

jonhoo:
Oooh, yes, I like the idea of `append` returning a builder instead! Just make sure you annotate it with `#[must_use`]

mblarsen:
So I refactored it to only have `AppendCmd` doing away with `AppendOptions`. I'm facing a new issue with lifetimes. In order to be able to invoke `run()` at the end we'll need a reference to session. I think this is where the lifetime issue arises:

image

Here is an example usage from the tests:

image

The original append is now reduced to this:

```rust
    pub fn append<'a, S: AsRef<str>, B: AsRef<[u8]>>(
        &mut self,
        mailbox: S,
        content: B,
    ) -> AppendCmd<'a, T> {
        AppendCmd {
            session: &self,
            content: content.as_ref(),
            mailbox: mailbox.as_ref(),
            flags: Vec::new(),
            date: None,
        }
    }
```

Again I tried a bunch of variations, e.g. adding a separate lifetime to the session ref, but couldn't quite grok it.

jonhoo:
Hmm... I _think_ you'll need:
```rust
    pub fn append<'a, S: AsRef<str>, B: AsRef<[u8]>>(
        &'a mut self,
        mailbox: S + 'a,
        content: B + 'a,
    ) -> AppendCmd<'a, T> {
        AppendCmd {
            session: &self,
            content: content.as_ref(),
            mailbox: mailbox.as_ref(),
            flags: Vec::new(),
            date: None,
        }
    }
```

mblarsen:
> Hmm... I _think_ you'll need:

Almost, but not quite. But it showed me in the right direction, so thank you for that. You'd better do a final review.

I've updated the tests + docs (moved flag + date docs to the AppendCmd)

Also added a `flags()` in addition to `flag()` in case you have your flags already come as a slice of flags.

mblarsen:
I don't know if I should worry about AppendCmd::flags is a vector. Would IMAP complain if I passed two Seen flags?

src/client.rs
```diff
@@ -84,6 +84,84 @@ pub struct Connection<T: Read + Write> {
     pub greeting_read: bool,
 }
 
+/// A builder for the append command
+pub struct AppendCmd<'a, T: Read + Write> {
```
mblarsen, line 88: `AppendCmd` probably shouldn't be public.

jonhoo: Hmm, why not? I think it probably needs to be.

jonhoo: Also, I think this struct should have `#[must_use]` added to it.

src/client.rs
```diff
@@ -84,6 +84,84 @@ pub struct Connection<T: Read + Write> {
...
+    /// The [`\Recent` flag](https://tools.ietf.org/html/rfc3501#section-2.3.2) is not
+    /// allowed as an argument to `APPEND` and will be filtered out if present in `flags`.
+    pub fn flag(&mut self, flag: Flag<'a>) -> &mut Self {
+        self.flags.push(flag);
+        self
+    }
+
+    /// Append an array of flags
+    pub fn flags(&mut self, flags: &'a [Flag<'a>]) -> &mut Self {
+        self.flags.append(&mut flags.to_vec());
```
jonhoo, line 114:
```suggestion
        self.flags.extend(flags);
```

src/client.rs
```diff
@@ -84,6 +84,84 @@ pub struct Connection<T: Read + Write> {
...
+    ///
+    /// The [`\Recent` flag](https://tools.ietf.org/html/rfc3501#section-2.3.2) is not
+    /// allowed as an argument to `APPEND` and will be filtered out if present in `flags`.
+    pub fn flag(&mut self, flag: Flag<'a>) -> &mut Self {
+        self.flags.push(flag);
+        self
+    }
+
+    /// Append an array of flags
+    pub fn flags(&mut self, flags: &'a [Flag<'a>]) -> &mut Self {
```
jonhoo, line 113:
```suggestion
    pub fn flags(&mut self, flags: impl IntoIterator<Item = Flag<'a>>) -> &mut Self {
```

src/client.rs
```diff
@@ -84,6 +84,84 @@ pub struct Connection<T: Read + Write> {
...
+    /// > If a date-time is specified, the internal date SHOULD be set in
+    /// > the resulting message; otherwise, the internal date of the
+    /// > resulting message is set to the current date and time by default.
+    pub fn internal_date(&mut self, date: DateTime<FixedOffset>) -> &mut Self {
+        self.date = Some(date);
+        self
+    }
+
+    /// Run command
+    #[must_use = "always run a command once options are configured"]
```
jonhoo, line 129: I don't think the `must_use` will do anything here. `#[must_use]` means "you must use the return type of this function", which isn't really what we're going for here. Instead, the annotation should be on the `AppendCmd` type, to indicate that if you are given one, you must use it.

src/client.rs
```diff
@@ -84,6 +84,84 @@ pub struct Connection<T: Read + Write> {
...
+    ///
+    /// > If a date-time is specified, the internal date SHOULD be set in
+    /// > the resulting message; otherwise, the internal date of the
+    /// > resulting message is set to the current date and time by default.
+    pub fn internal_date(&mut self, date: DateTime<FixedOffset>) -> &mut Self {
+        self.date = Some(date);
+        self
+    }
+
+    /// Run command
```
jonhoo, line 128: This could probably use some better docs.

src/client.rs
```diff
@@ -84,6 +84,84 @@ pub struct Connection<T: Read + Write> {
...
+    #[must_use = "always run a command once options are configured"]
+    pub fn run(&mut self) -> Result<()> {
```
jonhoo, line 130: I wonder if this function should instead be called `finish`?

jonhoo:
This is starting to look really good!

> I don't know if I should worry about AppendCmd::flags is a vector. Would IMAP complain if I passed two Seen flags?

I think it's probably okay for us to ignore that here, and leave it to the caller to just not do that.

src/client.rs
mblarsen: I see, that makes sense. So whenever the type is returned it should be used.

src/client.rs
mblarsen: Sounds good to me.

src/client.rs
mblarsen: My argument for not having it public was that I wanted to hide it "behind" the `append()` function so as to not let the user manually construct it.

But you are right it needs to be pub when used as a return type of pub function.

mblarsen:
Thanks for the feedback. I've refactored accordingly + some improvements to docs.

The tests are failing to compile right now because of `.flags()` the function that takes multiple flags at once.

The issue has to do with the `impl IntoIterator<...>` changes. The error given seems a bit inverted to me:

image

When called like this:

```rust
    let flags: &[Flag] = &[Flag::Seen, Flag::Flagged];
    c.append(mbox, e.message_to_string().unwrap().as_bytes())
        .flags(flags)
        .finish()
        .unwrap();
```

What makes it think that a `&Flag<'_>` is expected? Is it because values in a vector (what we extend) are all references?

jonhoo:
Ah, yes, it's because `flags` takes an iterator of _owned_ `Flag`s, not of references to `Flag`. Try calling it like:
```rust
    let flags = vec![Flag::Seen, Flag::Flagged];
    c.append(mbox, e.message_to_string().unwrap().as_bytes())
        .flags(flags)
        .finish()
        .unwrap();
```
or
```rust
    let flags = &[Flag::Seen, Flag::Flagged];
    c.append(mbox, e.message_to_string().unwrap().as_bytes())
        .flags(flags.into_iter().cloned())
        .finish()
        .unwrap();
```

mblarsen:
I think we are there now.

I'm still not sure why it expected a reference to Flag. It seems the other way around, that it got a reference to Flag and expected an owned Flag. (I'm referencing the note in the error message)

Are `&[T]`  always references?

jonhoo:
Yeah, the error message did seem a little unhelpful, I agree. :shrug: 
But yes, iterators you get from `&[T]` always produce `&T`, so that's likely where the error stems from :)

PR #177, opened 2021-01-12
Avoid trying to FETCH an empty set of messages

### LINKED ISSUE #176: `fetch_uid` sends a FETCH even if the given uid_set is empty
Such FETCH commands are syntactically invalid and the server will respond with, well, some kind of error—the exact behavior is probably server-dependent.

Since default/empty values of `fetch_uid`'s precise return-type cannot be publicly constructed (i.e. as a user I cannot avoid calling `fetch_uid` on an empty uid list by instead simply mapping such lists to empty `ZeroCopy<Vec<_>>`s), I would instead have expected that `fetch_uid` itself would condition any nontrivial operation upon the nonemptiness of the given uid list.

(Tangentially, I'm mildly annoyed by the stringly-typed way that the uid set is input, contrasting the output from `uid_search`.)

mmirate:
Also, apply correct validation to FETCH arguments.

This should fix #176.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/177)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -576,18 +576,16 @@ impl<T: Read + Write> Session<T> {
         S1: AsRef<str>,
         S2: AsRef<str>,
     {
-        let command;
-        self.run_command_and_read_response(if sequence_set.as_ref().is_empty() {
-            "NOOP"
+        if sequence_set.as_ref().is_empty() {
+            parse_fetches(vec![], &mut self.unsolicited_responses_tx)
         } else {
-            command = format!(
+            self.run_command_and_read_response(&format!(
                 "FETCH {} {}",
                 validate_sequence_set(sequence_set.as_ref())?,
                 validate_str_noquote(query.as_ref())?
-            );
-            &command
-        })
-        .and_then(|lines| parse_fetches(lines, &mut self.unsolicited_responses_tx))
+            ))
+            .and_then(|lines| parse_fetches(lines, &mut self.unsolicited_responses_tx))
+        }
     }
 
     /// Equivalent to [`Session::fetch`], except that all identifiers in `uid_set` are
@@ -597,18 +595,16 @@ impl<T: Read + Write> Session<T> {
         S1: AsRef<str>,
         S2: AsRef<str>,
     {
-        let command;
-        self.run_command_and_read_response(if uid_set.as_ref().is_empty() {
-            "NOOP"
+        if sequence_set.as_ref().is_empty() {
+            parse_fetches(vec![], &mut self.unsolicited_responses_tx)
         } else {
-            command = format!(
+            self.run_command_and_read_response(&format!(
                 "UID FETCH {} {}",
-                validate_sequence_set(uid_set.as_ref())?,
+                validate_sequence_set(sequence_set.as_ref())?,
                 validate_str_noquote(query.as_ref())?
-            );
-            &command
-        })
-        .and_then(|lines| parse_fetches(lines, &mut self.unsolicited_responses_tx))
+            ))
+            .and_then(|lines| parse_fetches(lines, &mut self.unsolicited_responses_tx))
+        }
     }
 
     /// Noop always succeeds, and it does nothing.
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -28,17 +28,50 @@ macro_rules! quote {
...
+
 fn validate_str(value: &str) -> Result<String> {
     let quoted = quote!(value);
-    if quoted.find('\n').is_some() {
-        return Err(Error::Validate(ValidateError('\n')));
-    }
-    if quoted.find('\r').is_some() {
-        return Err(Error::Validate(ValidateError('\r')));
-    }
+    quoted
```
jonhoo, line 46: Can this now call `validate_str_noquote`?

src/client.rs
```diff
@@ -543,11 +576,17 @@ impl<T: Read + Write> Session<T> {
...
         S2: AsRef<str>,
     {
-        self.run_command_and_read_response(&format!(
-            "FETCH {} {}",
-            sequence_set.as_ref(),
-            query.as_ref()
-        ))
+        let command;
+        self.run_command_and_read_response(if sequence_set.as_ref().is_empty() {
+            "NOOP"
```
jonhoo, line 581: We can probably avoid contacting the server at all here by simply calling `parse_fetches` directly on an empty string.

src/client.rs
```diff
@@ -558,11 +597,17 @@ impl<T: Read + Write> Session<T> {
...
         S2: AsRef<str>,
     {
-        self.run_command_and_read_response(&format!(
-            "UID FETCH {} {}",
-            uid_set.as_ref(),
-            query.as_ref()
-        ))
+        let command;
+        self.run_command_and_read_response(if uid_set.as_ref().is_empty() {
+            "NOOP"
```
jonhoo, line 602: Same here -- we can just call `parse_fetches` on an empty string.

src/client.rs
```diff
@@ -28,17 +28,50 @@ macro_rules! quote {
...
+        .next()
+        .and_then(|s| s.chars().next())
+        .map(|offender| Error::Validate(ValidateError(offender)))
+        .err()?;
+    Ok(value)
+}
+
+fn validate_sequence_set(value: &str) -> Result<&str> {
+    value
+        .matches(|c: char| c.is_ascii_whitespace())
```
jonhoo, line 67: Is it documented anywhere that a sequence set cannot contain whitespace?

mmirate: I believe so. The formal grammar at https://tools.ietf.org/html/rfc3501#page-90 defines a sequence set in ABNF as follows:

```
sequence-set = (seq-number / seq-range) *("," sequence-set)
seq-range = seq-number ":" seq-number
seq-number = nz-number / "*"
nz-number       = digit-nz *DIGIT
digit-nz        = %x31-39
```

Note how sequence-set, and the rules reachable from sequence-set, only contain the terminals DIGIT, and "*", ":", ","; and not the terminals SP, LWSP or similar.

jonhoo: Would you mind adding a link to that in the code, just for others who may read the code years from now?

jonhoo: And while you're at it, could you also add a comment as to why we don't quote the fetch query argument?

mmirate: I've added doc-comments on all three of these functions.

Aside: reading section 9 of the IMAP RFC gives me the idea that a not-so-stringly-typed API could consist of builders that correspond to the (relevant and nontrivial parts of the) grammar contained therein.

jonhoo: Beautiful, thank you!

And yes, I agree, it seems pretty reasonable to construct the non-stringly-typed API using the building blocks from the spec. I wonder if we can leverage `imap-proto` to a greater extent there?

jonhoo (approved):

jonhoo:
Oh, I just barely remembered. Would you mind also updating the `CHANGELOG`?

jonhoo:
Thank you! Will release shortly.

mmirate:
You're welcome!

jonhoo:
Released as `2.4.1` :tada:

src/client.rs
mmirate: Funny you should mention it, this very thing appears on the TODO-list in imap-proto's repo.

PR #179, opened 2021-02-22
Handle unsolicited "* OK Searched 91% of the mailbox, ETA 0:01" message

paulpr0:
When fetching a large number of UIDs, rust-imap would sometimes produce an error.

Looking into it, dovecot sometimes responds with

* OK Searched 91% of the mailbox, ETA 0:01

which is not handled when expecting a response to a Fetch message.

This is a small change to handle this case. I have made it quite explicit - it only handles an OK status with no code, as I'm not sure whether other messages which have an unknown code (but do contain a code) would be better handled by returning an error as is currently the case.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -331,6 +331,9 @@ fn handle_unilateral<'a>(
         Response::Expunge(n) => {
             unsolicited.send(UnsolicitedResponse::Expunge(n)).unwrap();
         }
+        Response::Data {status, code, information} if imap_proto::Status::Ok == status && code.is_none() => {
+            unsolicited.send(UnsolicitedResponse::Uncategorised(information.unwrap_or("").to_string())).unwrap();
+        }
         res => {
             return Some(res);
         }
@@ -437,6 +440,16 @@ mod tests {
         assert_eq!(fetches.len(), 1);
         assert_eq!(fetches[0].message, 37);
         assert_eq!(fetches[0].uid, Some(74));
+    }
+
+    #[test]
+    fn parse_uncategorised_ok() {
+        let lines = b"\
+        * OK Searched 91% of the mailbox, ETA 0:01\r\n";
+        let (mut send, recv) = mpsc::channel();
+        parse_fetches(lines.to_vec(), &mut send).unwrap();
+        assert_eq!(recv.try_recv(), Ok(UnsolicitedResponse::Uncategorised("Searched 91% of the mailbox, ETA 0:01".to_string())));
+
     }
 
     #[test]
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -277,6 +277,10 @@ pub enum UnsolicitedResponse {
     /// sequence numbers 9, 8, 7, 6, and 5.
     // TODO: the spec doesn't seem to say anything about when these may be received as unsolicited?
     Expunge(Seq),
+    /// An uncategorised response - one which has no code associated with it, just a text
+    /// Dovecot is known to send a status message with an ETA for a long running fetch which
+    /// has no [] enclosed code.
+    Uncategorised(String),
 }
 
 /// This type wraps an input stream and a type that was constructed by parsing that input stream,


paulpr0:
Firstly, thank you for the rust-imap crate. I have found it very useful! 

If it's any help, the IMAP server I see this message from is Dovecot v2.2.36 with a pretty vanilla setup.

jonhoo:
Hi! Thank you for the PR :heart: 

This seems like really odd behavior. I don't think the specification allows for these kinds of random untagged `Ok` responses? @djc have you seen these before? Do you know if clients are expected to just blindly ignore them?

paulpr0:
Hi,

I don't know the IMAP specs well, but looking at the definition of the OK response suggests it's allowed (https://tools.ietf.org/html/rfc3501#section-7.1.1)

> " The untagged form indicates an information-only message; "

I have had a look over the dovecot source code though as it's probably the most popular IMAP server on Linux (and the one I saw this behaviour on).

I think the function responsible for these untagged OK messages is [notify_ok](https://github.com/dovecot/core/blob/master/src/imap/mail-storage-callbacks.c)

The call I was seeing was from https://github.com/dovecot/core/blob/master/src/lib-storage/index/index-search.c#L1545, but there are some others too -  they all seem to be sent when the server is taking time to complete an operation and is letting the client know in an unstructured way (this one just says "Hang in there.." https://github.com/dovecot/core/blob/master/src/lib-storage/index/maildir/maildir-sync.c#L259)

Hope that helps,

Paul

jonhoo:
Ah, that's helpful, thank you! In that case I'd be happy to land this. Unfortunately, it'll have to wait for the release of `imap 2.0.0`, since it's a backwards-incompatible change to add a variant to `UnsolicitedResponse`. I really need to get the last bits together to do that. In the meantime, I'll leave some minor nits so that this is ready to land when `2.0` rolls around!

src/parse.rs
```diff
@@ -331,6 +331,17 @@ fn handle_unilateral<'a>(
...
             unsolicited.send(UnsolicitedResponse::Expunge(n)).unwrap();
         }
+        Response::Data {
+            status,
+            code,
+            information,
+        } if imap_proto::Status::Ok == status && code.is_none() => {
+            unsolicited
+                .send(UnsolicitedResponse::Uncategorised(
+                    information.unwrap_or("").to_string(),
```
jonhoo, line 341: Huh, I'm surprised that the `information` is an `Option`. Judging from [the spec](https://tools.ietf.org/html/rfc3501#section-7.1.1) it doesn't seem like it can be left out? @djc

src/parse.rs
```diff
@@ -331,6 +331,17 @@ fn handle_unilateral<'a>(
         Response::Expunge(n) => {
             unsolicited.send(UnsolicitedResponse::Expunge(n)).unwrap();
         }
+        Response::Data {
+            status,
+            code,
+            information,
+        } if imap_proto::Status::Ok == status && code.is_none() => {
```
jonhoo, line 338: Is it necessary for the code to `None` of this to be considered informational? From what I can tell, it's enough for the response to be untagged for it to be considered unsolicited — or am I missing something?

jonhoo: In which case we should include the `code` in the unsolicited response type.

src/types/mod.rs
```diff
@@ -277,6 +277,10 @@ pub enum UnsolicitedResponse {
     /// sequence numbers 9, 8, 7, 6, and 5.
     // TODO: the spec doesn't seem to say anything about when these may be received as unsolicited?
     Expunge(Seq),
+    /// An uncategorised response - one which has no code associated with it, just a text
```
jonhoo, line 280: Let's try to be a little more helpful here to potentially-confused receivers of this code. How about:
```rust
/// An unsolicited [`OK` response](https://tools.ietf.org/html/rfc3501#section-7.1.1) that
/// indicates an information-only message; the nature of the information MAY be
/// indicated by a response code.
///
/// The untagged OK is also used as one of three possible greetings
/// at connection startup.  It indicates that the connection is not
/// yet authenticated and that a `LOGIN` command is needed.
Notice(String)
```

jonhoo:
@paulpr0 Now that we're getting closer to a 3.0.0, do you think you might have some spare time to finish this up? Shouldn't be _too_ much left. Otherwise, @mordak, this would be another good candidate to get over the finish line.

Issue #178, opened 2021-01-15
[Suggestion] RFC 5256 - SORT command


soywod:
It could be nice to implement the `SORT` command from the [RFC 5256](https://tools.ietf.org/html/rfc5256#section-3). Plus it's close to the `SEARCH` command, so it should not be that difficult to integrate.

jonhoo:
I agree, that seems reasonable! Would be happy to take a look at a PR.

soywod:
I'm building an [email CLI client](https://github.com/soywod/himalaya) and I would love this feature to sort messages by date. I will try to implement it ASAP and propose a PR.

soywod:
It seems that the lib `imap-proto` does not have the parser for the `SORT` command, I opened [an issue](https://github.com/djc/tokio-imap/issues/117) there.

jonhoo:
Fixed by #180.

soywod:
@jonhoo I integrated the `v3.0.0-alpha.3` in my tool (mostly to use the sort cmd), and I saw some breaking changes. I noticed that the changelog is not up-to-date, would you like me to update it with the changes I noticed or you prefer to wait for the release `v3.0.0` to list all the breaking changes?

jonhoo:
That'd be great, thank you!

soywod:
Yes, I recognized the format, I also use [the same](https://github.com/soywod/himalaya/blob/master/CHANGELOG.md) :wink:

soywod:
https://github.com/jonhoo/rust-imap/pull/199

PR #180, opened 2021-03-12
RFC 5256 - SORT command

soywod:
#178

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/180)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ default = ["tls"]
 native-tls = { version = "0.2.2", optional = true }
 regex = "1.0"
 bufstream = "0.1"
-imap-proto = "0.13.0"
+imap-proto = "0.13.1"
 nom = "6.0"
 base64 = "0.13"
 chrono = "0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ nom = "6.0"
 base64 = "0.13"
 chrono = "0.4"
 lazy_static = "1.4"
-linked_hash_set = "0.1"
 
 [dev-dependencies]
 lettre = "0.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,11 +21,12 @@ default = ["tls"]
 native-tls = { version = "0.2.2", optional = true }
 regex = "1.0"
 bufstream = "0.1"
-imap-proto = "0.14.0"
+imap-proto = "0.14"
 nom = "6.0"
 base64 = "0.13"
 chrono = "0.4"
 lazy_static = "1.4"
+linked_hash_set = "0.1"
 
 [dev-dependencies]
 lettre = "0.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ default = ["tls"]
 native-tls = { version = "0.2.2", optional = true }
 regex = "1.0"
 bufstream = "0.1"
-imap-proto = "0.14"
+imap-proto = "0.14.0"
 nom = "6.0"
 base64 = "0.13"
 chrono = "0.4"
--- a/src/client.rs
+++ b/src/client.rs
@@ -1536,7 +1536,6 @@ mod tests {
     use super::*;
     use imap_proto::types::*;
     use std::borrow::Cow;
-    use std::collections::HashSet;
 
     macro_rules! mock_session {
         ($s:expr) => {
--- a/src/client.rs
+++ b/src/client.rs
@@ -1301,16 +1301,16 @@ impl<T: Read + Write> Session<T> {
     ///  - `SIZE`: Size of the message in octets
     ///  - `SUBJECT`: Base subject text
     ///  - `TO`: IMAP addr-mailbox of the first "To" address
-    pub fn sort<'c, H: AsRef<str>, Q: AsRef<str>>(
+    pub fn sort<'s, S: AsRef<str>>(
         &mut self,
-        criteria: &'c [extensions::sort::SortCriterion<'c>],
-        charset: H,
-        query: Q,
+        criteria: &'s [extensions::sort::SortCriterion<'s>],
+        charset: extensions::sort::SortCharset<'s>,
+        query: S,
     ) -> Result<Vec<Seq>> {
         self.run_command_and_read_response(&format!(
             "SORT {} {} {}",
             extensions::sort::SortCriteria(criteria),
-            charset.as_ref(),
+            charset,
             query.as_ref()
         ))
         .and_then(|lines| parse_ordered_ids(&lines, &mut self.unsolicited_responses_tx))
@@ -1319,16 +1319,16 @@ impl<T: Read + Write> Session<T> {
     /// Equivalent to [`Session::sort`], except that the returned identifiers
     /// are [`Uid`] instead of [`Seq`]. See also the [`UID`
     /// command](https://tools.ietf.org/html/rfc3501#section-6.4.8).
-    pub fn uid_sort<'c, S: AsRef<str>>(
+    pub fn uid_sort<'s, S: AsRef<str>>(
         &mut self,
-        criteria: &'c [extensions::sort::SortCriterion<'c>],
-        charset: S,
+        criteria: &'s [extensions::sort::SortCriterion<'s>],
+        charset: extensions::sort::SortCharset<'s>,
         query: S,
     ) -> Result<Vec<Seq>> {
         self.run_command_and_read_response(&format!(
             "UID SORT {} {} {}",
             extensions::sort::SortCriteria(criteria),
-            charset.as_ref(),
+            charset,
             query.as_ref()
         ))
         .and_then(|lines| parse_ordered_ids(&lines, &mut self.unsolicited_responses_tx))
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,9 +1,9 @@
 use bufstream::BufStream;
 use chrono::{DateTime, FixedOffset};
 use imap_proto::Response;
-use linked_hash_set::LinkedHashSet;
 #[cfg(feature = "tls")]
 use native_tls::{TlsConnector, TlsStream};
+use std::collections::HashSet;
 use std::io::{Read, Write};
 use std::net::{TcpStream, ToSocketAddrs};
 use std::ops::{Deref, DerefMut};
@@ -1273,7 +1273,7 @@ impl<T: Read + Write> Session<T> {
     ///
     ///  - `BEFORE <date>`: Messages whose internal date (disregarding time and timezone) is earlier than the specified date.
     ///  - `SINCE <date>`: Messages whose internal date (disregarding time and timezone) is within or later than the specified date.
-    pub fn search<S: AsRef<str>>(&mut self, query: S) -> Result<LinkedHashSet<Seq>> {
+    pub fn search<S: AsRef<str>>(&mut self, query: S) -> Result<HashSet<Seq>> {
         self.run_command_and_read_response(&format!("SEARCH {}", query.as_ref()))
             .and_then(|lines| parse_ids(&lines, &mut self.unsolicited_responses_tx))
     }
@@ -1281,7 +1281,7 @@ impl<T: Read + Write> Session<T> {
     /// Equivalent to [`Session::search`], except that the returned identifiers
     /// are [`Uid`] instead of [`Seq`]. See also the [`UID`
     /// command](https://tools.ietf.org/html/rfc3501#section-6.4.8).
-    pub fn uid_search<S: AsRef<str>>(&mut self, query: S) -> Result<LinkedHashSet<Uid>> {
+    pub fn uid_search<S: AsRef<str>>(&mut self, query: S) -> Result<HashSet<Uid>> {
         self.run_command_and_read_response(&format!("UID SEARCH {}", query.as_ref()))
             .and_then(|lines| parse_ids(&lines, &mut self.unsolicited_responses_tx))
     }
@@ -1313,19 +1313,14 @@ impl<T: Read + Write> Session<T> {
     ///  - `TO`: IMAP addr-mailbox of the first "To" address.
     ///
     ///  - `REVERSE <sort-key>`: Followed by another sort criterion, has the effect of that criterion but in reverse (descending) order.
-    pub fn sort<S: AsRef<str>>(
-        &mut self,
-        criteria: S,
-        charset: S,
-        query: S,
-    ) -> Result<LinkedHashSet<Seq>> {
+    pub fn sort<S: AsRef<str>>(&mut self, criteria: S, charset: S, query: S) -> Result<Vec<Seq>> {
         self.run_command_and_read_response(&format!(
             "SORT {} {} {}",
             criteria.as_ref(),
             charset.as_ref(),
             query.as_ref()
         ))
-        .and_then(|lines| parse_ids(&lines, &mut self.unsolicited_responses_tx))
+        .and_then(|lines| parse_ordered_ids(&lines, &mut self.unsolicited_responses_tx))
     }
 
     /// Equivalent to [`Session::sort`], except that the returned identifiers
@@ -1336,14 +1331,14 @@ impl<T: Read + Write> Session<T> {
         criteria: S,
         charset: S,
         query: S,
-    ) -> Result<LinkedHashSet<Seq>> {
+    ) -> Result<Vec<Seq>> {
         self.run_command_and_read_response(&format!(
             "UID SORT {} {} {}",
             criteria.as_ref(),
             charset.as_ref(),
             query.as_ref()
         ))
-        .and_then(|lines| parse_ids(&lines, &mut self.unsolicited_responses_tx))
+        .and_then(|lines| parse_ordered_ids(&lines, &mut self.unsolicited_responses_tx))
     }
 
     // these are only here because they are public interface, the rest is in `Connection`
--- a/src/client.rs
+++ b/src/client.rs
@@ -1880,6 +1880,51 @@ mod tests {
     }
 
     #[test]
+    fn sort() {
+        use extensions::sort::{SortCharset, SortCriterion};
+
+        let response = b"* SORT 1 2 3 4 5\r\n\
+            a1 OK Sort completed\r\n"
+            .to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let ids = session
+            .sort(&[SortCriterion::Arrival], SortCharset::Utf8, "ALL")
+            .unwrap();
+        let ids: Vec<u32> = ids.iter().cloned().collect();
+        assert!(
+            session.stream.get_ref().written_buf == b"a1 SORT (ARRIVAL) UTF-8 ALL\r\n".to_vec(),
+            "Invalid sort command"
+        );
+        assert_eq!(ids, [1, 2, 3, 4, 5].iter().cloned().collect::<Vec<_>>());
+    }
+
+    #[test]
+    fn uid_sort() {
+        use extensions::sort::{SortCharset, SortCriterion};
+
+        let response = b"* SORT 1 2 3 4 5\r\n\
+            a1 OK Sort completed\r\n"
+            .to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let ids = session
+            .uid_sort(
+                &[SortCriterion::Reverse(&SortCriterion::Size)],
+                SortCharset::UsAscii,
+                "SUBJECT",
+            )
+            .unwrap();
+        let ids: Vec<Uid> = ids.iter().cloned().collect();
+        assert!(
+            session.stream.get_ref().written_buf
+                == b"a1 UID SORT (REVERSE SIZE) US-ASCII SUBJECT\r\n".to_vec(),
+            "Invalid sort command"
+        );
+        assert_eq!(ids, [1, 2, 3, 4, 5].iter().cloned().collect::<Vec<_>>());
+    }
+
+    #[test]
     fn capability() {
         let response = b"* CAPABILITY IMAP4rev1 STARTTLS AUTH=GSSAPI LOGINDISABLED\r\n\
             a1 OK CAPABILITY completed\r\n"
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,9 +1,9 @@
 use bufstream::BufStream;
 use chrono::{DateTime, FixedOffset};
 use imap_proto::Response;
+use linked_hash_set::LinkedHashSet;
 #[cfg(feature = "tls")]
 use native_tls::{TlsConnector, TlsStream};
-use std::collections::HashSet;
 use std::io::{Read, Write};
 use std::net::{TcpStream, ToSocketAddrs};
 use std::ops::{Deref, DerefMut};
@@ -1273,7 +1273,7 @@ impl<T: Read + Write> Session<T> {
     ///
     ///  - `BEFORE <date>`: Messages whose internal date (disregarding time and timezone) is earlier than the specified date.
     ///  - `SINCE <date>`: Messages whose internal date (disregarding time and timezone) is within or later than the specified date.
-    pub fn search<S: AsRef<str>>(&mut self, query: S) -> Result<HashSet<Seq>> {
+    pub fn search<S: AsRef<str>>(&mut self, query: S) -> Result<LinkedHashSet<Seq>> {
         self.run_command_and_read_response(&format!("SEARCH {}", query.as_ref()))
             .and_then(|lines| parse_ids(&lines, &mut self.unsolicited_responses_tx))
     }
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -21,7 +21,7 @@ default = ["tls"]
 native-tls = { version = "0.2.2", optional = true }
 regex = "1.0"
 bufstream = "0.1"
-imap-proto = "0.13.0"
+imap-proto = "0.13.1"
```
soywod, line 24: https://github.com/djc/tokio-imap/issues/117#issuecomment-797059582

soywod:
In fact we can't keep a `HashSet<Seq>` for the `SORT` command, because the order matters.That's why tests fail randomly. I will investigate more.

src/client.rs
```diff
@@ -1286,6 +1286,66 @@ impl<T: Read + Write> Session<T> {
...
+    ///  - `SUBJECT`: Base subject text.
+    ///  - `TO`: IMAP addr-mailbox of the first "To" address.
+    ///
+    ///  - `REVERSE <sort-key>`: Followed by another sort criterion, has the effect of that criterion but in reverse (descending) order.
+    pub fn sort<S: AsRef<str>>(
+        &mut self,
+        criteria: S,
+        charset: S,
+        query: S,
+    ) -> Result<HashSet<Seq>> {
```
soywod, line 1321: In fact we can't keep a `HashSet<Seq>` for the `SORT` command, because the order matters.That's why tests fail randomly. I will investigate more.

soywod: I propose to replace the `HashSet` by a [`LinkedHashSet`](https://crates.io/crates/linked_hash_set), so the insertion order is preserved, what do you think?

jonhoo: I'd probably then go with [`IndexSet`](https://docs.rs/indexmap/1/indexmap/set/struct.IndexSet.html) instead, as that crate is much more widely used :)

Another option is to just switch this to a `Vec`, and leave it to the caller whether they actually want a set :sweat_smile: @mordak what do you think?

mordak: I would be inclined to just use `Vec` - we should not have duplicate IDs in the response anyway (and even if we do it is not catastrophic - the client will just deal with the same ID more than once, which is apparently something the server thinks they should do anyway?). As near as I can tell the original motivation for using `HashSet` was [API convenience](https://github.com/mattnenterprise/rust-imap/pull/89#discussion_r222800445) rather than correctness, so there doesn't seem to be any technical reason for using a `Set`. If we still want things like additional APIs, then `IndexSet` does seem like a more widely used crate.

But since the `SORT` command explicitly cares about the order of responses from the server and ensuring they are passed to the client in that order, then something straightforward like `Vec` is the obvious choice. It is nice to avoid adding additional dependencies for library users if possible, so for me it boils down to if the additional dependency and API uniformity cost is worth the gain. I don't see a really compelling reason to use anything fancy here, so I would just go with `Vec`. :)

soywod: I proposed this to avoid changing to much the existing code (since `HashSet` and `LinkedHashSet` have more or less the same API), but I totally agree that `Vec` is a much better choice. I will change this whenever I can!

soywod: Done :heavy_check_mark:

Cargo.toml
```diff
@@ -21,7 +21,7 @@ default = ["tls"]
 native-tls = { version = "0.2.2", optional = true }
 regex = "1.0"
 bufstream = "0.1"
-imap-proto = "0.14.0"
+imap-proto = "0.14"
```
jonhoo, line 24: I don't think this should be necessary? Semantic versioning will already take care to use the latest version. If anything this should list the patch version that added any feature you now depend on.

soywod: Changed :heavy_check_mark:

src/client.rs
```diff
@@ -1286,6 +1286,61 @@ impl<T: Read + Write> Session<T> {
...
+    ///  - `ARRIVAL`: Internal date and time of the message.  This differs from the ON criteria in SEARCH, which uses just the internal date.
+    ///  - `CC`: IMAP addr-mailbox of the first "cc" address.
+    ///  - `DATE`: Sent date and time, as described in section 2.2.
+    ///  - `FROM`: IMAP addr-mailbox of the first "From" address.
+    ///  - `SIZE`: Size of the message in octets.
+    ///  - `SUBJECT`: Base subject text.
+    ///  - `TO`: IMAP addr-mailbox of the first "To" address.
+    ///
+    ///  - `REVERSE <sort-key>`: Followed by another sort criterion, has the effect of that criterion but in reverse (descending) order.
+    pub fn sort<S: AsRef<str>>(&mut self, criteria: S, charset: S, query: S) -> Result<Vec<Seq>> {
```
jonhoo, line 1316: I'm just now taking a more careful look, and I wonder if we can do better here to avoid the caller having to ensure they match the official syntax. Can we switch `criteria` to take a `&[SortCriteria]` instead, and then make
```rust
struct SortCriteria {
  reverse: bool,
  key: SortKey,
}
#[non_exhaustive]
enum SortKey {
  /// Internal date and time of the message.
  ///
  /// This differs from ...
  Arrival,
  /// ...
  Cc,
  Date,
  From,
  Size,
  Subject,
  To,
}
```
That way we wouldn't need to include all the official syntax definitions.
I also think we should make each `S` be `impl AsRef<str>` instead — that will allow the caller to use different types (e.g., `String` and `&'static str`) for each string-like argument.

src/client.rs
```diff
@@ -1286,6 +1286,61 @@ impl<T: Read + Write> Session<T> {
...
+    ///
+    /// sort            = ["UID" SP] "SORT" SP sort-criteria SP search-criteria
+    /// sort-criteria   = "(" sort-criterion *(SP sort-criterion) ")"
+    /// sort-criterion  = ["REVERSE" SP] sort-key
+    /// sort-key        = "ARRIVAL" / "CC" / "DATE" / "FROM" / "SIZE" /
+    ///                   "SUBJECT" / "TO"
+    /// search-criteria = charset 1*(SP search-key)
+    /// charset         = atom / quoted
+    ///                     ; CHARSET values MUST be registered with IANA
+    /// sort-data       = "SORT" *(SP nz-number)
```
jonhoo, line 1302: Where does `sort-data` come into play?

jonhoo: I _think_ that's only used in the response, so probably not worth including in the docs.

jonhoo:
I really like the new direction of the parsing, thank you!

soywod:
Sorry, I was quite busy these days. I try to update my PR this week!

src/client.rs
soywod: :heavy_check_mark: Done

soywod:
I updated my proposal. I opted for a single enum `SortCriterion`. The `Reverse` variant takes a ref to another criterion to avoid infinite recursion (I guess a `Box` could also do the trick there?). I am still new at Rust, I may have implemented sth wrongly. Please correct me if needed :wink:, I'm open to discussion!

src/client.rs
```diff
@@ -1286,6 +1286,54 @@ impl<T: Read + Write> Session<T> {
...
+            extensions::sort::SortCriteria(criteria),
+            charset,
+            query.as_ref()
+        ))
+        .and_then(|lines| parse_ordered_ids(&lines, &mut self.unsolicited_responses_tx))
+    }
+
+    /// Equivalent to [`Session::sort`], except that the returned identifiers
+    /// are [`Uid`] instead of [`Seq`]. See also the [`UID`
+    /// command](https://tools.ietf.org/html/rfc3501#section-6.4.8).
```
jonhoo, line 1321: In Rust, the first paragraph is considered a summary, and often rendered on its own. Let's take advantage of that and make this:
```suggestion
    /// Equivalent to [`Session::sort`], except that it returns [`Uid`]s.
    ///
    /// See also [`Session::uid_search`].
```

src/extensions/sort.rs
```diff
@@ -0,0 +1,148 @@
...
+            write!(f, "({})", criteria.join(" "))
+        }
+    }
+}
+
+/// The defined sort criteria are as follows. Refer to the Formal
+/// Syntax section for the precise syntactic definitions of the
+/// arguments. If the associated [RFC-822](https://tools.ietf.org/html/rfc822)
+/// header for a particular criterion is absent, it is treated as the empty string.
+/// The empty string always collates before non-empty strings.
```
jonhoo, line 28: This text reflects stuff that we as the protocol implementors need to know, but it's not really something the user of `imap` needs to know given that we've abstracted that away from them. Let's update this to something simpler:
```suggestion
/// Message sorting preferences used for [`Session::sort`] and [`Session::uid_sort`].
```

src/extensions/sort.rs
```diff
@@ -0,0 +1,148 @@
...
+    Reverse(&'c SortCriterion<'c>),
+
+    /// Size of the message in octets.
+    Size,
+
+    /// Base subject text.
+    Subject,
+
+    /// [IMAP](https://tools.ietf.org/html/rfc5256#ref-IMAP) addr-mailbox
+    /// of the first "To" address.
```
jonhoo, line 57: Can we make this documentation a little clearer for someone who doesn't know the exact spec? What is an `addr-mailbox`?

src/extensions/sort.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// [IMAP](https://tools.ietf.org/html/rfc5256#ref-IMAP) addr-mailbox
+    /// of the first "cc" address.
+    Cc,
+
+    /// Sent date and time, as described in
+    /// [section 2.2](https://tools.ietf.org/html/rfc5256#section-2.20).
+    Date,
+
+    /// [IMAP](https://tools.ietf.org/html/rfc5256#ref-IMAP) addr-mailbox
+    /// of the first "From" address.
```
jonhoo, line 43: Can we make this documentation a little clearer for someone who doesn't know the exact spec? What is an `addr-mailbox`?

src/extensions/sort.rs
```diff
@@ -0,0 +1,148 @@
...
+/// arguments. If the associated [RFC-822](https://tools.ietf.org/html/rfc822)
+/// header for a particular criterion is absent, it is treated as the empty string.
+/// The empty string always collates before non-empty strings.
+pub enum SortCriterion<'c> {
+    /// Internal date and time of the message. This differs from the
+    /// ON criteria in SEARCH, which uses just the internal date.
+    Arrival,
+
+    /// [IMAP](https://tools.ietf.org/html/rfc5256#ref-IMAP) addr-mailbox
+    /// of the first "cc" address.
```
jonhoo, line 35: Can we make this documentation a little clearer for someone who doesn't know the exact spec? What is an `addr-mailbox`?

src/extensions/sort.rs
```diff
@@ -0,0 +1,148 @@
...
+            Subject => write!(f, "SUBJECT"),
+            To => write!(f, "TO"),
+        }
+    }
+}
+
+/// The charset argument is mandatory (unlike SEARCH) and indicates
+/// the CHARSET of the strings that appear in the searching
+/// criteria. The US-ASCII and UTF-8 charsets MUST be implemented.
+/// All other charsets are optional.
```
jonhoo, line 81:
```suggestion
/// The character encoding to use for strings that are subject to a [`SortCriterion`].
///
/// Servers are only required to implement [`SortCharset::UsAscii`] and [`SortCharset::Utf8`].
```

src/extensions/sort.rs
```diff
@@ -0,0 +1,148 @@
...
+pub enum SortCharset<'c> {
+    /// Mandatory UTF-8
```
jonhoo, line 83:
```suggestion
    /// Strings are UTF-8 encoded.
```

src/extensions/sort.rs
```diff
@@ -0,0 +1,148 @@
...
+    Utf8,
+
+    /// Mandatory US-ASCII
```
jonhoo, line 86:
```suggestion
    /// Strings are encoded with ASCII.
```

src/extensions/sort.rs
```diff
@@ -0,0 +1,148 @@
...
+    UsAscii,
+
+    /// Optional custom
```
jonhoo, line 89:
```suggestion
    /// Strings are encoded using some other character set.
    ///
    /// Note that this option is subject to server support for the specified character set.
```

src/parse.rs
```diff
@@ -350,6 +354,20 @@ pub fn parse_ids(
     }
 }
 
+pub fn parse_ids(
```
jonhoo, line 357:
```suggestion
pub fn parse_id_set(
```

src/parse.rs
```diff
@@ -350,6 +354,20 @@ pub fn parse_ids(
...
+    lines: &[u8],
+    unsolicited: &mut mpsc::Sender<UnsolicitedResponse>,
+) -> Result<HashSet<u32>> {
+    parse_ids_with(lines, unsolicited, HashSet::new())
+}
+
+pub fn parse_ordered_ids(
```
jonhoo, line 364:
```suggestion
pub fn parse_id_sequence(
```

jonhoo (changes requested):
I like the way you're going with this :+1: And yes, I think `&` is fine for `SortCriterion`.

I've left a bunch of mostly minor comments in there that should hopefully not be _too_ onerous to fix!

src/client.rs
```diff
@@ -1286,6 +1286,54 @@ impl<T: Read + Write> Session<T> {
...
+    pub fn uid_sort<'s, S: AsRef<str>>(
+        &mut self,
+        criteria: &'s [extensions::sort::SortCriterion<'s>],
+        charset: extensions::sort::SortCharset<'s>,
+        query: S,
+    ) -> Result<Vec<Seq>> {
```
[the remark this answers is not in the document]
soywod: Done :heavy_check_mark:

src/extensions/sort.rs
```diff
@@ -0,0 +1,148 @@
...
+pub(crate) struct SortCriteria<'c>(pub(crate) &'c [SortCriterion<'c>]);
+
+impl<'c> fmt::Display for SortCriteria<'c> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.0.is_empty() {
+            write!(f, "")
+        } else {
+            let criteria: Vec<String> = self.0.iter().map(|c| c.to_string()).collect();
+            write!(f, "({})", criteria.join(" "))
+        }
```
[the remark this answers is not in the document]
soywod: I half agree on this one. Indeed there is an *unnecessary* allocation, but we are talking about few strings that are automatically freed at the end of the bloc. And we loose a lot in readability. For me it looks like an extreme *unnecessary* optimization, but maybe I miss sth? Maybe there is a more readable version without allocating? An other opinion could be nice.

src/extensions/sort.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+}
+
+/// The defined sort criteria are as follows. Refer to the Formal
+/// Syntax section for the precise syntactic definitions of the
+/// arguments. If the associated [RFC-822](https://tools.ietf.org/html/rfc822)
+/// header for a particular criterion is absent, it is treated as the empty string.
+/// The empty string always collates before non-empty strings.
+pub enum SortCriterion<'c> {
```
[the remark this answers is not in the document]
soywod: :heavy_check_mark: Done

(Just by curiosity, what does the `#[non_exhaustive]` implies?)

src/client.rs
```diff
@@ -1286,6 +1286,54 @@ impl<T: Read + Write> Session<T> {
             .and_then(|lines| parse_ids(&lines, &mut self.unsolicited_responses_tx))
     }
 
+    /// The SORT command is a variant of SEARCH with sorting semantics for
```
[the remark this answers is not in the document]
soywod: :heavy_check_mark: Done

src/client.rs
```diff
@@ -1286,6 +1286,54 @@ impl<T: Read + Write> Session<T> {
...
+    ///  - `DATE`: Sent date and time
+    ///  - `FROM`: IMAP addr-mailbox of the first "From" address
+    ///  - `REVERSE <sort-key>`: Followed by another sort criterion, has the effect of that criterion but in reverse (descending) order
+    ///  - `SIZE`: Size of the message in octets
+    ///  - `SUBJECT`: Base subject text
+    ///  - `TO`: IMAP addr-mailbox of the first "To" address
+    pub fn sort<'s, S: AsRef<str>>(
+        &mut self,
+        criteria: &'s [extensions::sort::SortCriterion<'s>],
+        charset: extensions::sort::SortCharset<'s>,
```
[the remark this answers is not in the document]
soywod: Ah yes, indeed! Each time I try this, it never works :laughing:

:heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/parse.rs
soywod: :heavy_check_mark: Done

src/parse.rs
soywod: :heavy_check_mark: Done

src/client.rs
```diff
@@ -1286,6 +1286,54 @@ impl<T: Read + Write> Session<T> {
...
+        .and_then(|lines| parse_ordered_ids(&lines, &mut self.unsolicited_responses_tx))
+    }
+
+    /// Equivalent to [`Session::sort`], except that the returned identifiers
+    /// are [`Uid`] instead of [`Seq`]. See also the [`UID`
+    /// command](https://tools.ietf.org/html/rfc3501#section-6.4.8).
+    pub fn uid_sort<'s, S: AsRef<str>>(
+        &mut self,
+        criteria: &'s [extensions::sort::SortCriterion<'s>],
+        charset: extensions::sort::SortCharset<'s>,
```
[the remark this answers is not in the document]
soywod: :heavy_check_mark: Done

src/client.rs
soywod: > In Rust, the first paragraph is considered a summary, and often rendered on its own.

I did not know about this, thank you :+1: 

:heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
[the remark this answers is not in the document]
soywod: I am not even sure that the `Vec<String>` allocation is less performant than `2 + criteria.len()` function calls (`write!`) :thinking:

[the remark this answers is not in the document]
jonhoo: Yeah, you're right. Let's value readability here over performance, and only optimize if necessary.

I _really_ wish that the standard library supported joining a slice of `&str`, not _just_ `String` though. The extra pass that clones all the `&str` _and_ collect them into a new `Vec` makes me so sad.

src/extensions/sort.rs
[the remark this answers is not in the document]
jonhoo: Ah, you can read a bunch more about it in [RFC 2008](https://rust-lang.github.io/rfcs/2008-non-exhaustive.html), but _basically_ it means that we can add more variants to the `enum` in the future without having to issue a new breaking version release.

src/client.rs
```diff
@@ -1275,15 +1275,51 @@ impl<T: Read + Write> Session<T> {
...
     /// command](https://tools.ietf.org/html/rfc3501#section-6.4.8).
     pub fn uid_search<S: AsRef<str>>(&mut self, query: S) -> Result<HashSet<Uid>> {
         self.run_command_and_read_response(&format!("UID SEARCH {}", query.as_ref()))
-            .and_then(|lines| parse_ids(&lines, &mut self.unsolicited_responses_tx))
+            .and_then(|lines| parse_id_set(&lines, &mut self.unsolicited_responses_tx))
+    }
+
+    /// This issues the [SORT command](https://tools.ietf.org/html/rfc5256#section-3),
+    /// which returns sorted search results. This command is like [`Session::search`], except that
+    /// the results are also sorted according to the supplied criteria (subject to the given charset).
```
jonhoo, line 1291: You'll want to add two newlines (a paragraph break) just after the first sentence here. It'll make the documentation render more nicely :)
```suggestion
    /// This issues the [SORT command](https://tools.ietf.org/html/rfc5256#section-3),
    /// which returns sorted search results.
    ///
    /// This command is like [`Session::search`], except that
    /// the results are also sorted according to the supplied criteria (subject to the given charset).
```

src/extensions/sort.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+}
+
+/// Message sorting preferences used for [`Session::sort`] and [`Session::uid_sort`].
+///
+/// References:
+/// - [IMAP ref](https://tools.ietf.org/html/rfc5256#ref-IMAP)
+/// - [Address formal syntax](https://tools.ietf.org/html/rfc3501#section-9)
+/// - [Address spec](https://tools.ietf.org/html/rfc2822#section-3.4.1)
```
jonhoo, line 29: Let's write this in a more helpful way:
```suggestion
/// Any sorting criterion that refers to an address (`From`, `To`, etc.) sorts according to the "addr-mailbox" of the indicated address. You can find the formal syntax for addr-mailbox [in the IMAP spec](https://tools.ietf.org/html/rfc3501#section-9), and a more detailed discussion of the relevant semantics [in RFC 2822](https://tools.ietf.org/html/rfc2822#section-3.4.1). Essentially, the address refers _either_ to the name of the contact _or_ to its local-part (the left part of the email address, before the `@`).
```

After adding ^, we can simplify all the documentation strings for the individual variants below.

src/extensions/sort.rs
```diff
@@ -0,0 +1,155 @@
...
+            Subject => write!(f, "SUBJECT"),
+            To => write!(f, "TO"),
+        }
+    }
+}
+
+/// The character encoding to use for strings that are subject to a [`SortCriterion`].
+///
+/// Servers are only required to implement [`SortCharset::UsAscii`] and [`SortCharset::Utf8`].
+pub enum SortCharset<'c> {
```
jonhoo, line 87: Here too we should derive the standard traits.

src/extensions/sort.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Strings are UTF-8 encoded.
+    Utf8,
+
+    /// Strings are encoded with ASCII.
+    UsAscii,
+
+    /// Strings are encoded using some other character set.
+    ///
+    /// Note that this option is subject to server support for the specified character
```
jonhoo, line 96:
```suggestion
    /// Note that this option is subject to server support for the specified character set.
```

jonhoo:
We're getting really close!

src/client.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
soywod: :heavy_check_mark: Done

src/extensions/sort.rs
```diff
@@ -0,0 +1,153 @@
...
+pub enum SortCriterion<'c> {
+    /// Internal date and time of the message. This differs from the
+    /// ON criteria in SEARCH, which uses just the internal date.
+    Arrival,
+
+    /// IMAP addr-mailbox of the first "Cc" address.
+    Cc,
+
+    /// Sent date and time, as described in
+    /// [section 2.2](https://tools.ietf.org/html/rfc5256#section-2.20).
```
jonhoo, line 43:
```suggestion
    /// [section 2.2](https://tools.ietf.org/html/rfc5256#section-2.2).
```

jonhoo (approved):
This is in great shape now, thanks for sticking with it! One minor change which I _think_ I can commit myself through the GitHub interface, and then I'll land it!

src/parse.rs
```diff
@@ -344,6 +348,21 @@ pub fn parse_ids(
     }
 }
 
+
```
jonhoo, line 351:
```suggestion
```

jonhoo:
Wohoo, it's in! :tada:
We're getting closer and closer to that [3.0.0 milestone](https://github.com/jonhoo/rust-imap/milestone/2).

PR #183, opened 2021-03-19
Add vendored feature to set native-tls/vendored

Rulliam:
Proposal, because I need openssl-sys (so native-tls) to have the vendored feature
With this feature, we can allow cross-compilation for rust-imap (tested with android targets), and allow to link statically with env OPENSSL_STATIC=1

Note that this feature will not be set by default, so it's not a breaking change.
(I named this feature `vendored` because native-tls did the same thing to enable openssl-sys vendored feature)

Thank you!

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/183)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ categories = ["email", "network-programming"]
 
 [features]
 tls = ["native-tls"]
+vendored = ["tls", "native-tls/vendored"]
 default = ["tls"]
 
 [dependencies]


Rulliam:
Also note that you need `perl` and `libssl-dev` (or `openssl-devel` on fedora) to build openssl-sys/vendored

jonhoo:
I'm not sure why you need to add this as a feature of `imap`? You could instead just add `native-tls` as a direct dependency of your crate, and then enable the feature there. Since cargo shares features across all instances of a crate in the dependency graph, if you enable the feature once in your crate, that will also apply when it's compiled as part of `imap`.

Rulliam:
Well, you're right, I simply forgot about that... 😓
Thank you

Issue #152, opened 2019-11-21
IDLE example


Lem:
I'm fairly new to rust and would like to learn it by writing something with IMAP.

As I would like to use IDLE, too, I wanted to ask if its possible to add an example on how to use that extension.

hpk42:
you could look at this unmerged PR where i provided an 
working IDLE example: 

https://github.com/jonhoo/rust-imap/pull/148/

maybe you could take up the PR and finish it? :) 

In my project we were in need of manually interruptible IDLE
which is hard to do cross-platform (you need to write some bytes
to the FD on which IDLE is blocking -- or you need to use an
async stack).

jonhoo:
Fixed by #186.

PR #186, opened 2021-03-28
Pass IDLE responses to caller.

mordak:
Picking up #130, but doing it a little differently. Implement passing IDLE responses to the caller via a callback function.

Also add an example, for #152.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/186)


mordak:
Pipeline fail is a clippy lint `redundant_slicing` (on 1.52 beta) - I can fix that in another PR, or fix it here.

jonhoo:
Thank you for pushing on this! I'm hoping I'll have some time to review over the weekend — otherwise I'll get to it as soon as I can :tada:

src/extensions/idle.rs
```diff
@@ -13,8 +15,31 @@ use std::time::Duration;
...
+/// logic is needed to handle the IDLE response, and then returns a [`CallbackAction`]
+/// to `Continue` or `Stop` listening on the channel.
+/// For users that want the IDLE to exit on any change (the behavior proior to version 3.0),
+/// a convenience callback function `callback_stop` is provided.
+///
+/// ```no_run
+/// # use native_tls::TlsConnector;
+/// use imap::extensions::idle;
+/// let ssl_conn = TlsConnector::builder().build().unwrap();
+/// let client = imap::connect(("example.com", 993), "example.com", &ssl_conn)
```
jonhoo, line 31: Seeing this `TlsConnetor` business all over, I wonder if we should switch (in a separate PR) to a builder pattern with convenience constructors for common TLS implementations. See what we did [in `fantoccini`](https://github.com/jonhoo/fantoccini/pull/115/files), and in particular:
https://github.com/jonhoo/fantoccini/blob/2d1e24acc337e1f0c36e543eb52a20538f1dc18a/src/lib.rs#L180-L196

src/extensions/idle.rs
```diff
@@ -40,6 +65,21 @@ pub enum WaitOutcome {
...
+#[derive(Debug, PartialEq, Eq)]
+pub enum CallbackAction {
+    /// Continue receiving responses from the IDLE connection.
+    Continue,
+    /// Stop receiving responses, and exit the IDLE wait.
+    Stop,
+}
+
+/// A convenience function to always cause the IDLE handler to exit on any change.
+pub fn callback_stop(_response: UnsolicitedResponse) -> CallbackAction {
```
jonhoo, line 79: How about `stop_on_any`?

src/extensions/idle.rs
```diff
@@ -100,37 +140,65 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+            };
 
-            // Handle Dovecot's imap_idle_notify_interval message
-            if v.eq_ignore_ascii_case(b"* OK Still here\r\n") {
+            // Update remaining data with unparsed data if needed.
+            if rest.is_empty() {
                 v.clear();
-            } else {
-                break Ok(result);
+            } else if rest.len() != v.len() {
```
jonhoo, line 179: This clause confuses me. How would we get here?

src/extensions/idle.rs
```diff
@@ -100,37 +140,65 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+                v = rest.into();
```
jonhoo, line 180: It's unfortunate that we keep re-allocating `v` here. I _think_ it should be possible to instead truncate the start of the `Vec` directly.

src/extensions/idle.rs
```diff
@@ -100,37 +140,65 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
             }
+            (_, result) => result,
         }
     }
 
-    /// Block until the selected mailbox changes.
-    pub fn wait(mut self) -> Result<()> {
-        self.wait_inner(true).map(|_| ())
+    /// Block until the given callback returns `Stop`, or until an unhandled
+    /// response arrives on the IDLE channel.
```
jonhoo, line 196: What is an "unhandled response" here?

src/extensions/idle.rs
```diff
@@ -159,26 +231,42 @@ impl<'a, T: SetReadTimeout + Read + Write + 'a> Handle<'a, T> {
...
+        self.timed_wait(keepalive, true, callback).map(|_| ())
     }
 
-    /// Block until the selected mailbox changes, or until the given amount of time has expired.
+    /// Block until the given amount of time has elapsed, or the given callback
+    /// returns `Stop`, or until an unhandled response arrives on the IDLE channel.
     #[deprecated(note = "use wait_with_timeout instead")]
-    pub fn wait_timeout(self, timeout: Duration) -> Result<()> {
-        self.wait_with_timeout(timeout).map(|_| ())
+    pub fn wait_timeout<F>(self, timeout: Duration, callback: F) -> Result<()>
```
jonhoo, line 240: This method is deprecated in 2.x and should just be removed.

src/extensions/idle.rs
jonhoo: Ah, looking at the parser, it seems like it's anything that we don't _explicitly_ recognize as a possible unsolicited response. That should be documented somewhere. I also wonder if we should instead have an `Unexpected` category under `UnsolicitedResponse` that just contains the raw `imap_proto::Response` — what do you think?

src/parse.rs
```diff
@@ -350,59 +351,35 @@ pub fn parse_ids(
     }
 }
 
+/// Parse a single unsolicited response from IDLE responses.
+pub fn parse_idle(lines: &[u8]) -> (&[u8], Option<Result<UnsolicitedResponse>>) {
+    match imap_proto::parser::parse_response(lines) {
+        Ok((rest, response)) => match UnsolicitedResponse::try_from(response) {
```
jonhoo, line 357: Are there any responses that can come during IDLE that cannot come as unsolicited responses otherwise?

src/types/unsolicited_response.rs
```diff
@@ -0,0 +1,309 @@
...
+        code: Option<ResponseCode>,
+        /// Information text that may be presented to the user.
+        information: Option<String>,
+    },
+
+    /// An unsolicited `BYE` response.
+    ///
+    /// The `BYE` response may have an optional `ResponseCode` that provides additional
+    /// information, per [RFC3501](https://tools.ietf.org/html/rfc3501#section-7.1.5).
+    Bye {
```
jonhoo, line 132: If we receive one of these, does that mean the connection is about to be terminated? Should we handle these directly in the library instead and turn them into an error?

src/types/unsolicited_response.rs
```diff
@@ -0,0 +1,309 @@
...
+    ///
+    /// The server may unilaterally send `FETCH` responses, as described in
+    /// [RFC3501](https://tools.ietf.org/html/rfc3501#section-7.4.2).
+    Fetch {
+        /// Message identifier.
+        id: u32,
+        /// Attribute values for this message.
+        attributes: Vec<AttributeValue>,
+    },
+}
```
jonhoo, line 149: We really ought to order the variants here lexiographically now that there are so many.

src/types/unsolicited_response.rs
```diff
@@ -0,0 +1,309 @@
...
+                    }),
+                    _ => Err(response),
+                }
+            }
+            _ => Err(response),
+        }
+    }
+}
+
+/// Owned version of ResponseCode that wraps a subset of [`imap_proto::ResponseCode`]
```
jonhoo, line 243: Do we actually need this subset enum? Does the set of possible valid response codes vary between solicited and unsolicited OKs?

src/types/unsolicited_response.rs
```diff
@@ -0,0 +1,309 @@
...
+    Uid(Uid),
+}
+
+impl<'a> TryFrom<&ImapProtoAttributeValue<'a>> for AttributeValue {
+    type Error = ParseError;
+
+    fn try_from(val: &ImapProtoAttributeValue<'a>) -> Result<Self, Self::Error> {
+        match val {
+            ImapProtoAttributeValue::Flags(flags) => {
+                let v = flags.iter().map(|v| Flag::from(v.to_string())).collect();
```
jonhoo, line 289: Given the number of places we do this iter+map+Flag::from, maybe we should have a helper for it?

examples/idle.rs
```diff
@@ -0,0 +1,77 @@
...
+    let ssl_conn = TlsConnector::builder().build().unwrap();
+    let client = imap::connect((opt.server.clone(), opt.port), opt.server, &ssl_conn)
+        .expect("Could not connect to imap server");
+    let mut imap = client
+        .login(opt.username, opt.password)
+        .expect("Could not authenticate");
+    imap.debug = true;
+    imap.select(opt.mailbox).expect("Could not select mailbox");
+
+    let idle = imap.idle().expect("Could not IDLE");
```
jonhoo, line 52: How do you feel about having the callback be passed to `idle` instead of to the individual `wait_*` methods?

jonhoo (commented):
Overall this looks great! I've left some comments inline, though nothing _too_ major I don't think.

examples/idle.rs
```diff
@@ -0,0 +1,77 @@
...
+fn main() {
+    let opt = Opt::from_args();
+
+    let ssl_conn = TlsConnector::builder().build().unwrap();
+    let client = imap::connect((opt.server.clone(), opt.port), opt.server, &ssl_conn)
+        .expect("Could not connect to imap server");
+    let mut imap = client
+        .login(opt.username, opt.password)
+        .expect("Could not authenticate");
+    imap.debug = true;
```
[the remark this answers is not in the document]
mordak: Ah, good idea. :)

examples/idle.rs
```diff
@@ -0,0 +1,77 @@
...
+    #[structopt(short, long, default_value = "INBOX")]
+    mailbox: String,
+
+    #[structopt(
+        short = "x",
+        long,
+        help = "The number of responses to receive before exiting",
+        default_value = "5"
+    )]
+    max_responses: usize,
```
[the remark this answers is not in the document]
mordak: Hmm.. I get what you're saying about complexity, but let me try to convince you that it is fine as is? If I am unsuccessful, then I don't really object to making these const.

- These options all already have sensible default values, so you don't need to specify them.
- It is more consistent to have everything passed into the imap client config come from the same place.
- Having these as options is consistent with what someone who is cribbing from the example is going to do anyway.
- It is not really any more code to have these as options, since having them as options is just a variable declaration and a structopt macro, instead of a const decl and maybe a comment.

Reasonable? :)

src/extensions/idle.rs
```diff
@@ -40,6 +65,21 @@ pub enum WaitOutcome {
     MailboxChanged,
 }
 
+/// Return type for IDLE response callbacks. Tells the IDLE connection
+/// if it should continue monitoring the connection or not.
+#[derive(Debug, PartialEq, Eq)]
+pub enum CallbackAction {
```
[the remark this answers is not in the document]
mordak: I went back and forth on this a bit. I can go either way? The only reason I did this is because I find it easier to read and remember what the return value _means_ if it is a named value. I don't foresee any other behaviours, so this is purely in the service of being descriptive.

I feel like if I were coming back to callback code that I wrote in a year I would be looking at the `return true` or `return false` and then would have to look up what those meant, whereas `Stop` and `Continue` is fairly unambiguous?

Perhaps I am overthinking it? I have worked in codebases before with these exact same semantics (override this function and return true or false to indicate continue or not) and after some time we changed it to returning a named type because it was just more clear what was intended when coming back to modify code that hadn't been touched in awhile.

src/extensions/idle.rs
mordak: Yes, this would be a nice improvement! :)

src/types/unsolicited_response.rs
mordak: It does mean the server is about to close the connection, but the RFC also says the client [SHOULD continue to read responses off the wire](https://tools.ietf.org/html/rfc3501#section-7.1.5). So as is, this will turn into an unsolicited response, and then when the server will eventually hang up on us and the client will get an io Error.

I think the 'right' thing to do is to handle the `Bye` and then continue to read responses, which may be of the kind we're looking for or not, and then just let the server hang up on us. The client will get an error in both cases, but if we keep reading then the client may get a response between the `Bye` and the hangup. If we return the `Bye` as an error, then the client will probably terminate the connection and maybe lose those responses?

FWIW, I can go either way and would be happy for suggestions. As is, the IDLE client will get a callback with a `Bye` and then they can do whatever they want with it, and then they will eventually get an io Error if they choose to continue the connection. Since clients should be tolerant to sudden io errors anyway I don't think this will cause a problem in that way, but am not 100% sure if it will be a problem for other situations where we get an unsolicited `Bye` response while issuing a `SELECT` or something. Should we keep parsing after seeing the `Bye` (and maybe get a solicited response?) or return an Error to the user immediately?

src/types/unsolicited_response.rs
mordak: Ah, good call. :)

src/types/unsolicited_response.rs
mordak: The immediate reason this exists is to avoid exposing the `imap_proto::ResponseCode` in our public API.

In practice, there are a bunch of possible response codes that will never show up in an unsolicited response, like `BadCharset` and [`Capabilities`](https://github.com/djc/tokio-imap/blob/bb553c9512393c1a040e19194f95595958f0cb32/imap-proto/src/types.rs#L66), because these codes are specifically in response to a request. I would actually say that _most_ of the ResponseCodes will never show up in an unsolicited response, but it is less effort and more complete to just turn the `imap_proto::ResponseCode` into the owned version if we're willing to expose it in out public API.

I think it is less work for us to not wrap this, but this is really about whether or not we want to expose the `imap_proto` types directly. I am fine either way.

src/types/unsolicited_response.rs
mordak: Yay helpers! I will do this. :)

src/extensions/idle.rs
mordak: Sure, that's better. :)

src/extensions/idle.rs
mordak: In practice, we shouldn't. In theory, it is possible that the response parser does not consume the entire input and there are leftovers, so this is parsing the hypothetical leftovers.

The only reason we shouldn't get here is because we are testing on every `session.readline() ` if this is something we can parse or not. If not, we update the buffer, read another line, and try again. So as soon as the thing parses then we _should_ parse the entire input and will hit the `rest.is_empty()` clause and just clear the input vector. But if in the future we make some change or have some weird situation where the input parser does _not_ actually clear the entire input buffer, then this clause will correctly update the input.

Basically, this is effectively exhaustively matching what `rest` could be.

src/extensions/idle.rs
mordak: I did it this way to avoid lifetime issues since `rest` is a reference to the end of `v`, so am not sure how to slide the remaining elements down, but maybe this is just [drain()](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain), where we immediately throw the returned iterator away?

Also since we are unlikely to hit this path I didn't worry about it too much. I will see what I can do.

src/extensions/idle.rs
mordak: Yes, it's just anything we do not recognize as an `UnsolicitedResponse`, so anything we haven't explicitly handled.

I am not sure if there will be ownership issues pushing a complete `imap_proto::Response` into the unsolicited responses `channel` - ownership issues are the primary reason why previous PRs for this functionality have wrapped the various `imap_proto` types, and I do not see `into_owned` on `imap_proto::Response` (but could just be blind).

I _do_ think it is kind of weird and confusing to have two types of "Response" in `UnsolicitedResponse` though - the kind we wrap, and the kind we don't. If we want to just give the user the `imap_proto::Response` directly, then why bother wrapping at all? Why not just push `imap_proto::Response` into the unsolicited responses `channel` all the time? This would be vastly simpler for us since we would not have to wrap things at all and all the ownership questions would be resolved in only one way, and it would be simpler for the user since there would only be one "Response" to match against.

mordak: Re documentation: Do you just mean that what is meant by "unhandled response" should be more explicit? I can fix up the language there.

If you mean that it isn't clear that `UnsolicitedResponse` is incomplete, then where would be the best place to make that clear? The documentation for the `UnsolicitedResponse` type says "these are the ones that we have encountered in the wild", which _implies_ incompleteness, but doesn't come right out and say it. The documentation for `TryFrom<Response<'a>> for UnsolicitedResponse` explicitly says that not all possible `Response` variants are supported.

src/extensions/idle.rs
mordak: I was thinking this as well, but wasn't sure if I should do it here. I can remove it. :)

src/parse.rs
mordak: Strictly speaking, no - nothing in the IDLE spec that I have seen says that there are special Responses that only come on an IDLE connection. It just says that it will send unilateral responses and doesn't itemize what those responses will be outside of some examples which are not meant to be complete. As near as I can tell, IDLE is just saying that the client is signing up for arbitrary unilateral responses of any kind. In practice, the practical range of responses that will come on an Idle channel is smaller than the set of responses that will come otherwise (at least as far as I have seen).

examples/idle.rs
mordak: I think it makes more sense to pass it to the `wait` functions? Simply because when passing a closure it seems more normal to do this:

`
let idle = imap.idle().unwrap();

let result = idle.wait_keepalive(|response| {
    // Do something
});
`

instead of this:

`
let idle = imap.idle(|response| {
    // Do something
}).unwrap();

let result = idle.wait_keepalive();
`

And also passing it to the `wait` functions means we do not have to store the callback inside the `Handle`, but this is probably not a big deal.

Do you have a preference?

jonhoo: When I see it written out, no, I agree with you :)

src/extensions/idle.rs
jonhoo: Given that IMAP command responses must be terminated at newlines anyway, I think I'm comfortable saying that this simply shouldn't happen. Maybe add in a `debug_assert!(rest.is_empty())`? Also, I think I'd keep this just `else` — the `!= v.len()` bit should always be true.

src/types/mod.rs
```diff
@@ -168,6 +168,11 @@ impl Flag<'static> {
             _ => None,
         }
     }
+
+    /// Helper function to transform a [`Vec`] of flag strings into a [`Vec`] of [`Flag`].
+    pub fn from_vec<S: ToString>(v: &[S]) -> Vec<Self> {
```
jonhoo, line 173: Let's not enforce an allocation here (e.g., for passing this to `extend`). For maximal re-use, try:
```rust
impl Flag<'a> {
    pub fn from_strs<S>(v: impl IntoIterator<Item = S>) -> impl Iterator<Item = Self> where Flag<'a>: From<S> {
      v.into_iter().map(|s| Flag::from(s))
    }
}
```

examples/idle.rs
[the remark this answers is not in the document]
jonhoo: I'm sold :)

src/extensions/idle.rs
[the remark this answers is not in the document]
jonhoo: I could go either way on this — the extra type does also increase the API surface, but perhaps in a small enough way that the improved clarity is worth it. I think maybe it'd help me if the name was more descriptive as well. How about:
```rust
enum OnResponse {
```

src/extensions/idle.rs
jonhoo: Yeah, I'm torn here. I think you're right that `Unexpected` is a bit of a cop-out. I was thinking of it as a "these are the responses you might want to handle, and here's an escape hatch for if there's something we didn't think of", but maybe that escape hatch doesn't buy us much. Plus, it means that we're more likely to hear about missed variants, which is a plus. I guess then let's stick with dropping unexpected unsolicited responses, and expect `UnsolicitedResponse` to grow over time instead.

And yes, I think documentation-wise this should be updated to refer more clearly to what "unhandled response" means. I _also_ think it'd be good to be even more explicit in the docs for `UnsolicitedResponse` ­— maybe even link to the issue tracker and suggest that people file an issue if they see an unsolicited response that isn't captured by the enum — what do you think?

src/types/unsolicited_response.rs
jonhoo: I'm torn because by having a wrapper we also need to make sure we remember to add to this over time as new response codes are handled by `imap_proto`. But I also agree with you that it probably makes sense to only expose the possible subset here. Happy to defer to your preference here.

src/extensions/idle.rs
jonhoo: Yeah, I think `drain` is what you want here to remove the prefix of a `Vec`.

src/types/unsolicited_response.rs
jonhoo: I think you've convinced me — we should surface this as an unsolicited BYE and let the client choose how to deal with it :+1:

src/extensions/idle.rs
mordak: Hm. The default case here is if `v` does not parse as a complete command (multi-line response), so `rest` will not be empty and will be equal to `v` (nothing parsed). In this case we want to just loop again.

So we could just simplify the entire thing by replacing the if / else clause with this:
```
v.drain(0..(v.len() - rest.len()));
```

- If nothing parsed, then `rest == v`, and we drain `0..0`, so nothing.
- If we parsed a full command, then `rest.len() == 0` and we drain the whole `v`.
- Otherwise we drain some partial bit of `v`, and we think that shouldn't happen but we handle it anyway.

I think the only reason to _not_ do this would be if the common case (we drain the entire `v`) is much slower than just `v.clear()`. Looking at the impl of `Drain`, I think it is.

I think the way it is is actually optimal - the common case (all of `v` is parsed) is on the fast path with `v.clear()`; the uncommon case (multi-line response) is handled transparently, and we handle the unexpected case of a partial parse but don't expect to ever hit it. I could add a `debug_assert!()` inside the `rest.len() != v.len()` clause, but if we really don't think it's possible then we might as well just put a `panic!()` in there and not bother with `drain()` at all.

src/extensions/idle.rs
mordak: I like the idea of just making it explicit that `UnsolicitedResponse` is only covering things we actually expect to see, and adding a link to the issue tracker if someone finds something new. I changed the language a bit on `UnsolicitedResponse` to this effect, and then changed the language on the `wait_*` functions to be more clear and linked them to `UnsolicitedResponse`.

src/extensions/idle.rs
jonhoo: So, the reason I like a `debug_assert!` in there is that we get a panic to tell us that this _can_ happen, but we don't break "real" code and instead handle that correctly if it turns out that it _is_ possible. But I'm okay going whichever way you want with this.

I think the part I'm missing is why the `rest.len() != v.len()` check is necessary. Why can't it just be a simple `else`? If `rest.is_empty()` then there's no remainder in the input (we parsed it all), and the only alternative is that there _is_ more left in `rest`, which is what `else` would capture. What does the inequality check guard against?

src/extensions/idle.rs
```diff
@@ -193,8 +193,8 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
         }
     }
 
-    /// Block until the given callback returns `Stop`, or until an unhandled
-    /// response arrives on the IDLE channel.
+    /// Block until the given callback returns `Stop`, or until a response
+    /// arrives that is not explicitly handled by [`UnsolicitedResponse`].
```
jonhoo, line 197: Actually, here's a crazy though — maybe `WaitOutcome` should have a variant `Unexpected` and _that_ would hold the `imap_proto::Response` that was unexpected.

src/types/unsolicited_response.rs
mordak: Well, if we want to keep up with `imap_proto`, then the easiest way is to not wrap it at all, but this same argument eventually applies to `imap_proto::Response`. :-) 

For me this is really about whether or not we want to republish the `imap_proto` types. If we do, then we probably shouldn't wrap `ResponseCode` or `AttributeValue`. We already republish `StatusAttribute`, so if that's fine then it's probably fine to republish the others. It's just that in these cases the user will have to hypothetically deal with more variants than they will practically see, but that's fine and they already have to use a `_` arm because those types are non-exhaustive.

src/extensions/idle.rs
mordak: Oh, the only reason to have `rest.len() != v.len()` is to avoid going into this branch at all in the multi-line response case where we don't parse anything from the input (because we only need to do anything if `rest != v` so why waste cycles?).

But I think I see what you're getting at here - we can do the `debug_assert` that flags the unexpected behaviour and also does the right thing anyway. I didn't see it until I tried to implement what you were getting at. I pushed a change that I think is what you mean?

src/types/unsolicited_response.rs
jonhoo: I think we're unlikely to want to commit fully to having no `imap_proto` types in our public interface, so I think the measuring stick should be "do we gain something else useful by wrapping". For example, in the `UnsolicitedResponse` case it allows us to list (and thus document) _specifically_ the variants users can expect to come unsolicited. If it's a matter of one or two extra variants, it's not too bad, but half of them are impossible, wrapping seems worthwhile. It's definitely not an exact science.

src/extensions/idle.rs
mordak: I tried this, but quickly bumped into the reality that the `imap_proto::Response` is referring to data that was read off the wire in `wait_inner`. :)

So we'd have to make sure the stuff coming off the wire lives long enough to be returned to the `wait_*` caller. I guess the simplest way would be to have the caller pass in the buffer, or return a tuple `Unexpected(Vec<u8>, imap_proto::Response<'a>)`. Unless there is an owned version of `Response` that I do not see??

I get the impression you want the user to have the `imap_proto::Response` somehow. Is this so they can match on it, or do we just want them to know 'I saw this'. If we want them to match on it and do something with it, we could just give it to them in the callback instead of an `UnexpectedResponse`. The only reason we're using `UnexpectedResponse` here is because previous versions were sending it down the `unsolicited_responses` channel. Since we're not doing that anymore, we aren't actually bound to it, and it would make this diff a lot simpler. I am not sure I like the disconnect between getting an `imap_proto` type in the callback and `imap` types everywhere else, but maybe we want to see where that goes?

src/types/mod.rs
mordak: Ah, nice. Will do. :)

src/extensions/idle.rs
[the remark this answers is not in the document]
mordak: I think I'll just go with `bool` here. You don't seem sold, and I don't feel strongly enough to haggle over names :)

src/types/unsolicited_response.rs
mordak: Since it is easier to implement, easier to maintain, and also uses the `into_owned` bits that are obviously there to be used in `imap_proto`, I have changed this to just use the owned versions of `ResponseCode` and `AttributeValue` from `imap_proto`. :)

src/extensions/idle.rs
mordak: I added a ticket for this so we do not forget: #190

src/types/mod.rs
```diff
@@ -169,9 +169,11 @@ impl Flag<'static> {
         }
     }
 
-    /// Helper function to transform a [`Vec`] of flag strings into a [`Vec`] of [`Flag`].
-    pub fn from_vec<S: ToString>(v: &[S]) -> Vec<Self> {
-        v.iter().map(|s| Self::from(s.to_string())).collect()
+    /// Helper function to transform Strings into owned Flags
+    pub fn from_strs<S: ToString>(
```
jonhoo, line 173: I was specifically thinking we'd want `where Flag<'a>: From<S>` and place the `impl` on `Flag<'a>` (rather than `Flag<'static>`) so that this would automatically do the right thing both for `String` and for `&str`. The way it is now calling `from_strs` on an iterator over `String` will clone each `String` through a call to `to_string`. I _think_
```rust
impl Flag<'a> {
    pub fn from_strs<S>(v: impl IntoIterator<Item = S>) -> impl Iterator<Item = Self> where Flag<'a>: From<S> {
      v.into_iter().map(|s| Flag::from(s))
    }
}
```
should do it?

src/extensions/idle.rs
```diff
@@ -100,37 +130,74 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+            (_, result) => result,
         }
     }
 
-    /// Block until the selected mailbox changes.
-    pub fn wait(mut self) -> Result<()> {
-        self.wait_inner(true).map(|_| ())
+    /// Block until the given callback returns `false`, or until a response
+    /// arrives that is not explicitly handled by [`UnsolicitedResponse`].
+    pub fn wait<F>(mut self, callback: F) -> Result<()>
```
jonhoo, line 196: I think a good compromise here might be to name the function `wait_while` (and similarly for the other methods). That way it _should_ be clear what the `true`/`false` actually means. What do you think?

src/extensions/idle.rs
```diff
@@ -100,37 +130,74 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
-            // Handle Dovecot's imap_idle_notify_interval message
-            if v.eq_ignore_ascii_case(b"* OK Still here\r\n") {
+            // Update remaining data with unparsed data if needed.
+            if rest.is_empty() {
                 v.clear();
             } else {
-                break Ok(result);
+                // Assert on partial parse in debug builds - we expect to always parse all
+                // or none of the input buffer. On release builds, we still do the right thing.
+                debug_assert!(
```
jonhoo, line 172: Continuing the discussion from https://github.com/jonhoo/rust-imap/pull/186#discussion_r606870729.

Yes, this is what I meant!

Looking at this version of the change, I think I now understand better what bothers me about this. We're using `rest.len() != v.len()` as a way of checking "is this a complete parse", when we already have a good place to do this — the `None` case for `match parse_idle`! I suspect this could would be much nicer if we simply moved all the bits in the `if` here up into the appropriate match arms above.

src/extensions/idle.rs
jonhoo: You know, I think I just straight-up forgot to add an `into_owned` method for `imap_proto::Response` in https://github.com/djc/tokio-imap/pull/115. That was an oversight, and could trivially be added here:
https://github.com/djc/tokio-imap/blob/bb553c9512393c1a040e19194f95595958f0cb32/imap-proto/src/types.rs#L46

and would look roughly like this:
https://github.com/djc/tokio-imap/blob/bb553c9512393c1a040e19194f95595958f0cb32/imap-proto/src/types.rs#L102

May be worth just submitting a PR there and then we could give out `imap_proto::Response<'static>`!

Well, so, I want the user to be handed an `imap_proto::Response` _only_ if that response doesn't fit into what we expect should be possible responses. You can sort of think of it as a "recoverable error". Basically, in my head, `WaitOutcome` is sort of like a `Result<Timeout, imap_proto::Response>` where the `Response` is yielded if a response was encountered that doesn't fit into `UnsolicitedResponse`, and thus _shouldn't_ be possible.

> I am not sure I like the disconnect between getting an `imap_proto` type in the callback and `imap` types everywhere else, but maybe we want to see where that goes?

Ah, no, that's not what I'm proposing. I'm proposing that the callback continues to get `UnsolicitedResponse`, but that `wait` returns an `imap_proto::Response` _if_ it sees an unexpected response (i.e., one that should be able to come unsolicited).

src/types/mod.rs
mordak: Hmm. I did try with the version you had pasted in there, but it is hard to actually use because everywhere we want to use this we need a copying version because the destination is `'static`, and we usually are being given a `Cow` here that has non static lifetime.

We need an owning version in `UnsolicitedResponse` because it has to send it down a `channel`. So that's that.

For the uses in `parse.rs`, we are *already* copying ([here](https://github.com/jonhoo/rust-imap/blob/4108917c58c4fe5083418b00b6a337e8d3b04ece/src/parse.rs#L113), [here](https://github.com/jonhoo/rust-imap/blob/4108917c58c4fe5083418b00b6a337e8d3b04ece/src/parse.rs#L275) and [here](https://github.com/jonhoo/rust-imap/blob/4108917c58c4fe5083418b00b6a337e8d3b04ece/src/parse.rs#L275)). And we're already copying because we actually need to. Or, at least I could not make the compiler happy, since it complained about not being able to resolve `'a` from the source and `'static` from the dest. I assume this is also why this code is copying already.

So since everywhere we were using this we ended up copying, and we were being given a `Cow`, I didn't end up using the non-copying version, so I removed it as dead code.

Does this make sense? It is possible I just couldn't make it work and am missing something obvious.

jonhoo: Oh, I see what you mean. That's unfortunate, but does make sense — I don't think you're missing anything. I'll stop pushing my crazy generics on you :sweat_smile:

src/extensions/idle.rs
mordak: Oh, I am not super torqued about it? I feel like either way you'll have to read the docs to know for sure. And I like the simplicity of `wait` :)

I was also thinking we could just provide a `pub const CONTINUE: bool = true`, and `pub const STOP: bool = false`, but this feels slightly silly? It's super normal in C/C++, but doesn't seem super normal in Rust?

src/extensions/idle.rs
mordak: Ahh. Okay. The reason the parse / match is separate is because I usually like to separate buffer management from parse result handling. Inside the parse match arms you're usually holding a ref to the buffer you parsed out of, so sometimes it's a hassle to try to do buffer management right there.

I pushed a change that integrates the buffer management (with debug_assert on unexpected partial parse), since we can get away with it here. I also added some comments on the match arms to indicate what they're handling.

src/extensions/idle.rs
mordak: Having `into_owned` on `imap_proto::Result` would solve a lot of these kinds of problems! :)

I can have a stab at this on the weekend? Unless you want to have a go.

Having unexpected responses returned sounds reasonable. So that's all good then. Do we want to hold this PR open until the owned `Response` thing gets sorted, or add it in a later PR?

src/extensions/idle.rs
jonhoo: I think I prefer `wait_while` since it also leans on the precedent set in the standard library with things like `take_while`. While `wait` is nice and short, it's not _super_ descriptive of what the function now actually does, whereas I think `wait_while` is.

So, I've seen that occasionally in Rust, but only defined in the _user_ of the library, rarely in the library itself. As in, `imap` wouldn't define those `const`s, but rather the user of `imap` would define it to make their calls clearer. But I think with a rename to `wait_while`, that's already clear enough!

src/extensions/idle.rs
jonhoo: Beautiful, yes, I think that reads _much_ better. I agree with you in general that it's nice to do the buffer management separately, but I think in this instance the code reads much more clearly this way.

src/extensions/idle.rs
```diff
@@ -150,34 +150,35 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+                                    v,
+                                    rest
+                                );
+                                let used = v.len() - rest.len();
+                                v.drain(0..used);
+                            }
                         }
-                        (rest, None) => rest,
+                        // Incomplete parse - do nothing and read more.
+                        (_rest, None) => (),
```
jonhoo, line 177: Nit, but this is usually written as `=> {}`.

src/extensions/idle.rs
```diff
@@ -150,34 +150,35 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
                             }
-                            rest
+                            if rest.is_empty() {
+                                v.clear();
+                            } else {
+                                // Assert on partial parse in debug builds - we expect
+                                // to always parse all or none of the input buffer.
+                                // On release builds, we still do the right thing.
+                                debug_assert!(
+                                    rest.len() != v.len(),
```
jonhoo, line 167: I _think_ that with this reorganization this _must_ now be true, no? In which case I _think_ it makes more sense to make this `debug_assert!(rest.is_empty()` and place it up above the `if rest.is_empty()`. Alternatively, just make this `debug_assert!(false`.

src/extensions/idle.rs
mordak: np, nits are all good :)

src/extensions/idle.rs
mordak: Done. I went with the pattern `wait_thing_while()` when there were more words in the function signature, instead of `wait_while_thing`, since the latter seems misleading (ie. `wait_while_keepalive` is confusing).

I also only did the public functions, since I don't think the internal ones don't need the usage hint?

Since we are changing the API it might be nice to switch this to a builder pattern so there is only one wait, ie.:

```
session.idle()
    .keepalive()
    .wait_while(|res| {
        // do something
    })
```

But this can happen later if at all .

src/extensions/idle.rs
mordak: Done!

src/extensions/idle.rs
mordak: So I went to do this, and ... we *already* return unhandled `Response`s in the `Err` path. Specifically, as `Err(Error::Parse(ParseError::Unexpected(format!("{:?}", err))))` (from [here](https://github.com/jonhoo/rust-imap/blob/4108917c58c4fe5083418b00b6a337e8d3b04ece/src/error.rs#L115)). We're stringifying the unexpected `Response` instead of giving an owned version back to the caller (but we could make that a non-stringy version with an owned `Response` if we wanted).

And now it seems weird to instead be returning `Ok(WaitOutcome::Unexpected(Response))` when this is actually really more of an error condition (ie. this *shouldn't happen*), and we're already doing it as an `Err`.

Thoughts?

jonhoo: Oh, nice! Yeah, let's just return the owned response right there. Although maybe not as a `ParseError`? That feels a little odd. Though I understand why it used to be under there.

src/extensions/idle.rs
jonhoo: Thank you! I agree builder would be nice here. Happy to do in this PR or in a follow-up — up to you.

src/extensions/idle.rs
mordak: PR for owned `Response` is [here](https://github.com/djc/tokio-imap/pull/120). Had to do `MailboxDatum` at the same time. :)

Am I correct that the reason this is under `Error::ParseError` is because it seems to mostly (entirely?) come out of the various `parse_*` functions. So if we make  this just `Error::Unexpected` I think the `ParseError::Unexpected` variant will never be constructed and can be removed?

It makes sense to me to separate "I could not parse this thing" from "I parsed it but do not know what to do with it", so moving `Unexpected` from a `ParseError` variant to an `Error` variant makes sense to me. Also makes sense if we want to keep it as a `ParseError` variant because "I was parsing and something went wrong". Semantics are fun! :)

src/extensions/idle.rs
mordak: Let's do a follow up, since it'll be fine as a standalone and we the TLS one to do as well.

src/extensions/idle.rs
jonhoo: Yes, that's exactly right. We also used to roll our own parsing, and those errors got wrapped up in `ParseError`, but now that we use `imap_proto` that's also a little different. I _think_ a top-level `Unexpected` variant is the way to go here. And if we then have no more `ParseError` that make sense, then I suppose we should get rid of it! The `enum` is `#[non_exhaustive]` anyway, so we could always add it back in later.

jonhoo:
Re our discussion about `bool` vs a rich type for continue vs break, it looks like maybe we'll get that type directly in the standard library! See https://github.com/rust-lang/rfcs/pull/3058 and specifically [the `ControlFlow` type](https://github.com/scottmcm/rfcs/blob/do-or-do-not/text/0000-try-trait-v2.md#opscontrolflow).

jonhoo:
Better yet: https://github.com/rust-lang/rust/issues/75744

scottmcm:
I have high hopes that `ControlFlow` will stabilize soon™ -- the major thing blocking it is [unwanted interconversions](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=11a2c7018d39a644668068463cb4d1a1), which will be fixed as soon as RFC 3058 goes in on nightly.  (It oughtn't need to wait for full `Try` stabilization.)

src/extensions/idle.rs
mordak: Just updating - I have something for this queued up, but waiting on the imap-proto change to make it into a release. I asked today when that might be, so MTF.

mordak:
> Re our discussion about `bool` vs a rich type for continue vs break, it looks like maybe we'll get that type directly in the standard library! See [rust-lang/rfcs#3058](https://github.com/rust-lang/rfcs/pull/3058) and specifically [the `ControlFlow` type](https://github.com/scottmcm/rfcs/blob/do-or-do-not/text/0000-try-trait-v2.md#opscontrolflow).

Oh, that will be nice!

src/extensions/idle.rs
mordak: Pushed an update for this change just now.

jonhoo:
Let's maybe open a ticket to consider moving to `ControlFlow` if it lands in time for 3.0.0?

jonhoo (approved):
I think we're finally there! Thanks for the awesome work :heart:

PR #192, opened 2021-04-17
Support parsing STATUS responses.

### LINKED ISSUE #185: Client::status does not work
`Client::status` does not parse the response correctly and always returns a default `Mailbox` object.

Suppose I call `session.status("INBOX", "(UIDNEXT UIDVALIDITY)")`.  This causes the following to be sent:
```
a2 STATUS "INBOX" (UIDNEXT UIDVALIDITY)
```
The server (Dovecot) responds:
```
* STATUS INBOX (UIDNEXT 7 UIDVALIDITY 1616297529)
```
`Client::status` calls `parse_mailbox`, which parses the above into the following `imap_proto::Response`:
```
MailboxData(Status { mailbox: "INBOX", status: [UidNext(7), UidValidity(1616297529)] }))
```

But `parse_mailbox` only accepts these attributes if seen as a `ResponseCode` within `Response::Data`.  When it encounters a `Response::MailboxData` containing `MailboxDatum::Status`, it always treats it as unsolicited:
```rust
            Ok((rest, Response::MailboxData(m))) => {
                lines = rest;

                match m {
                    MailboxDatum::Status { mailbox, status } => {
                        unsolicited
                            .send(UnsolicitedResponse::Status {
                                mailbox: mailbox.into(),
                                attributes: status,
                            })
                            .unwrap();
                    }
```

comex:
Fixes #185.

My use of ParseError::Unexpected for error messages feels a bit sketchy,
but I wasn't sure how you'd prefer to handle them.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/192)

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1,4 +1,4 @@
-use imap_proto::{MailboxDatum, Response, ResponseCode};
+use imap_proto::{MailboxDatum, Response, ResponseCode, StatusAttribute};
 use lazy_static::lazy_static;
 use regex::Regex;
 use std::collections::HashSet;
@@ -313,6 +313,53 @@ pub fn parse_mailbox(
             break Ok(mailbox);
         }
     }
+}
+
+pub fn parse_status(
+    mut lines: &[u8],
+    mailbox_name: &str,
+    unsolicited: &mut mpsc::Sender<UnsolicitedResponse>,
+) -> Result<Mailbox> {
+    let mut mailbox = Mailbox::default();
+    let mut got_anything = false;
+    while !lines.is_empty() {
+        match imap_proto::parser::parse_response(lines) {
+            Ok((
+                rest,
+                Response::MailboxData(MailboxDatum::Status {
+                    mailbox: their_mailbox_name,
+                    status,
+                }),
+            )) if their_mailbox_name == mailbox_name => {
+                lines = rest;
+                got_anything = true;
+                for attr in status {
+                    match attr {
+                        StatusAttribute::HighestModSeq(v) => mailbox.highest_mod_seq = Some(v),
+                        StatusAttribute::Messages(v) => mailbox.exists = v,
+                        StatusAttribute::Recent(v) => mailbox.recent = v,
+                        StatusAttribute::UidNext(v) => mailbox.uid_next = Some(v),
+                        StatusAttribute::UidValidity(v) => mailbox.uid_validity = Some(v),
+                        StatusAttribute::Unseen(v) => mailbox.unseen = Some(v),
+                        _ => {} // needed because StatusAttribute is #[non_exhaustive]
+                    }
+                }
+            }
+            Ok((rest, data)) => {
+                lines = rest;
+                if let Some(resp) = try_handle_unilateral(data, unsolicited) {
+                    return Err(resp.into());
+                }
+            }
+            _ => {
+                return Err(Error::Parse(ParseError::Invalid(lines.to_vec())));
+            }
+        }
+    }
+    if !got_anything {
+        return Err(Error::MissingStatusResponse);
+    }
+    Ok(mailbox)
 }
 
 fn parse_ids_with<T: Extend<u32>>(
--- a/tests/imap_integration.rs
+++ b/tests/imap_integration.rs
@@ -9,6 +9,7 @@ use lettre::Transport;
 use std::net::TcpStream;
 
 use crate::imap::extensions::sort::{SortCharset, SortCriterion};
+use crate::imap::types::Mailbox;
 
 fn tls() -> native_tls::TlsConnector {
     native_tls::TlsConnector::builder()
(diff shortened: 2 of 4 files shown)


src/parse.rs
```diff
@@ -321,6 +321,69 @@ pub fn parse_mailbox(
     }
 }
 
+pub fn parse_status(
```
jonhoo, line 324: We really should write a macro to help write these `parse_*` methods @mordak. They all have basically the same outer structure. Doesn't need to happen as part of this PR though.

src/parse.rs
```diff
@@ -321,6 +321,69 @@ pub fn parse_mailbox(
...
+    mut lines: &[u8],
+    mailbox_name: &str,
+    unsolicited: &mut mpsc::Sender<UnsolicitedResponse>,
+) -> Result<Mailbox> {
+    let mut mailbox_opt = None;
+
+    loop {
+        if lines.is_empty() {
+            break mailbox_opt.ok_or_else(|| {
```
jonhoo, line 333: Can this just be `while !lines.is_empty()` with the `mailbox_opt.ok_or_else` placed below the loop?

src/parse.rs
```diff
@@ -321,6 +321,69 @@ pub fn parse_mailbox(
...
+                        StatusAttribute::Unseen(v) => mailbox.unseen = Some(v),
+                        _ => {} // needed because StatusAttribute is #[non_exhaustive]
+                    }
+                }
+                mailbox_opt = Some(mailbox);
+            }
+            Ok((rest, data)) => {
+                lines = rest;
+                if let Some(resp) = try_handle_unilateral(data, unsolicited) {
+                    break Err(resp.into());
```
jonhoo, line 377: There is a sort of weird case where we enter this case and `mailbox_opt.is_some()`. I'm not sure if the right thing to do in that instance is to return an `Err` or just ignore the spurious response and give the `Ok`. What do you think?

tests/imap_integration.rs
```diff
@@ -390,3 +390,12 @@ fn append_with_flags_and_date() {
...
     assert_eq!(inbox.len(), 0);
 }
+
+#[test]
+fn status() {
+    let mut s = session("readonly-test@localhost");
+    let mb = s.status("INBOX", "(UIDNEXT UIDVALIDITY)").unwrap();
+    assert!(mb.uid_next.is_some());
+    assert!(mb.uid_validity.is_some());
+    assert!(mb.unseen.is_none());
```
jonhoo, line 400: Are there more things we can test for here as well that we know will be in the response?

jonhoo:
Thanks for taking this up! I left a few notes inline.

src/parse.rs
comex: Well, even if `mailbox_opt` is `None` (i.e. we haven't gotten the expected `STATUS` response yet), that response could still show up later in `lines`.  In particular, if the unknown response is some sort of unsolicited data sent on the server's initiative and unrelated to the `STATUS` command, then the ordering between it and the `STATUS` response would be a matter of chance.

It might be a good idea to make unknown untagged responses non-fatal in all contexts.  The IMAP spec does say: "A client MUST be prepared to accept any server response at all times.  This includes server data that was not requested."  This wording doesn't explicitly differentiate between known response types (that simply don't correspond to requests) and unknown response types, but arguably it could be read to cover both.

But I don't think it makes sense to make them non-fatal only in this case.

comex:
I updated the PR:
- Rebased on latest master.
- Addressed PR comments.
- Changed to be more tolerant of unexpected server responses, as the IMAP spec requires:
  - Receiving a STATUS response for the wrong mailbox is no longer an error; it's appended to `unsolicited` like any other unsolicited response.  
  - Receiving multiple STATUS responses for the right mailbox is no longer an error.  It's possible that one of them is the "real" response, and the other ones are unsolicited STATUS responses for the same mailbox, which the server just happened to send around the same time.
  - On the other hand, we can't just stick unsolicited responses with a matching mailbox name into `unsolicited`, because we have no way of knowing which response is the "real" one (if that's even a well-defined concept).  In theory, we could guess that the "real" response is the last response that has (at least) all of the requested attributes, and put other responses into `unsolicited`.  But this would complicate the code, and I doubt it would actually benefit any practical use case.  As a simpler alternative, just merge attributes from all received STATUS responses with a matching mailbox name.  This is the same behavior that `parse_mailbox` already has.

src/parse.rs
jonhoo: I think it's probably okay to return it as an error, because the caller can always then decide what to do next. If they don't want to treat "weird" server messages as errors, they have that option. It only really gets weird in the case where there is _also_ successful return data for the command, since then all the caller can do is call the operation again. But that doesn't work if the operation isn't idempotent.

Honestly not sure what the right thing to do here is, but I agree with you that it's broader than the scope of this PR, so resolving.

cc @mordak if you want to think about what the right thing to do here is too.

jonhoo (approved):
I'm happy with this now! Would you mind also opening a PR that adds this change to the changelog? Thanks!

PR #196, opened 2021-04-24
Add test for dovecot unsolicited OK progress message.

mordak:
Follow up on #179.

We seem to have handled this with the recent work on `UnsolicitedResponse`, so I just added a test to make sure we continue to handle this.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/196)

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -505,6 +505,28 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


jonhoo (commented):
Good call, thanks!

Issue #190, opened 2021-04-08
Make TLS connections easier to set up


mordak:
Reference [here](https://github.com/jonhoo/rust-imap/pull/186#discussion_r606870111), it would be nice if setting up TLS connections was easier to do. Perhaps with a nice builder pattern like [here](https://github.com/jonhoo/fantoccini/blob/2d1e24acc337e1f0c36e543eb52a20538f1dc18a/src/lib.rs#L180-L196).

PR #197, opened 2021-05-06
Add ClientBuilder helper to make setting up TLS connections easy.

mordak:
Addresses #190 

Also add support for rustls in addition to native-tls. I think using `rustls` should be easier (and could even be the default?), so I just made it first class like `native-tls`. I have tested both the `native_tls()` and `rustls()` paths.

This just adds the `ClientBuilder`. If we're happy with the API here and the way it works, then I can take care of:
- Fixing up the examples (doc and `/examples`) to use it where appropriate.
- If we want to get rid of any code that is superseded by this?

I have not tested the STARTTLS code, since I do not have a STARTTLS enabled server to work with. How do we want to handle automating the tests to make sure it all works?

I have not added it, but I can see an argument for adding some knobs to control which CA bundle is added, or adding private CAs or something. The alternative it to just leave the existing API as is, which allows users to make their own TLS connection using whatever configuration that they want.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/197)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,14 @@ required-features = ["default"]
 name = "idle"
 required-features = ["default"]
 
+[[example]]
+name = "starttls"
+required-features = ["default"]
+
+[[example]]
+name = "timeout"
+required-features = ["default"]
+
 [[test]]
 name = "imap_integration"
 required-features = ["default"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,10 @@ name = "idle"
 required-features = ["default"]
 
 [[example]]
+name = "rustls"
+required-features = ["rustls-tls"]
+
+[[example]]
 name = "starttls"
 required-features = ["default"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,10 +15,12 @@ categories = ["email", "network-programming"]
 
 [features]
 tls = ["native-tls"]
+rustls-tls = ["rustls-connector"]
 default = ["tls"]
 
 [dependencies]
 native-tls = { version = "0.2.2", optional = true }
+rustls-connector = { version = "0.13.1", optional = true }
 regex = "1.0"
 bufstream = "0.1"
 imap-proto = "0.14.1"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -75,6 +75,18 @@ jobs:
        greenmail: greenmail
      env:
        TEST_HOST: greenmail
+ - job: features
+   displayName: "Check feature combinations"
+   pool:
+     vmImage: ubuntu-latest
+   steps:
+     - template: install-rust.yml@templates
+       parameters:
+         rust: stable
+     - script: cargo install cargo-hack
+       displayName: install cargo-hack
+     - script: cargo hack --feature-powerset check --all-targets
+       displayName: cargo hack
 
 resources:
   repositories:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,13 +25,13 @@ jobs:
      - template: install-rust.yml@templates
        parameters:
          rust: $(rust)
-     - script: cargo check --all-targets
+     - script: cargo check --all-targets --all-features
        displayName: cargo check
-     - script: cargo test --examples
+     - script: cargo test --examples --all-features
        displayName: cargo test --examples
-     - script: cargo test --doc
+     - script: cargo test --doc --all-features
        displayName: cargo test --doc
-     - script: cargo test --lib
+     - script: cargo test --lib --all-features
        displayName: cargo test --lib
      - script: |
          set -e
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,11 +1,8 @@
 use bufstream::BufStream;
 use chrono::{DateTime, FixedOffset};
 use imap_proto::Response;
-#[cfg(feature = "tls")]
-use native_tls::{TlsConnector, TlsStream};
 use std::collections::HashSet;
 use std::io::{Read, Write};
-use std::net::{TcpStream, ToSocketAddrs};
 use std::ops::{Deref, DerefMut};
 use std::str;
 use std::sync::mpsc;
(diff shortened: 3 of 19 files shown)


src/client.rs
```diff
@@ -414,6 +414,14 @@ impl<T: Read + Write> Client<T> {
         }
     }
 
+    /// Yield the underlying connection for this Client.
+    /// This consumes `self` since the Client is not much use without
```
jonhoo, line 418:
```suggestion
    /// Yield the underlying connection for this Client.
    ///
    /// This consumes `self` since the Client is not much use without
```

src/client_builder.rs
```diff
@@ -0,0 +1,106 @@
+use crate::{Client, Result};
+use std::net::TcpStream;
+
+#[cfg(feature = "native-tls")]
```
jonhoo, line 4: Given that the feature is called `tls` (the dependency is `native-tls`), I think this should be `tls` everywhere. Or we should re-jig the `Cargo.toml` so that the feature is called `native-tls`.

src/client_builder.rs
```diff
@@ -0,0 +1,106 @@
...
+            let ssl_conn = RustlsConnector::new_with_native_certs()?;
+            let tls = ssl_conn.connect(self.domain.as_ref(), client.into_inner()?)?;
+            Ok(Client::new(tls))
+        } else {
+            let ssl_conn = RustlsConnector::new_with_native_certs()?;
+            let tls = ssl_conn.connect(self.domain.as_ref(), tcp)?;
+            Ok(Client::new(tls))
+        }
+    }
+}
```
jonhoo, line 106: Let's also provide something like
```rust
pub fn connect<F, C>(&mut self, handshake: F) -> Result<Client<C>>
  where F: FnOnce(&str, TcpStream) -> Result<C>,
        C: Read + Write
```

That way users can construct their own connection in custom ways if they so desire.

jonhoo: And, in fact, let's have the other two methods call that one, so that we can share the STARTTLS code.

jonhoo (changes requested):
I like the approach! I left a suggestion in there that I think can deal with the more customized setup routines users may have. As for testing STARTTLS, I thought our integration test already did that? If not, we should add one.

src/client_builder.rs
mordak: Oops. My bad. Will use `tls` everywhere.

src/client_builder.rs
mordak: Oh this is a good idea!

mordak: I think this actually came out much nicer. :)

mordak:
> As for testing STARTTLS, I thought our integration test already did that? If not, we should add one.

We have one, but it is marked [`ignore`](https://github.com/jonhoo/rust-imap/blob/1db492ed3baac999e9dfbc4802481ef9a3b2fd2e/tests/imap_integration.rs#L56). Is there another one?

src/client_builder.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// and yield a `Result<C>` where `C` is `Read + Write`. It should only perform
+    /// TLS initialization over the given `tcp` socket and return the encrypted stream
+    /// object, such as a [`native_tls::TlsStream`] or a [`rustls_connector::TlsStream`].
+    ///
+    /// If the caller is using `STARTTLS` and previously called [`starttls`](Self::starttls)
+    /// then the `tcp` socket given to the `handshake` function will be connected and will
+    /// have initiated the `STARTTLS` handshake.
+    ///
+    /// ```no_run
+    /// # use imap::ClientBuilder;
```
jonhoo, line 108: This probably needs
```rust
#![cfg(feature = "rustls-tls")]
```
So that `cargo test --doc --no-default-features`
```
Doesn't fail.

File: src/client_builder.rs [hunk 5]
```diff
@@ -0,0 +1,138 @@
...
+#[cfg(feature = "tls")]
+use native_tls::{TlsConnector, TlsStream};
+#[cfg(feature = "rustls-tls")]
+use rustls_connector::{RustlsConnector, TlsStream as RustlsStream};
+
+/// A convenience builder for [`Client`] structs over various encrypted transports.
+///
+/// Creating a [`Client`] using `native-tls` transport is straightforward:
+/// ```no_run
+/// # use imap::ClientBuilder;
```
jonhoo, line 14: I think this needs
```rust
#![cfg(feature = "tls")]
```

src/client_builder.rs
```diff
@@ -0,0 +1,138 @@
...
+/// # fn main() -> Result<(), imap::Error> {
+/// let client = ClientBuilder::new("imap.example.com", 993).native_tls()?;
+/// # Ok(())
+/// # }
+/// ```
+///
+/// Similarly, if using the `rustls-tls` feature you can create a [`Client`] using rustls:
+/// ```no_run
+/// # use imap::ClientBuilder;
```
jonhoo, line 23: I think this needs
```rust
#![cfg(feature = "rustls-tls")]
```

src/client_builder.rs
```diff
@@ -0,0 +1,138 @@
...
+/// # fn main() -> Result<(), imap::Error> {
+/// let client = ClientBuilder::new("imap.example.com", 993).rustls()?;
+/// # Ok(())
+/// # }
+/// ```
+///
+/// To use `STARTTLS`, just call `starttls()` before one of the [`Client`]-yielding
+/// functions:
+/// ```no_run
+/// # use imap::ClientBuilder;
```
jonhoo, line 33: I think this needs
```rust
#![cfg(feature = "rustls-tls")]
```

azure-pipelines.yml
```diff
@@ -25,13 +25,13 @@ jobs:
...
+     - script: cargo check --all-targets --all-features
        displayName: cargo check
-     - script: cargo test --examples
+     - script: cargo test --examples --all-features
        displayName: cargo test --examples
-     - script: cargo test --doc
+     - script: cargo test --doc --all-features
        displayName: cargo test --doc
-     - script: cargo test --lib
+     - script: cargo test --lib --all-features
```
jonhoo, line 34: Now that we have more features, I'd like us to start using [`cargo hack`](https://github.com/taiki-e/cargo-hack) to test all combinations of features. There's an example job template for it [over in inferno's CI](https://github.com/jonhoo/inferno/blob/43ed7801cdea55af186c054fe8280c6ca426bff3/azure-pipelines.yml#L11-L22).

jonhoo (changes requested):
Ugh, you're right, I forgot that Greenmail [doesn't support STARTTLS](https://github.com/greenmail-mail-test/greenmail/issues/135). That's awkward. I suppose we'll just have to leave that one for users to test then :cry:

azure-pipelines.yml
mordak: Oh, this is nice. I added this to the azure pipeline. The `cargo hack` line works for me locally, if it blows up in Azure then I'll figure it out.

src/client_builder.rs
mordak: Haha! Almost all the doc tests fail with `cargo test --doc --no-default-features`. 

I fixed all of them. I am not really super happy with all of them? Just using `#![cfg(feature = "xyz")]` to blank out the whole doc test makes the test fail because there is no main and no body. So you have to have something there, so I ended up just scoping the parts of the test that used the features so the test compiles with / without the feature.

The alternative way to do this would be to include some dummy code under `#[cfg(not(feature = "xyz"))]` so it compiles both ways. Do you have a preference?

jonhoo: Ugh, that's annoying.. I _feel_ like from memory there's a little bit of magic sauce if you place `#![cfg(feature = "")]` on the very first line of the doctest, but I could be wrong. I do think I'd prefer the `#![cfg]` approach rather than strategically placing extra blocks for `#[cfg()]` though.

jonhoo:
Ugh, separately, we really should add a codecov configuration file that sets less onerous restrictions for PRs Currently they basically always fail. Something like [this one](https://github.com/jonhoo/left-right/blob/master/codecov.yml) (docs [here](https://docs.codecov.io/docs/coverage-configuration)).

src/client_builder.rs
mordak: Yeah, I am kind of surprised that it doesn't just accept an empty doc test block? This seems like the sort of thing that should work, but somehow using `#![cfg(feature = "xyz")]` isn't accepted. I would be happy to be educated here.

Anyway, I pushed something that I think will work and be minimally ugly:

`# {} #[cfg(feature = "tls")]`

The `{}` satisfies the 'something is there' requirement, and the `cfg` line will include / exclude the entire doc `fn main`.

mordak:
> Ugh, separately, we really should add a codecov configuration file that sets less onerous restrictions for PRs Currently they basically always fail. Something like [this one](https://github.com/jonhoo/left-right/blob/master/codecov.yml) (docs [here](https://docs.codecov.io/docs/coverage-configuration)).

Happily added. Having the codecov always fail is a bit annoying. I copied the one you linked and tweaked it for the values that we're seeing in PRs here (ie 70% coverage low line, and don't fail for -2%). No problem if you want to change them to be tighter.

mordak:
Separately - what, if anything, do we want to move to the builder API?
- Doc examples?
- Examples?
- Anything in the lib?

I can see some of the docs and examples being kind of obvious. For the lib, some of the `Client` setup stuff could be moved the builder API internally while preserving the public API, or we could remove the public API in favour of the builder API, or we could just leave the lib as is and call it a day. Do you have an inclination here?

src/client_builder.rs
jonhoo: That's... awful :sweat_smile: But sure, I still think it's better. I feel like this might be worthwhile to file a bug in rust-lang/rust about actually.

jonhoo:
I think my preference here is actually to _only_ have the builder API + `Client::new` for custom streams. What do you think?

src/client_builder.rs
mordak: Yeah, I find this kind of unsatisfying and it seems like the kind of thing that should work.

I checked through the rust-lang/rust issue tracker but did not see anything like this. If you have messed around with this a bit and confirmed what I am seeing, then we can put together a bug report.

mordak:
> Ugh, you're right, I forgot that Greenmail [doesn't support STARTTLS](https://github.com/greenmail-mail-test/greenmail/issues/135). That's awkward. I suppose we'll just have to leave that one for users to test then 😢

I put together a local test for this and confirmed the STARTTLS implementation in the `ClientBuilder` works.

mordak:
> I think my preference here is actually to _only_ have the builder API + `Client::new` for custom streams. What do you think?

Oh that sounds good to me. I prefer having one good way to do it: it is less code to maintain and nicer to for users.

I pushed a diff that removes `connect()` and `connect_starttls()` and updates the examples, tests and docs to use the `ClientBuilder` instead. I think the only possible wrinkle is the greenmail tests, which may not be happy with the certificate validation. If that's the case then I can figure it out.

src/client_builder.rs
jonhoo: I think what I'd do is include in the report:
```rust
/// ```
/// #![cfg(feature = "foo")]
/// assert_eq!(1, 1);
/// ```
pub struct Foo;
```
with the error message suggesting "consider adding a `main` function at the crate level" (which is clearly wrong), and then
```
/// ```rust
/// #![cfg(feature = "foo")]
/// fn main() {
/// assert_eq!(1, 1);
/// }
/// ```
pub struct Foo;
```
which gives the _same_ error even though there _is_ an explicit `fn main` there. At the very least, this is a compiler error problem (cc @estebank), but I would argue it's _also_ a bug/missing feature.

Cargo.toml
```diff
@@ -15,10 +15,12 @@ categories = ["email", "network-programming"]
 
 [features]
 tls = ["native-tls"]
+rustls-tls = ["rustls-connector"]
```
jonhoo, line 18: Separately, I wonder if we want to take this opportunity to rename the "other" `tls` feature `native-tls`, since it isn't really "more" standard than `rustls`.

jonhoo (approved):
Looks great, thanks!

src/client_builder.rs
jonhoo: @estebank I'm hoping to coax @mordak into doing it since I want more people to get into the habit of reporting unhelpful UX errors, even if they're unsure about whether the underlying feature _should_ be supported. @mordak if you have a chance to write one up, that'd be great, otherwise I'd be happy to do it!

mordak: No problem, done!

https://github.com/rust-lang/rust/issues/85239

Cargo.toml
mordak: I think this is a good idea, so did a PR up [here](https://github.com/jonhoo/rust-imap/pull/201).

PR #201, opened 2021-05-23
Change "tls" feature to "native-tls"

mordak:
Suggested [here](https://github.com/jonhoo/rust-imap/pull/197#discussion_r629806956) and it's a good idea. :-)

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/201)

Issue #191, opened 2021-04-11
IDLE Builder


mordak:
It might be nice to switch the Idle setup to a builder pattern so there is only one wait, ie.:

```
session.idle()
    .keepalive()
    .wait_while(|res| {
        // do something
    })
```

_Originally posted by @mordak in https://github.com/jonhoo/rust-imap/pull/186#discussion_r610227188_

PR #202, opened 2021-05-23
Idle builder

mordak:
Addresses #191 .

Also includes the changes in #201 for convenience.

Note that this change effectively requires that the underlying transport implements the `SetReadTimeout` trait from `extensions::idle`. I am not sure if this is a problem or not. If so, I can add the old `wait_while()` back with a different name (`wait_notimeout_while()`?) to restore the functionality. I opted to remove it because the old `wait_keepalive_while` function was already broken for non-`SetReadTimeout` transports, and it was the recommended way to do it. I am unsure if the extra API is worth it since implementing `SetReadTimeout` is so trivial. Happy to get feedback on this.

Note that this also makes `keepalive` the default, since it is the recommended way to do it. An opt-out function is provided (`no_keepalive()`) for users who do not want this. I figured that the recommended path should also be the default.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/202)

### DIFF
--- a/src/extensions/idle.rs
+++ b/src/extensions/idle.rs
@@ -206,7 +206,7 @@ impl<'a, T: SetReadTimeout + Read + Write + 'a> Handle<'a, T> {
     /// Do not continuously refresh the IDLE connection in the background.
     ///
     /// By default, connections will periodically be refreshed in the background using the
-    /// timeout duration set by [`timeout`]. If you do not want this behaviour, call
+    /// timeout duration set by [`Handle::timeout`]. If you do not want this behaviour, call
     /// this function and the connection will simply IDLE until `wait_while` returns or
     /// the timeout expires.
     pub fn keepalive(&mut self, keepalive: bool) -> &mut Self {
--- a/src/extensions/idle.rs
+++ b/src/extensions/idle.rs
@@ -48,7 +48,7 @@ use std::time::Duration;
 /// such a server has an inactivity timeout it MAY log the client off implicitly at the end of its
 /// timeout period. Because of that, clients using IDLE are advised to terminate the IDLE and
 /// re-issue it at least every 29 minutes to avoid being logged off. This is done by default, but
-/// can be disabled by calling [`Handle::no_keepalive`]
+/// can be disabled by calling [`Handle::keepalive`]
 ///
 /// As long as a [`Handle`] is active, the mailbox cannot be otherwise accessed.
 #[derive(Debug)]
@@ -198,7 +198,7 @@ impl<'a, T: SetReadTimeout + Read + Write + 'a> Handle<'a, T> {
     /// at which the connection is refreshed.
     ///
     /// The interval defaults to 29 minutes as given in RFC 2177.
-    pub fn set_timeout(&mut self, interval: Duration) -> &mut Self {
+    pub fn timeout(&mut self, interval: Duration) -> &mut Self {
         self.timeout = interval;
         self
     }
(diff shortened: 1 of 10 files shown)


src/extensions/idle.rs
```diff
@@ -196,73 +191,55 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+    /// Do not continuously refresh the IDLE connection in the background.
     ///
-    /// The interval defaults to 29 minutes as dictated by RFC 2177.
-    pub fn set_keepalive(&mut self, interval: Duration) {
-        self.keepalive = interval;
+    /// By default, connections will periodically be refreshed in the background using the
+    /// timeout duration set by [`set_timeout`]. If you do not want this behaviour, call
+    /// this function and the connection will simply IDLE until `wait_while` returns or
+    /// the timeout expires.
+    pub fn no_keepalive(&mut self) -> &mut Self {
```
jonhoo, line 212: Let's make this: `fn keepalive(&mut self, enabled: bool)`.

src/extensions/idle.rs
```diff
@@ -196,73 +191,55 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
-
-    /// Block until the given given amount of time has elapsed, the given callback
-    /// returns `false`, or until a response arrives that is not explicitly handled
-    /// by [`UnsolicitedResponse`].
-    pub fn wait_with_timeout_while<F>(self, timeout: Duration, callback: F) -> Result<WaitOutcome>
-    where
-        F: FnMut(UnsolicitedResponse) -> bool,
-    {
-        self.timed_wait(timeout, false, callback)
+        self.timed_wait(callback).map(|_| ())
```
jonhoo, line 231: At this point is there even a point in having `timed_wait` be its own method?

src/extensions/idle.rs
```diff
@@ -196,73 +191,55 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
-    }
 }
 
 impl<'a, T: SetReadTimeout + Read + Write + 'a> Handle<'a, T> {
-    /// Set the keep-alive interval to use when `wait_keepalive_while` is called.
+    /// Set the timeout duration on the connection. This will also set the frequency
+    /// at which the connection is refreshed.
+    ///
+    /// The interval defaults to 29 minutes as given in RFC 2177.
+    pub fn set_timeout(&mut self, interval: Duration) -> &mut Self {
```
jonhoo, line 201: I think [the convention for builders](https://rust-lang.github.io/api-guidelines/type-safety.html#c-builder) is to not include the `set_` prefix.

jonhoo (changes requested):
Nice! I think it makes complete sense to make keep-alive the default, and I think it's fine to require `SetReadTimeout` :+1:

src/extensions/idle.rs
mordak: Done! I actually had it this way before, but changed it because it will only ever be used one way (to set the opposite value from the default), so the argument felt redundant.

src/extensions/idle.rs
mordak: Oh, I didn't know this existed! Thanks!

src/extensions/idle.rs
mordak: Not really. Do we want to propagate the `WaitOutcome` back to the caller since it's right there anyway?

mordak:
I think we are good to go on this one?

src/extensions/idle.rs
jonhoo: So, _technically_ there's a difference in that someone could choose to re-use a builder and want no keepalive at first but then keepalive subsequently. But yeah, it seems unlikely. Even so, I think sticking to the standard pattern is probably the way to go, if only for consistency.

jonhoo (approved):
Excellent, thanks!

Separately, we really need to figure out what causes that CI failure all the time. Maybe it's time to drop the FreeBSD compile check altogether..

Issue #205, opened 2021-06-29
[Question] Removing lifetime parameter from imap::types::Flag?


TornaxO7:
Hello!

I'm helping in writing a [mail-cli program](https://github.com/soywod/himalaya) and for that I need to store some mails in the memory for better performance (for example, if you want to read mail1 and then mail2 quickly). This also means, that I want to store the flags of the mails as well but it's not really possible to store the output of [`imap::types::Fetch::flags()`](https://docs.rs/imap/2.4.1/imap/types/struct.Fetch.html#method.flags) due to the lifetime-parameter so I'm wondering if it's planned to remove it.

Otherwise I think that I'll write an extra-enum which represents your Flag-enum but it uses `Custom(String)` instead to make it "independent" from the lifetime or do you have any other suggestions how I could avoid/fix this problem?

jonhoo:
#206 is exactly how I was going to propose we fix this :)

jonhoo:
Ah, yes, in the latest 3.0.0 alpha we do have `Flag::into_owned` now!

PR #206, opened 2021-06-30
Flag static

TornaxO7:
Adding a function which will makes it possible to convert a `Flag<'a>` into a
`Flag<'static>` in order to be able to store flags.

This should also fix [my issue](https://github.com/jonhoo/rust-imap/issues/205).

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/206)

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -177,6 +177,24 @@ impl Flag<'static> {
     }
 }
 
+impl<'flag> Flag<'flag> {
+    /// Converts a `Flag<'flag>` into a `Flag<'static>` or in other words: It's
+    /// making sure that `Flag::Custom` includes a `String` and not a `&str`
+    /// anymore.
+    pub fn into_static(&self) -> Flag<'static> {
+        match self {
+            Flag::Seen => Flag::Seen,
+            Flag::Answered => Flag::Answered,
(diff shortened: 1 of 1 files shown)


src/types/mod.rs
```diff
@@ -177,6 +177,24 @@ impl Flag<'static> {
     }
 }
 
+impl<'flag> Flag<'flag> {
+    /// Converts a `Flag<'flag>` into a `Flag<'static>` or in other words: It's
+    /// making sure that `Flag::Custom` includes a `String` and not a `&str`
+    /// anymore.
+    pub fn into_static(&self) -> Flag<'static> {
```
jonhoo, line 184: We've generally been calling these methods [`into_owned`](https://docs.rs/imap/3.0.0-alpha.4/imap/types/enum.AttributeValue.html#method.into_owned), but otherwise very much on board with this change!

TornaxO7:
Oh, I didn't knew that it was already implemented in the `imap-3.0.0-alpha.4` version. Should I close this pull request then?

jonhoo:
Ah, no, the `into_owned` method doesn't exist for `Flag` in 3.0.0, I was just pointing out that that's the name we've used for _other_ types that have similar functionality :)

PR #208, opened 2021-07-06
Convert ZeroCopy to ouroboros.

### LINKED ISSUE #125: Move from ZeroCopy to the rental crate
The `imap::ZeroCopy` struct is effectively an implementation of the "rental pattern", which already has a good [crate implementation](https://docs.rs/rental/). It would be good to switch to the `rental` crate, but it would also constitute a breaking change, as `rental` creates a separate struct for each wrapped type.

mordak:
This converts `Capabilities` to use ouroboros instead of the local `ZeroCopy`.

Because of all the magic that ouroboros `self_referencing` macro adds (`CapabiltiesTryBuilder` and friends) I found it was simpler to move the parsing bit into the `Capabilities` struct itself. I actually prefer this, since it keeps everything related to `Capabilities` in one place, but I understand if you prefer to keep everything related to parsing in `parse.rs`.

Anyway, if this approach for moving from `ZeroCopy` to ouroboros looks like a good approach, then I can do the rest of it similarly here, and `ZeroCopy` can go away, and we can close #125 .

Thoughts?

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/208)

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -3,6 +3,7 @@ use lazy_static::lazy_static;
 use regex::Regex;
 use std::collections::HashSet;
 use std::convert::TryFrom;
+use std::iter::Extend;
 use std::sync::mpsc;
 
 use super::error::{Error, ParseError, Result};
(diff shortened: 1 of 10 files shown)


jonhoo:
Yes, this looks very promising indeed! Also cc @joshua-maros just because they may be interested in an application of their crate :)

joshua-maros:
Thanks for the mention, I always appreciate seeing something I made help someone out! As a side note, I've recently published `0.10.0` which fixes a potential source of undefined behavior, it should be a drop-in replacement for your use case.

mordak:
Pushed an update the switches `Name` and `Fetch` over to ouroboros. This lets us remove `ZeroCopy`.

src/parse.rs
```diff
@@ -23,105 +24,51 @@ pub fn parse_authenticate_response(line: &str) -> Result<&str> {
...
-            for attr in &fetch.fetch {
-                match attr {
-                    AttributeValue::Flags(flags) => {
-                        fetch.flags.extend(Flag::from_strs(flags));
-                    }
-                    AttributeValue::Uid(uid) => fetch.uid = Some(*uid),
-                    AttributeValue::Rfc822Size(sz) => fetch.size = Some(*sz),
-                    _ => {}
+                match map(resp)? {
+                    MapOrNot::Map(t) => into.extend(Some(t)),
```
jonhoo, line 57: I personally prefer `std::iter::once` for this, but not for any particularly good reason. Your call!

src/parse.rs
```diff
@@ -488,30 +403,31 @@ mod tests {
...
         let lines = b"* LIST (\\HasNoChildren) \".\" \"INBOX\"\r\n";
         let (mut send, recv) = mpsc::channel();
-        let names = parse_names(lines.to_vec(), &mut send).unwrap();
+        let names = Names::parse(lines.to_vec(), &mut send).unwrap();
         assert!(recv.try_recv().is_err());
         assert_eq!(names.len(), 1);
+        let first = names.iter().next().unwrap();
         assert_eq!(
-            names[0].attributes(),
+            first.attributes(),
```
jonhoo, line 419: I think it'd be hugely valuable to continue to support indexing here if at all possible. I suspect all that'd be needed is to implement `Index` for `Names` and `Fetches`.

src/types/fetch.rs
```diff
@@ -24,16 +100,13 @@ pub struct Fetch {
...
-    // Note that none of these fields are *actually* 'static. Rather, they are tied to the lifetime
-    // of the `ZeroCopy` that contains this `Name`. That's also why they can't be public -- we can
-    // only return them with a lifetime tied to self.
-    pub(crate) fetch: Vec<AttributeValue<'static>>,
+    pub(crate) fetch: Vec<AttributeValue<'a>>,
     pub(crate) flags: Vec<Flag<'static>>,
 }
 
-impl Fetch {
+impl<'a> Fetch<'a> {
```
jonhoo, line 107: Now that `Fetch` has a lifetime parameter, I'd like to have a `into_owned` method on it to get a `Fetch<'static>` from any `Fetch<'a>`.

src/types/flag.rs
```diff
@@ -0,0 +1,108 @@
...
+/// A flag can be permanent or session-only on a per-flag basis. Permanent flags are those which
+/// the client can add or remove from the message flags permanently; that is, concurrent and
+/// subsequent sessions will see any change in permanent flags.  Changes to session flags are valid
+/// only in that session.
+///
+/// > Note: The `\Recent` system flag is a special case of a session flag.  `\Recent` can not be
+/// > used as an argument in a `STORE` or `APPEND` command, and thus can not be changed at all.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Flag<'a> {
```
jonhoo, line 17: We should add a `Flag::into_owned` as well.

src/types/name.rs
```diff
@@ -34,7 +92,7 @@ pub enum NameAttribute<'a> {
     Custom(Cow<'a, str>),
 }
 
-impl NameAttribute<'static> {
+impl<'a> NameAttribute<'a> {
```
jonhoo, line 95: Why was this changed? This should be able to still return `<'static>`.

src/types/name.rs
```diff
@@ -76,9 +134,9 @@ impl<'a> From<&'a str> for NameAttribute<'a> {
     }
 }
 
-impl Name {
+impl<'a> Name<'a> {
```
jonhoo, line 137: Now that this has a lifetime, we should add `Name::into_owned` as well.

jonhoo (changes requested):
This is such an exciting change! I left some notes inline. Also just a heads up that I'll be out on vacation for a few weeks, so may take a while to get back to this.

mordak:
All of those comments look entirely reasonable! I'll be able to get it done soonish, and it should be gtg for when you get back from vacation.

src/types/fetch.rs
mordak: Done!

src/parse.rs
mordak: TIL `std::iter::once` :)

src/types/flag.rs
mordak: Done!

src/types/name.rs
mordak: Oops. :)

src/types/name.rs
mordak: Done!

src/parse.rs
mordak: I struggled to implement `Index` or `Deref` for this. The issue is the lifetime param on the `Output` associated type, since `type Output = Name<'a>`. The compiler wants the same lifetime on `Output`, but [generic associated types are unstable](https://github.com/rust-lang/rust/issues/44265).

We could probably get around this by replacing `Names` and `Fetches` with a generic container that takes a type param (we could call it.. `ZeroCopy`?), and then implement `Deref` or `Index` the same way that it is done for `ZeroCopy`, since the lifetime would be encapsulated in the generic type. Then `Names` is just `ZeroCopy<Name<'_>>`. When I try this I end up struggling against the constructor function signatures that ouroboros wants in the `TryBuilder` structure to initialize the self ref structure.

So instead I implemented a `get()` function, which trivially maps the underlying vector `get()` function and lets you just get an indexed member from the container. I personally kind of prefer this anyway, since `Index` can panic, but I get that the `Index` trait gives nice `[]` syntax.

If we want to impl `Index` or `Deref` (I think `Deref` would be better?), I am not sure how to make this work in the ouroboros context. I imagine that if it is doable then it is probably kind of simple, but I couldn't get it going.

jonhoo: Oof, yeah, that's awkward. You can `impl<'a> Index<usize> for &'a Fetches` ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7e4293bf8789913779e2ab9264a52ce4)), but it means `&fetches[0]` won't work, only `(&fetches)[0]`, which is both hard to discover and ugly. I'm okay with `fn get` :+1:

jonhoo (approved):
Looks good to me, thanks! Are there other things you wanted to change since this still stands as WIP?

mordak:
> Looks good to me, thanks! Are there other things you wanted to change since this still stands as WIP?

Nope! Let me fix the title there... :)

jonhoo:
Looks like the tarpaulin failure is due to https://github.com/xd009642/tarpaulin/issues/756 btw.

PR #212, opened 2021-07-26
Handle BYE responses explicitly.

### LINKED ISSUE #210: Support BYE response status
This example code:
```
fn main() {
    print!("{}", fetch_inbox_top().unwrap().unwrap());
}

fn fetch_inbox_top() -> imap::error::Result<Option<String>> {
    let domain = "imap.wp.pl";
    let tls = native_tls::TlsConnector::builder().build().unwrap();

    // we pass in the domain twice to check that the server's TLS
    // certificate is valid for the domain we're connecting to.
    let client = imap::connect((domain, 993), domain, &tls).unwrap();

    // the client we have here is unauthenticated.
    // to do anything useful with the e-mails, we need to log in
    let mut imap_session = client
        .login("<USERNAME>", "<PASSWORD>")
        .map_err(|e| e.0)?;

    // we want to fetch the first email in the INBOX mailbox
    imap_session.select("INBOX")?;

    // fetch message number 1 in this mailbox, along with its RFC822 field.
    // RFC 822 dictates the format of the body of e-mails
    let messages = imap_session.fetch("1", "RFC822")?;
    let message = if let Some(m) = messages.iter().next() {
        m
    } else {
        return Ok(None);
    };

    // extract the message's body
    let body = message.body().expect("message did not have a body!");
    let body = std::str::from_utf8(body)
        .expect("message was not valid utf-8")
        .to_string();

    // be nice to the server and log out
    imap_session.logout()?;

    Ok(Some(body))
}
```
throws exception on the statement `imap_session.logout()?;`:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Parse(Invalid([97, 52, 32, 66, 89, 69, 32, 73, 77, 65, 80, 52, 114, 101, 118, 49, 32, 83, 101, 114, 118, 101, 114, 32, 108, 111, 103, 103, 105, 110, 103, 32, 111, 117, 116, 13, 10]))'
```
After decoding content of `Parse(Invalid([...]))` as ASCII:
```
a4 BYE IMAP4rev1 Server logging out
```
It looks like the BYE response status is not supported.

mordak:
In Session::logout(), ignore it. Fixes #210.

I also added a test for the untagged BYE response followed by tagged OK, since that is what the spec says we should get.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -8,7 +8,7 @@ use std::str;
 use std::sync::mpsc;
 
 use super::authenticator::Authenticator;
-use super::error::{Bad, Error, No, ParseError, Result, ValidateError};
+use super::error::{Bad, Bye, Error, No, ParseError, Result, ValidateError};
 use super::extensions;
 use super::parse::*;
 use super::types::*;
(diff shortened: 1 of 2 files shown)


jonhoo (approved):
Looks great, thanks!

PR #218, opened 2021-10-30
Improve error message for `ValidationError`

### LINKED ISSUE #215: Better error messages for validation errors
Hi. I accidentally added a newline to a password and my program errored-out with the very undescriptive message:

```
Invalid character in input: '\n'
```

resp

```
Validate(ValidateError('\n'))
```

I think a message along the lines of 

```
Invalid character '\n' in argument 'password' of command 'LOGIN userid password'
```

resp

```
Validate(ValidateError { command: "LOGIN userid password", argument: "password", offending_char: '\n' })
```

would be more helpful. 

I could propose a corresponding pull request, if you are interested.

lu-fennell:
In addition to the invalid character, the error now also contains a
`command_synopsis` and `argument` which identifies the command and
argument that failed to validate.

Fixes #215 

Currently the error message just shows a "synopsis" of the command, e.g., `FETCH seq query`, not the actual arguments. Maybe it makes sense to also show the invalid argument value.. except if it is a secret of course. If you think this is a good idea, and provided this PR goes into the right direction, I can add this functionality also.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/218)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -41,8 +41,15 @@ impl<E> OptionExt<E> for Option<E> {
 /// grammar](https://tools.ietf.org/html/rfc3501#section-9)
 /// calls "quoted", which is reachable from "string" et al.
 /// Also ensure it doesn't contain a colliding command-delimiter (newline).
-pub(crate) fn validate_str(value: &str) -> Result<String> {
-    validate_str_noquote(value)?;
+///
+/// The arguments `synopsis` and `arg_name` are used to construct the error message of
+/// [ValidateError] in case validation fails.
+pub(crate) fn validate_str(
+    synopsis: impl Into<String>,
+    arg_name: impl Into<String>,
+    value: &str,
+) -> Result<String> {
+    validate_str_noquote(synopsis, arg_name, value)?;
     Ok(quote!(value))
 }
 
@@ -61,13 +68,26 @@ pub(crate) fn validate_str(value: &str) -> Result<String> {
 /// >             "BODY.PEEK" section ["<" number "." nz-number ">"]
 ///
 /// Note the lack of reference to any of the string-like rules or the quote characters themselves.
-fn validate_str_noquote(value: &str) -> Result<&str> {
+///
+/// The arguments `synopsis` and `arg_name` are used to construct the error message of
+/// [ValidateError] in case validation fails.
+fn validate_str_noquote(
+    synopsis: impl Into<String>,
+    arg_name: impl Into<String>,
+    value: &str,
+) -> Result<&str> {
     value
         .matches(|c| c == '\n' || c == '\r')
         .next()
         .and_then(|s| s.chars().next())
-        .map(|offender| Error::Validate(ValidateError(offender)))
-        .err()?;
+        .err()
+        .map_err(|c| {
+            Error::Validate(ValidateError {
+                command_synopsis: synopsis.into(),
+                argument: arg_name.into(),
+                offending_char: c,
+            })
+        })?;
     Ok(value)
 }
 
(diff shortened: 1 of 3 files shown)


src/error.rs
```diff
@@ -252,24 +252,53 @@ impl StdError for ParseError {
...
 #[derive(Debug)]
-pub struct ValidateError(pub char);
+pub struct ValidateError {
+    /// the synopsis of the invalid command
+    pub command_synopsis: String,
+    /// the name of the invalid argument
+    pub argument: String,
+    /// the invalid character contained in the argument
+    pub offending_char: char,
+}
```
jonhoo, line 265: I think we should make these all non-`pub` — they're not really useful to users except insofar as they are included in the printed representation.

This is also a breaking change because previously this struct had only public fields, and thus had a public constructor and could be matched against, which is kind of awkward. I don't know if anyone previously actually _inspected_ a `ValidateError`, but it's possible. I think I'm okay with including this in a minor release though since we're still in alpha.

src/client.rs
```diff
@@ -347,8 +377,11 @@ impl<T: Read + Write> Client<T> {
         username: U,
         password: P,
     ) -> ::std::result::Result<Session<T>, (Error, Client<T>)> {
-        let u = ok_or_unauth_client_err!(validate_str(username.as_ref()), self);
-        let p = ok_or_unauth_client_err!(validate_str(password.as_ref()), self);
+        let synopsis = "LOGIN username password";
```
jonhoo, line 380: It's really weird to me that the synopsis always includes the argument list, and then the argument is repeated again. I would prefer to have synopsis just be "command", and then emit:

> Invalid character {:?} in argument 'username' of command 'LOGIN'

As long as the parameter name matches the name of an argument to a function in `imap`, I think users will be able to backtrack where the error occurred without us making the errors needlessly verbose.

jonhoo (changes requested):
This is a _fantastic_ change and PR, thank you!

I left two comments inline that I think we should address, but otherwise this is good to land.

src/error.rs
lu-fennell: Ok, the fields are `pub(crate)` now.

Thanks for the hint about the breaking change... I totally missed that.

src/client.rs
lu-fennell: I see your point. I have removed the arguments from the "synopsis".

jonhoo (approved):
Excellent, thank you!

jonhoo:
Now we just need CI to cooperate. I've just pushed a commit to `master` that should fix some of it. The tarpaulin failure is due to https://github.com/xd009642/tarpaulin/issues/756, which we're still bit by only because we need to update our dependencies. Bumping the currently-outdated dependencies in `Cargo.toml` should take care of it (though let's do that in a separate PR):

```console
$ cargo outdated -d1
Name              Project  Compat  Latest  Kind    Platform
----              -------  ------  ------  ----    --------
imap-proto        0.14.3   ---     0.15.0  Normal  ---
nom               6.2.1    ---     7.1.0   Normal  ---
ouroboros         0.9.5    ---     0.13.0  Normal  ---
regex             1.4.6    ---     1.5.4   Normal  ---
rustls-connector  0.13.1   ---     0.15.0  Normal  ---
```

The only change that actually seems related to this PR is the failure of `rustfmt` _on beta_. That's optional to fix, though good to do ahead of time now that we know about it.

jonhoo:
(though please do rebase onto or merge with master just to make sure we didn't miss any regressions)

lu-fennell:
Ok, done: rebased and fixed formatting.

jonhoo (approved):
Thanks!

PR #222, opened 2022-02-15
Bump various dependencies & edition 2021

nbittich:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/222)


jonhoo:
I did some other cleanup too, and ended up doing this as well as part of that, so these changes are now on `master` :+1:

PR #225, opened 2022-04-06
Add support for Gmail Labels

CrispinStichart:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/225)

### DIFF
--- a/src/types/fetch.rs
+++ b/src/types/fetch.rs
@@ -219,6 +219,17 @@ impl<'a> Fetch<'a> {
         })
     }
 
+    /// Extract the `X-GM-LABELS` of a `FETCH` response
+    ///
+    /// This is a Gmail-specific extension. See their
+    /// [developer's page](https://developers.google.com/gmail/imap/imap-extensions) for details.
+    pub fn gmail_labels(&'a self) -> Option<impl Iterator<Item = &'a str>> {
+        self.fetch.iter().find_map(|av| match av {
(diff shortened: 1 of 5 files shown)


src/types/fetch.rs
```diff
@@ -219,6 +220,17 @@ impl<'a> Fetch<'a> {
         })
     }
 
+    /// Extract the `X-GM-LABELS` of a `FETCH` response
+    ///
+    /// This is a Gmail-specific extension. See their
+    /// [developer's page](https://developers.google.com/gmail/imap/imap-extensions) for details.
+    pub fn gmail_labels(&self) -> Option<&Vec<Cow<'a, str>>> {
```
jonhoo, line 227: I think this will actually end up being nicer to use:
```suggestion
    pub fn gmail_labels(&self) -> Option<impl Iterator<Item = &'a str>> {
```

jonhoo (changes requested):
Just one note, otherwise this looks good! Makes me glad we didn't release 3.0.0 yet :sweat_smile: 

I do worry about the fact that we directly expose `imap_proto` types in our API though, since it means we need to cut new major versions whenever `imap_proto` does. May be worth re-wrapping their types ourselves. We have a couple of types this affects, like `Envelope`, `BodyStructure`, and now `NameAttribute`. What do you think @mordak?

src/types/fetch.rs
CrispinStichart: Updated. Pretty new to rust, not sure how idiomatic it is to return a `Map` directly like I'm doing, or if there's a better way entirely.

PR #227, opened 2022-04-23
add in ACL extension support

urkle:
Requires djc/tokio-imap#142 (now merged, awaiting crate release) to handle the parsing-side of the responses.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/227)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,7 +48,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: -- --ignored
+          args: --features test-full-imap
     services:
       cyrus_imapd:
         image: outoforder/cyrus-imapd-tester:latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,13 +16,15 @@ categories = ["email", "network-programming"]
 [features]
 rustls-tls = ["rustls-connector"]
 default = ["native-tls"]
+# Used to activate full integration tests when running against a more complete IMAP server
+test-full-imap = []
 
 [dependencies]
 native-tls = { version = "0.2.2", optional = true }
 rustls-connector = { version = "0.16.1", optional = true }
 regex = "1.0"
 bufstream = "0.1.3"
-imap-proto = "0.16.0"
+imap-proto = "0.16.1"
 nom = { version = "7.1.0", default-features = false }
 base64 = "0.13"
 chrono = { version = "0.4", default-features = false, features = ["std"]}
--- a/src/client.rs
+++ b/src/client.rs
@@ -374,10 +374,10 @@ impl<T: Read + Write> Client<T> {
     /// }
     /// # }
     /// ```
-    pub fn login<U: AsRef<str>, P: AsRef<str>>(
+    pub fn login(
         mut self,
-        username: U,
-        password: P,
+        username: impl AsRef<str>,
+        password: impl AsRef<str>,
     ) -> ::std::result::Result<Session<T>, (Error, Client<T>)> {
         let synopsis = "LOGIN";
         let u =
@@ -432,10 +432,10 @@ impl<T: Read + Write> Client<T> {
     ///     };
     /// }
     /// ```
-    pub fn authenticate<A: Authenticator, S: AsRef<str>>(
+    pub fn authenticate(
         mut self,
-        auth_type: S,
-        authenticator: &A,
+        auth_type: impl AsRef<str>,
+        authenticator: &impl Authenticator,
     ) -> ::std::result::Result<Session<T>, (Error, Client<T>)> {
         ok_or_unauth_client_err!(
             self.run_command(&format!("AUTHENTICATE {}", auth_type.as_ref())),
@@ -528,7 +528,7 @@ impl<T: Read + Write> Session<T> {
     /// [`Connection::run_command_and_read_response`], you *may* see additional untagged `RECENT`,
     /// `EXISTS`, `FETCH`, and `EXPUNGE` responses. You can get them from the
     /// `unsolicited_responses` channel of the [`Session`](struct.Session.html).
-    pub fn select<S: AsRef<str>>(&mut self, mailbox_name: S) -> Result<Mailbox> {
+    pub fn select(&mut self, mailbox_name: impl AsRef<str>) -> Result<Mailbox> {
         self.run(&format!(
             "SELECT {}",
             validate_str("SELECT", "mailbox", mailbox_name.as_ref())?
@@ -540,7 +540,7 @@ impl<T: Read + Write> Session<T> {
     /// however, the selected mailbox is identified as read-only. No changes to the permanent state
     /// of the mailbox, including per-user state, will happen in a mailbox opened with `examine`;
     /// in particular, messagess cannot lose [`Flag::Recent`] in an examined mailbox.
-    pub fn examine<S: AsRef<str>>(&mut self, mailbox_name: S) -> Result<Mailbox> {
+    pub fn examine(&mut self, mailbox_name: impl AsRef<str>) -> Result<Mailbox> {
         self.run(&format!(
             "EXAMINE {}",
             validate_str("EXAMINE", "mailbox", mailbox_name.as_ref())?
@@ -606,11 +606,11 @@ impl<T: Read + Write> Session<T> {
     ///  - `RFC822.HEADER`: Functionally equivalent to `BODY.PEEK[HEADER]`.
     ///  - `RFC822.SIZE`: The [RFC-2822](https://tools.ietf.org/html/rfc2822) size of the message.
     ///  - `UID`: The unique identifier for the message.
-    pub fn fetch<S1, S2>(&mut self, sequence_set: S1, query: S2) -> Result<Fetches>
-    where
-        S1: AsRef<str>,
-        S2: AsRef<str>,
-    {
+    pub fn fetch(
+        &mut self,
+        sequence_set: impl AsRef<str>,
+        query: impl AsRef<str>,
+    ) -> Result<Fetches> {
         if sequence_set.as_ref().is_empty() {
             Fetches::parse(vec![], &mut self.unsolicited_responses_tx)
         } else {
@@ -626,11 +626,11 @@ impl<T: Read + Write> Session<T> {
 
     /// Equivalent to [`Session::fetch`], except that all identifiers in `uid_set` are
     /// [`Uid`]s. See also the [`UID` command](https://tools.ietf.org/html/rfc3501#section-6.4.8).
-    pub fn uid_fetch<S1, S2>(&mut self, uid_set: S1, query: S2) -> Result<Fetches>
-    where
-        S1: AsRef<str>,
-        S2: AsRef<str>,
-    {
+    pub fn uid_fetch(
+        &mut self,
+        uid_set: impl AsRef<str>,
+        query: impl AsRef<str>,
+    ) -> Result<Fetches> {
         if uid_set.as_ref().is_empty() {
             Fetches::parse(vec![], &mut self.unsolicited_responses_tx)
         } else {
@@ -688,7 +688,7 @@ impl<T: Read + Write> Session<T> {
     /// the mailbox UNLESS the new incarnation has a different unique identifier validity value.
     /// See the description of the [`UID`
     /// command](https://tools.ietf.org/html/rfc3501#section-6.4.8) for more detail.
-    pub fn create<S: AsRef<str>>(&mut self, mailbox_name: S) -> Result<()> {
+    pub fn create(&mut self, mailbox_name: impl AsRef<str>) -> Result<()> {
         self.run_command_and_check_ok(&format!(
             "CREATE {}",
             validate_str("CREATE", "mailbox", mailbox_name.as_ref())?
@@ -714,7 +714,7 @@ impl<T: Read + Write> Session<T> {
     /// incarnation, UNLESS the new incarnation has a different unique identifier validity value.
     /// See the description of the [`UID`
     /// command](https://tools.ietf.org/html/rfc3501#section-6.4.8) for more detail.
-    pub fn delete<S: AsRef<str>>(&mut self, mailbox_name: S) -> Result<()> {
+    pub fn delete(&mut self, mailbox_name: impl AsRef<str>) -> Result<()> {
         self.run_command_and_check_ok(&format!(
             "DELETE {}",
             validate_str("DELETE", "mailbox", mailbox_name.as_ref())?
@@ -746,7 +746,7 @@ impl<T: Read + Write> Session<T> {
     /// to a new mailbox with the given name, leaving `INBOX` empty.  If the server implementation
     /// supports inferior hierarchical names of `INBOX`, these are unaffected by a rename of
     /// `INBOX`.
-    pub fn rename<S1: AsRef<str>, S2: AsRef<str>>(&mut self, from: S1, to: S2) -> Result<()> {
+    pub fn rename(&mut self, from: impl AsRef<str>, to: impl AsRef<str>) -> Result<()> {
         self.run_command_and_check_ok(&format!(
             "RENAME {} {}",
             quote!(from.as_ref()),
@@ -762,7 +762,7 @@ impl<T: Read + Write> Session<T> {
     /// The server may validate the mailbox argument to `SUBSCRIBE` to verify that it exists.
     /// However, it will not unilaterally remove an existing mailbox name from the subscription
     /// list even if a mailbox by that name no longer exists.
-    pub fn subscribe<S: AsRef<str>>(&mut self, mailbox: S) -> Result<()> {
+    pub fn subscribe(&mut self, mailbox: impl AsRef<str>) -> Result<()> {
         self.run_command_and_check_ok(&format!("SUBSCRIBE {}", quote!(mailbox.as_ref())))
     }
 
@@ -770,7 +770,7 @@ impl<T: Read + Write> Session<T> {
     /// specified mailbox name from the server's set of "active" or "subscribed" mailboxes as
     /// returned by [`Session::lsub`].  This command returns `Ok` only if the unsubscription is
     /// successful.
-    pub fn unsubscribe<S: AsRef<str>>(&mut self, mailbox: S) -> Result<()> {
+    pub fn unsubscribe(&mut self, mailbox: impl AsRef<str>) -> Result<()> {
         self.run_command_and_check_ok(&format!("UNSUBSCRIBE {}", quote!(mailbox.as_ref())))
     }
 
@@ -813,7 +813,7 @@ impl<T: Read + Write> Session<T> {
     ///
     /// Alternatively, the client may fall back to using just [`Session::expunge`], risking the
     /// unintended removal of some messages.
-    pub fn uid_expunge<S: AsRef<str>>(&mut self, uid_set: S) -> Result<Deleted> {
+    pub fn uid_expunge(&mut self, uid_set: impl AsRef<str>) -> Result<Deleted> {
         self.run_command(&format!("UID EXPUNGE {}", uid_set.as_ref()))?;
         self.read_response()
             .and_then(|(lines, _)| parse_expunge(lines, &mut self.unsolicited_responses_tx))
@@ -896,11 +896,11 @@ impl<T: Read + Write> Session<T> {
     ///     Ok(())
     /// }
     /// ```
-    pub fn store<S1, S2>(&mut self, sequence_set: S1, query: S2) -> Result<Fetches>
-    where
-        S1: AsRef<str>,
-        S2: AsRef<str>,
-    {
+    pub fn store(
+        &mut self,
+        sequence_set: impl AsRef<str>,
+        query: impl AsRef<str>,
+    ) -> Result<Fetches> {
         self.run_command_and_read_response(&format!(
             "STORE {} {}",
             sequence_set.as_ref(),
@@ -911,11 +911,11 @@ impl<T: Read + Write> Session<T> {
 
     /// Equivalent to [`Session::store`], except that all identifiers in `sequence_set` are
     /// [`Uid`]s. See also the [`UID` command](https://tools.ietf.org/html/rfc3501#section-6.4.8).
-    pub fn uid_store<S1, S2>(&mut self, uid_set: S1, query: S2) -> Result<Fetches>
-    where
-        S1: AsRef<str>,
-        S2: AsRef<str>,
-    {
+    pub fn uid_store(
+        &mut self,
+        uid_set: impl AsRef<str>,
+        query: impl AsRef<str>,
+    ) -> Result<Fetches> {
         self.run_command_and_read_response(&format!(
             "UID STORE {} {}",
             uid_set.as_ref(),
@@ -931,10 +931,10 @@ impl<T: Read + Write> Session<T> {
     ///
     /// If the `COPY` command is unsuccessful for any reason, the server restores the destination
     /// mailbox to its state before the `COPY` attempt.
-    pub fn copy<S1: AsRef<str>, S2: AsRef<str>>(
+    pub fn copy(
         &mut self,
-        sequence_set: S1,
-        mailbox_name: S2,
+        sequence_set: impl AsRef<str>,
+        mailbox_name: impl AsRef<str>,
     ) -> Result<()> {
         self.run_command_and_check_ok(&format!(
             "COPY {} {}",
@@ -945,10 +945,10 @@ impl<T: Read + Write> Session<T> {
 
     /// Equivalent to [`Session::copy`], except that all identifiers in `sequence_set` are
     /// [`Uid`]s. See also the [`UID` command](https://tools.ietf.org/html/rfc3501#section-6.4.8).
-    pub fn uid_copy<S1: AsRef<str>, S2: AsRef<str>>(
+    pub fn uid_copy(
         &mut self,
-        uid_set: S1,
-        mailbox_name: S2,
+        uid_set: impl AsRef<str>,
+        mailbox_name: impl AsRef<str>,
     ) -> Result<()> {
         self.run_command_and_check_ok(&format!(
             "UID COPY {} {}",
@@ -987,10 +987,10 @@ impl<T: Read + Write> Session<T> {
     /// orphaned).  The server will generally not leave any message in both mailboxes (it would be
     /// bad for a partial failure to result in a bunch of duplicate messages).  This is true even
     /// if the server returns with [`Error::No`].
-    pub fn mv<S1: AsRef<str>, S2: AsRef<str>>(
+    pub fn mv(
         &mut self,
-        sequence_set: S1,
-        mailbox_name: S2,
+        sequence_set: impl AsRef<str>,
+        mailbox_name: impl AsRef<str>,
     ) -> Result<()> {
         self.run_command_and_check_ok(&format!(
             "MOVE {} {}",
@@ -1003,10 +1003,10 @@ impl<T: Read + Write> Session<T> {
     /// [`Uid`]s. See also the [`UID` command](https://tools.ietf.org/html/rfc3501#section-6.4.8)
     /// and the [semantics of `MOVE` and `UID
     /// MOVE`](https://tools.ietf.org/html/rfc6851#section-3.3).
-    pub fn uid_mv<S1: AsRef<str>, S2: AsRef<str>>(
+    pub fn uid_mv(
         &mut self,
-        uid_set: S1,
-        mailbox_name: S2,
+        uid_set: impl AsRef<str>,
+        mailbox_name: impl AsRef<str>,
     ) -> Result<()> {
         self.run_command_and_check_ok(&format!(
             "UID MOVE {} {}",
@@ -1121,10 +1121,10 @@ impl<T: Read + Write> Session<T> {
     ///  - `UNSEEN`: The number of messages which do not have [`Flag::Seen`] set.
     ///
     /// `data_items` is a space-separated list enclosed in parentheses.
-    pub fn status<S1: AsRef<str>, S2: AsRef<str>>(
+    pub fn status(
         &mut self,
-        mailbox_name: S1,
-        data_items: S2,
+        mailbox_name: impl AsRef<str>,
+        data_items: impl AsRef<str>,
     ) -> Result<Mailbox> {
         let mailbox_name = mailbox_name.as_ref();
         self.run_command_and_read_response(&format!(
@@ -1233,7 +1233,7 @@ impl<T: Read + Write> Session<T> {
     ///
     ///  - `BEFORE <date>`: Messages whose internal date (disregarding time and timezone) is earlier than the specified date.
     ///  - `SINCE <date>`: Messages whose internal date (disregarding time and timezone) is within or later than the specified date.
-    pub fn search<S: AsRef<str>>(&mut self, query: S) -> Result<HashSet<Seq>> {
+    pub fn search(&mut self, query: impl AsRef<str>) -> Result<HashSet<Seq>> {
         self.run_command_and_read_response(&format!("SEARCH {}", query.as_ref()))
             .and_then(|lines| parse_id_set(&lines, &mut self.unsolicited_responses_tx))
     }
@@ -1241,7 +1241,7 @@ impl<T: Read + Write> Session<T> {
     /// Equivalent to [`Session::search`], except that the returned identifiers
     /// are [`Uid`] instead of [`Seq`]. See also the [`UID`
     /// command](https://tools.ietf.org/html/rfc3501#section-6.4.8).
-    pub fn uid_search<S: AsRef<str>>(&mut self, query: S) -> Result<HashSet<Uid>> {
+    pub fn uid_search(&mut self, query: impl AsRef<str>) -> Result<HashSet<Uid>> {
         self.run_command_and_read_response(&format!("UID SEARCH {}", query.as_ref()))
             .and_then(|lines| parse_id_set(&lines, &mut self.unsolicited_responses_tx))
     }
@@ -1251,11 +1251,11 @@ impl<T: Read + Write> Session<T> {
     ///
     /// This command is like [`Session::search`], except that
     /// the results are also sorted according to the supplied criteria (subject to the given charset).
-    pub fn sort<S: AsRef<str>>(
+    pub fn sort(
         &mut self,
         criteria: &[extensions::sort::SortCriterion<'_>],
         charset: extensions::sort::SortCharset<'_>,
-        query: S,
+        query: impl AsRef<str>,
     ) -> Result<Vec<Seq>> {
         self.run_command_and_read_response(&format!(
             "SORT {} {} {}",
@@ -1269,11 +1269,11 @@ impl<T: Read + Write> Session<T> {
     /// Equivalent to [`Session::sort`], except that it returns [`Uid`]s.
     ///
     /// See also [`Session::uid_search`].
-    pub fn uid_sort<S: AsRef<str>>(
+    pub fn uid_sort(
         &mut self,
         criteria: &[extensions::sort::SortCriterion<'_>],
         charset: extensions::sort::SortCharset<'_>,
-        query: S,
+        query: impl AsRef<str>,
     ) -> Result<Vec<Uid>> {
         self.run_command_and_read_response(&format!(
             "UID SORT {} {} {}",
@@ -1286,12 +1286,12 @@ impl<T: Read + Write> Session<T> {
 
     // these are only here because they are public interface, the rest is in `Connection`
     /// Runs a command and checks if it returns OK.
-    pub fn run_command_and_check_ok<S: AsRef<str>>(&mut self, command: S) -> Result<()> {
+    pub fn run_command_and_check_ok(&mut self, command: impl AsRef<str>) -> Result<()> {
         self.run_command_and_read_response(command).map(|_| ())
     }
 
     /// Runs any command passed to it.
-    pub fn run_command<S: AsRef<str>>(&mut self, untagged_command: S) -> Result<()> {
+    pub fn run_command(&mut self, untagged_command: impl AsRef<str>) -> Result<()> {
         self.conn.run_command(untagged_command.as_ref())
     }
 
@@ -1303,7 +1303,7 @@ impl<T: Read + Write> Session<T> {
     /// additional untagged `RECENT`, `EXISTS`, `FETCH`, and `EXPUNGE` responses!
     ///
     /// The response includes the final [`Response::Done`], which starts at the returned index.
-    pub fn run<S: AsRef<str>>(&mut self, untagged_command: S) -> Result<(Vec<u8>, usize)> {
+    pub fn run(&mut self, untagged_command: impl AsRef<str>) -> Result<(Vec<u8>, usize)> {
         self.conn.run(untagged_command.as_ref())
     }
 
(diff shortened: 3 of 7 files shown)


src/client.rs
```diff
@@ -1280,6 +1280,91 @@ impl<T: Read + Write> Session<T> {
         .and_then(|lines| parse_id_seq(&lines, &mut self.unsolicited_responses_tx))
     }
 
+    /// The [`SETACL` command](https://datatracker.ietf.org/doc/html/rfc4314#section-3.1)
+    /// Modifies the ACLs on the given mailbox for the specified identifier.
+    /// Return [`Error::No`] if the logged in user does not have `a` rights on the mailbox.
```
jonhoo, line 1285: We should at the very least document that these methods aren't supported by all servers, and how users might check for server support. Alternatively, we could go one step further and do:

```rust
#[doc(alias = "RFC 4313")]
pub fn extension_acls(&mut self) -> Result<extensions::Acl<'_>> {
    if self.capabilities()?.has(Capability::TODO) {
        Ok(extensions::Acl(self))
    } else {
        Err("server does not support RFC 4314 ACL extensions")
    }
}
```

where all these methods are provided on the `extensions::Acl` type. That would also be a nice way to encapsulate other potential extensions in the future. What do you think?

jonhoo: In fact, with GATs we could go even further and have

```rust
pub trait Extension {
    type Extended<'session>;
    fn extend<'session>(self, s: &'session mut Session) -> Result<Self::Extended<'session>>;
}

pub struct Rfc4314;

impl Extension for Rfc4314 {
    type Extended<'session> = extensions::Acl<'session>;
    fn extend<'session>(self, s: &'session mut Session) -> Result<Self::Extended<'session>> {
        if session.capabilities()?.has(Capability::TODO) {
            Ok(extensions::Acl(session))
        } else {
            Err("server does not support RFC 4314 ACL extensions")
        }
    }
}

impl Session {
    pub fn extension<E: Extension>(&mut self, ext: E) -> Result<E::Extended<'_>> {
        ext.extend(self)
    }
}
```

but given that GATs aren't stable I mostly just wanted to write that because it was interesting :)

src/types/acls.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+impl AclRightList {
+    /// Returns if the AclRightList has the provided ACL (either as a char or an AclRight enum)
+    pub fn has_right<T: Into<AclRight>>(&self, right: T) -> bool {
+        self.data.contains(&right.into())
+    }
+}
+
+impl ToString for AclRightList {
```
jonhoo, line 34: We should implement `Display` instead.

src/types/acls.rs
```diff
@@ -0,0 +1,287 @@
...
+            data: vec.into_iter().collect(),
+        }
+    }
+}
+
+impl From<&str> for AclRightList {
+    fn from(i: &str) -> Self {
+        i.chars()
+            .into_iter()
+            .map(|c| c.into())
```
jonhoo, line 62: This seems dubious. Is _every_ `char` a valid `AclRight`? What happens if one of the characters is some super obscure Unicode code point?

src/types/acls.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+}
+
+/// From [section 3.6 of RFC 4313](https://datatracker.ietf.org/doc/html/rfc4314#section-3.6).
+///
+/// The ACL response from the getacl IMAP command
+#[derive(Debug, Eq, PartialEq)]
+pub struct Acl {
+    /// the mailbox these rights list are for
+    pub mailbox: String,
```
jonhoo, line 74: This should almost certainly be `Cow<'a, str>` so that we can do zero-copy parsing and not require copying to use this in a request.

src/types/acls.rs
```diff
@@ -0,0 +1,287 @@
...
+                    return Err(Error::Parse(ParseError::Invalid(lines.to_vec())));
+                }
+            }
+        }
+
+        acl.ok_or_else(|| Error::Parse(ParseError::Invalid(lines.to_vec())))
+    }
+}
+
+impl Into<Acl> for imap_proto::types::Acl<'_> {
```
jonhoo, line 110: We should avoid this `impl` and make it an inherent method instead since it ends up being another public part of the API tied to `imap_proto`.

src/types/acls.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+}
+
+/// From [section 3.6 of RFC 4313](https://datatracker.ietf.org/doc/html/rfc4314#section-3.6).
+///
+/// The list of identifiers and rights for the ACL response
+#[derive(Debug, Eq, PartialEq)]
+pub struct AclEntry {
+    /// The user identifier the rights are for
+    pub identifier: String,
```
jonhoo, line 125: `Cow<'a, str>`

src/types/acls.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// the rights for the provided identifier
+    pub rights: AclRightList,
+}
+
+impl Into<AclEntry> for imap_proto::types::AclEntry<'_> {
```
jonhoo, line 130: Same here: inherent method.

src/types/acls.rs
jonhoo: The same applies to the various later structs.

jonhoo (changes requested):
Finally got around to taking a look at this. Thanks for doing all the groundwork! I left some notes inline.

Cargo.toml
```diff
@@ -17,6 +17,9 @@ categories = ["email", "network-programming"]
 rustls-tls = ["rustls-connector"]
 default = ["native-tls"]
 
+[patch.crates-io]
+imap-proto = { git = "https://github.com/djc/tokio-imap" }
```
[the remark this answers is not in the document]
urkle: I pinged the developer, asking about a release.

src/client.rs
```diff
@@ -1280,6 +1280,91 @@ impl<T: Read + Write> Session<T> {
         .and_then(|lines| parse_id_seq(&lines, &mut self.unsolicited_responses_tx))
     }
 
+    /// The [`SETACL` command](https://datatracker.ietf.org/doc/html/rfc4314#section-3.1)
```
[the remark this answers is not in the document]
urkle: No problem..  It's your project, so nitpick away.

Cargo.toml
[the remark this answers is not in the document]
urkle: They released a new version. so updated the cargo w/ the new dependency (0.16.1)

src/types/acls.rs
urkle: the spec only allows lower alpha characters to be defined as extra rights. And the way I built the AclRight enum is that the core-defined rights (in 4314 and the annotate right in 5257) are enumerated values in the enum.  And any other "char" is put under a "Custom(char)" enumeration.  In this way any custom right provided by the server (via RIGHTS= capability) will be handled correctly by this iteration without issue. (I can add another test to show that case)

src/types/acls.rs
urkle: @jonhoo is something like this what you are meaning here? d0a0f0d530

src/types/acls.rs
urkle: I've added a new commit that does this..  It makes things more complex IMHO.
See the changes here: d901b36273cff675a566c6b36e10070e56280112

src/client.rs
urkle: @jonhoo please look through the latest code changes and the two fixup commits that adjust the structure of things (the Cow'ifying).   Once we settle on those approaches then I'll work on this refactoring.

src/client.rs
```diff
@@ -2020,6 +2104,146 @@ mod tests {
         assert_eq!(ids, [1, 2, 3, 4, 5].iter().cloned().collect::<Vec<_>>());
     }
 
+    #[test]
```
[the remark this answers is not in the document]
urkle: I can implement some integration tests,  The issue I see right now is I can build tests but only can test them on a real server (cyrus imap in my case). as Greenmail does not currently support the ACL extension. (they have comments in their code for where it would be added).   I've added a commit ( 85f8d0c6932cac714843fc5cce1b7da4df78b18a ) to pull out the host/port methods so that I can override ALL via ENV to allow me to more easily test that.  ( I can also provide a Dockerfile to setup a cyrus imap setup that will work like greenmail once I get the SMTP part done)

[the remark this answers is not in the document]
jonhoo: Ahh, that's a shame. If you have a test that works locally against a real server, then I'm okay merging with that test not being run against Greenmail :+1: 

Having integration test setup against Cyrus as well as Greenmail would be amazing! The fact that we currently only test against a single server implementation is certainly a shortcoming.

src/types/acls.rs
jonhoo: Yep, exactly!

src/types/acls.rs
jonhoo: Right, but this doesn't check for "lower alpha characters". It takes _any_ Unicode symbol (`char`). I think we'll want to have this fail if the `&str` contains any non-lower-ASCII-alpha characters.

src/types/acls.rs
jonhoo: It does, but it also adds a feature :)

src/types/acls.rs
```diff
@@ -0,0 +1,351 @@
...
+
+/// From [section 3.6 of RFC 4313](https://datatracker.ietf.org/doc/html/rfc4314#section-3.6).
+///
+/// The ACL response from the getacl IMAP command
+#[self_referencing]
+pub struct Acl {
+    data: Vec<u8>,
+    #[borrows(data)]
+    #[covariant]
+    pub(crate) acl: InnerAcl<'this>,
```
jonhoo, line 77: I think this would become nicer if you kept the mailbox in this struct. So:
```suggestion
    #[borrows(data)]
    #[covariant]
    pub(crate) mailbox: Cow<'this, str>,
    /// The list of identifier/rights pairs for the mailbox
    #[borrows(data)]
    #[covariant]
    pub(crate) acls: Vec<AclEntry<'a>>,
```

It'll also let you get rid of the `InnerAcl` type.

src/types/acls.rs
```diff
@@ -0,0 +1,351 @@
...
+
+/// From [section 3.7 of RFC 4313](https://datatracker.ietf.org/doc/html/rfc4314#section-3.7).
+///
+/// The LISTRIGHTS response from the listrights IMAP command
+#[self_referencing]
+pub struct ListRights {
+    data: Vec<u8>,
+    #[borrows(data)]
+    #[covariant]
+    pub(crate) rights: InnerListRights<'this>,
```
jonhoo, line 161: I don't think you need this inner type here either — you can just declare the fields directly here.

src/types/acls.rs
```diff
@@ -0,0 +1,351 @@
...
+
+/// From [section 3.8 of RFC 4313](https://datatracker.ietf.org/doc/html/rfc4314#section-3.8).
+///
+/// The MYRIGHTS response from the myrights IMAP command
+#[self_referencing]
+pub struct MyRights {
+    data: Vec<u8>,
+    #[borrows(data)]
+    #[covariant]
+    pub(crate) rights: InnerMyRights<'this>,
```
jonhoo, line 245: And same here.

jonhoo (changes requested):
Would you mind not force-pushing going forward — it makes reviewing incremental changes a pain (mostly because GitHub handles force-pushes so poorly) :sweat_smile:

src/types/acls.rs
urkle: ok.. I'll change it to a TryFrom.

src/types/acls.rs
urkle: I started that way, however the way ouroboros's build works it has separate parsers for each member. (mailbox_builder, and acls_builder both receicing the [u8] input). so I couldn't find any easy way to NOT end up parsing twice (which definitely would be undesirable). 

What I did instead was replace the named inner structs with inline anonymous tuples.

src/types/acls.rs
```diff
@@ -56,16 +56,32 @@ impl From<Vec<AclRight>> for AclRightList {
...
             .map(|c| c.into())
             .collect::<HashSet<AclRight>>()
-            .into()
+            .into())
     }
 }
 
+/// Error from parsing AclRight strings
+#[derive(Debug, Eq, PartialEq)]
+pub enum AclRightError {
```
jonhoo, line 80: We'll definitely want to implement `std::error::Error` for this. I'm also not sure it needs to be an `enum`? If we truly expect new variants might be added in the future though, this should be marked `#[non_exhaustive]`.

src/types/acls.rs
jonhoo: Ahh, that's awkward. I think I'm in favor of reverting 00d3b288f2476b0dd87942815bbfb794cef43c00 then — tuple types with 4+ members get unwieldy and error-prone real fast!

urkle: I reverted that commit.

src/types/acls.rs
urkle: I added non_exhausted and implemented Error.  And by using an enum it makes it easier to extend in the future w/o breaking API.

src/client.rs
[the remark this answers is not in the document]
urkle: @jonhoo are you OK with skipping the integration tests in this PR and focusing on building those in #228 ?

src/client.rs
```diff
@@ -1280,6 +1280,89 @@ impl<T: Read + Write> Session<T> {
...
     }
 
+    /// The [`SETACL` command](https://datatracker.ietf.org/doc/html/rfc4314#section-3.1)
+    ///
+    /// Modifies the ACLs on the given mailbox for the specified identifier.
+    /// Return [`Error::No`] if the logged in user does not have `a` rights on the mailbox.
+    pub fn set_acl<S1: AsRef<str>, S2: AsRef<str>>(
+        &mut self,
+        mailbox_name: S1,
+        identifier: S2,
```
jonhoo, line 1290: nit: I'd probably just use `impl AsRef<str>` for each of these (and in the other methods).

src/client.rs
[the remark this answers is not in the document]
jonhoo: I think maybe we should try to land #228 so that we can then add the tests directly in this PR. What do you think?

src/types/acls.rs
```diff
@@ -0,0 +1,398 @@
...
+use crate::types::UnsolicitedResponse;
+use imap_proto::types::AclRight;
+use imap_proto::Response;
+use ouroboros::self_referencing;
+use std::borrow::Cow;
+use std::collections::HashSet;
+use std::fmt::{write, Display, Formatter};
+use std::sync::mpsc;
+
+/// enum used for set_acl to specify how the ACL is to be modified.
```
jonhoo, line 12: We should use doc-links here. So `[Client::set_acl]` for example.

src/types/acls.rs
```diff
@@ -0,0 +1,398 @@
...
+    Replace,
+    /// Add the given ACLs to the identifier for the mailbox
+    Add,
+    /// Remove the given ACLs from the identifier for the mailbox
+    Remove,
+}
+
+/// Helpful wrapper around the ACL rights vector
+#[derive(Debug, Eq, PartialEq)]
+pub struct AclRightList {
```
jonhoo, line 24: Would this be better as just `AclRights`? Feels weird to call it a list when it's a set.

src/types/acls.rs
```diff
@@ -0,0 +1,398 @@
...
+    /// Returns a list of identifier/rights pairs for the mailbox
+    pub fn acls(&self) -> &[AclEntry<'_>] {
+        &*self.borrow_acl().acls
+    }
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct InnerAcl<'a> {
+    /// The mailbox the ACL Entries belong to
+    pub(crate) mailbox: Cow<'a, str>,
```
jonhoo, line 169: I'm a little confused about why this and other fields are `Cow` and not just `&'a str` when it doesn't appear they're ever `Cow::Owned`, and will never be settable directly by the user?

src/client.rs
urkle: Now, FYI it this does follow the pattern you use for `fetch`,  `uid_fetch`, and `rename`.  but I can do that for these.

src/types/acls.rs
urkle: If I set them as `&'a str`, then the ouroboros builders fail to compile as the Cow is how they become referenced against the internal data struct. So there technically is an "owned" but it's just a transfer from the imap-proto type to my type (see line 127 in `Acl::parse`)

src/types/acls.rs
```diff
@@ -0,0 +1,399 @@
...
+            data: owned,
+            acl_builder: |input| {
+                let mut lines: &[u8] = input;
+
+                // There should only be ONE single ACL response
+                while !lines.is_empty() {
+                    match imap_proto::parser::parse_response(lines) {
+                        Ok((_rest, Response::Acl(a))) => {
+                            // lines = rest;
+                            return Ok(InnerAcl {
```
urkle, line 126: @jonhoo so, as I was implementing #235 I'm thinking this implementation for these methods is probably bad.  As it immediately returns out once it finds the Acl, which means it won't necessarily process all of the unsolicited responses.

src/client.rs
jonhoo: Let's actually go ahead and change those too to be `impl AsRef<str>`. We should do so now since that's a breaking change.

src/types/acls.rs
jonhoo: Huh, that's very odd. I wonder if we could just match on the `Cow` to specifically grab the `'a` reference out of the `Cow::Borrowed` if we know it's always that variant. Or, phrased differently, I don't quite see how it'd ever end up being a `Cow::Owned` in that code, since the parser should already only be yielding things that borrow from `input` (I think). Unless it _sometimes_ does decoding I suppose, which could be the case.

src/types/acls.rs
jonhoo: Oh, yes, good catch! It does indeed need to keep looping until it sees a `Done` or hits the end of the string I think, to catch cases where an unsolicited message appears between the `Acl` response and the `Done`.

src/client.rs
urkle: @jonhoo I added a new commit that changes all methods to use impl.

src/types/acls.rs
urkle: I have reworked the code now.

urkle:
@jonhoo let me know if things are good now w/ this PR..  and I'll rebase (resolving conflicts) and squash my commits down.

src/client.rs
```diff
@@ -430,10 +430,10 @@ impl<T: Read + Write> Client<T> {
     ///     };
     /// }
     /// ```
-    pub fn authenticate<A: Authenticator, S: AsRef<str>>(
+    pub fn authenticate(
         mut self,
-        auth_type: S,
-        authenticator: &A,
+        auth_type: impl AsRef<str>,
+        authenticator: &impl Authenticator,
```
jonhoo, line 436: Ah, so, `Authenticator` I _think_ we may want to stay as `<A>`, since it's feasible that someone may want to specify the authenticator using turbofish syntax (which doesn't work with `:impl`). I don't know of a concrete case, I just have a feeling.

src/client.rs
jonhoo: I still think it'd be nice to do something like this, not just for this but also for the QRESYNC stuff that @bitfehler has been landing recently. We can do that in a separate PR though.

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+use crate::types::UnsolicitedResponse;
+use imap_proto::types::AclRight;
+use imap_proto::Response;
+use ouroboros::self_referencing;
+use std::borrow::Cow;
+use std::collections::HashSet;
+use std::fmt::{Display, Formatter};
+use std::sync::mpsc;
+
+/// enum used for [Session::set_acl] to specify how the ACL is to be modified.
```
jonhoo, line 12:
```suggestion
/// Specifies how [`Session::set_acl`] should modify an existing permission set.
```

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// Replace all ACLs on the identifier for the mailbox
+    Replace,
+    /// Add the given ACLs to the identifier for the mailbox
+    Add,
+    /// Remove the given ACLs from the identifier for the mailbox
+    Remove,
+}
+
+/// Helpful wrapper around the ACL rights vector
+/// Used as input for [Session::set_acl] as output in [ListRights], [MyRights], and [AclEntry]
```
jonhoo, line 23:
```suggestion
/// A set of [`AclRight`]s.
///
/// Used as input for [`Session::set_acl`] and as output in [`ListRights`], [`MyRights`], and [`AclEntry`].
```

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+#[derive(Debug, Eq, PartialEq)]
+pub struct AclRights {
+    pub(crate) data: HashSet<AclRight>,
+}
+
+impl AclRights {
+    /// Returns true if the AclRights has the provided ACL (either as a char or an AclRight enum)
+    pub fn has_right<T: Into<AclRight>>(&self, right: T) -> bool {
```
jonhoo, line 31: I wonder if we should just call this `has`?

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+use imap_proto::types::AclRight;
+use imap_proto::Response;
+use ouroboros::self_referencing;
+use std::borrow::Cow;
+use std::collections::HashSet;
+use std::fmt::{Display, Formatter};
+use std::sync::mpsc;
+
+/// enum used for [Session::set_acl] to specify how the ACL is to be modified.
+pub enum AclModifyMode {
```
jonhoo, line 13: This feels like it should be `Debug`, `Clone`, and `Copy` at the very least.

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+    pub fn parse(
+        owned: Vec<u8>,
+        unsolicited: &mut mpsc::Sender<UnsolicitedResponse>,
+    ) -> Result<Self, Error> {
+        AclTryBuilder {
+            data: owned,
+            acl_builder: |input| {
+                let mut acls = Vec::new();
+
+                // There should only be ONE single ACL response
```
jonhoo, line 121: Let's also document why we still use `parse_many_input`.

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+                            })
+                            .collect(),
+                    })),
+                    resp => Ok(MapOrNot::Not(resp)),
+                })?;
+
+                if acls.is_empty() {
+                    Err(Error::Parse(ParseError::Invalid(input.to_vec())))
+                } else {
+                    Ok(acls.remove(0))
```
jonhoo, line 140: Should we error if there are >1?

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+    Replace,
+    /// Add the given ACLs to the identifier for the mailbox
+    Add,
+    /// Remove the given ACLs from the identifier for the mailbox
+    Remove,
+}
+
+/// Helpful wrapper around the ACL rights vector
+/// Used as input for [Session::set_acl] as output in [ListRights], [MyRights], and [AclEntry]
+#[derive(Debug, Eq, PartialEq)]
```
jonhoo, line 24: Any particular reason why this isn't also `Clone`?

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// The mailbox the ACL Entries belong to
+    pub(crate) mailbox: Cow<'a, str>,
+    /// The list of identifier/rights pairs for the mailbox
+    pub(crate) acls: Vec<AclEntry<'a>>,
+}
+
+/// From [section 3.6 of RFC 4313](https://datatracker.ietf.org/doc/html/rfc4314#section-3.6).
+///
+/// The list of identifiers and rights for the [Acl] response
+#[derive(Debug, Eq, PartialEq)]
```
jonhoo, line 170: This also feels like it should be `Clone`.

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+                        required: a.required.into(),
+                        optional: a.optional.into(),
+                    })),
+                    resp => Ok(MapOrNot::Not(resp)),
+                })?;
+
+                if rights.is_empty() {
+                    Err(Error::Parse(ParseError::Invalid(input.to_vec())))
+                } else {
+                    Ok(rights.remove(0))
```
jonhoo, line 214: Same here — should we error if there are >1?

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+                        mailbox: a.mailbox,
+                        rights: a.rights.into(),
+                    })),
+                    resp => Ok(MapOrNot::Not(resp)),
+                })?;
+
+                if rights.is_empty() {
+                    Err(Error::Parse(ParseError::Invalid(input.to_vec())))
+                } else {
+                    Ok(rights.remove(0))
```
jonhoo, line 289: Same here — should we error if there are >1?

src/types/acls.rs
```diff
@@ -0,0 +1,372 @@
...
+    pub fn parse(
+        owned: Vec<u8>,
+        unsolicited: &mut mpsc::Sender<UnsolicitedResponse>,
+    ) -> Result<Self, Error> {
+        MyRightsTryBuilder {
+            data: owned,
+            rights_builder: |input| {
+                let mut rights = Vec::new();
+
+                // There should only be ONE single MYRIGHTS response
```
jonhoo, line 277: It does feel very weird to be using `parse_many_into` for this in so many places. I wonder if we should make a new helper, `parse_until_done`, that specifically handles the single-item case (and could internally call `parse_many_into` for now).

jonhoo (changes requested):
I did a pass over your new changes and a found a few more tweaks we should make.

src/types/acls.rs
urkle: I've reworked this like I did in quota so it'll Err on != 1 (and adding tests)

src/types/acls.rs
urkle: it wasn't needed by any of usage/tests :)  I'll add though.

src/types/acls.rs
urkle: I can look at that.

src/types/acls.rs
jonhoo: I think this is more akin to `HashSet` than `HashMap`. `HashMap::contains_key` has to specifically say "key" since there are also values, but `HashSet` just has `HashSet::contains`, so let's use that!

src/parse.rs
```diff
@@ -71,6 +71,31 @@ where
     }
 }
 
+/// Parse an expected single `T` Response with `F` and store it in `into`.
+/// Responses other than `T` go into the `unsolicited` channel.
+///
+/// If zero or more than one `T` is found then into will be `None`
+pub(crate) fn parse_until_done<'input, T, F>(
+    input: &'input [u8],
+    into: &mut Option<T>,
```
jonhoo, line 80: Instead of taking `&mut Option<T>`, let's just return a `T`

src/parse.rs
```diff
@@ -71,6 +71,31 @@ where
...
+    unsolicited: &mut mpsc::Sender<UnsolicitedResponse>,
+    map: F,
+) -> Result<()>
+where
+    F: FnMut(Response<'input>) -> Result<MapOrNot<'input, T>>,
+{
+    let mut temp_output = Vec::<T>::new();
```
jonhoo, line 87: I think we _should_ be able to to have `parse_many_into` accept an `&mut Option` as `IT` directly. At least I thought `Option` implemented `Extend` as well.

src/parse.rs
```diff
@@ -71,6 +71,31 @@ where
...
+
+    parse_many_into(input, &mut temp_output, unsolicited, map)?;
+
+    *into = match temp_output.len() {
+        1 => Some(temp_output.remove(0)),
+        _ => None,
```
jonhoo, line 93: Any particular reason not to turn this into an error right here?

urkle: So it leaves the decision up to the caller of whether it is an error (e.g. for cases where 0 or 1 response are acceptable).

urkle:
@jonhoo now that we have cyrus imapd testing in mainline I'll cherry pick those over (so as not to rebase yet) and add some integration tests in the next few days.

src/parse.rs
jonhoo: Huh, interesting. I could have sworn... Oh well, fair enough, let's do it this way then.

src/parse.rs
jonhoo: But they can't tell from this whether it's 0 or >1. I think it's okay to have it return an error on != 1, and then the caller can always recover the `Vec` from the error if they really want to.

urkle:
@jonhoo I have pushed up some acl integration tests using a cargo feature to enable.

src/client.rs
jonhoo: Let's document that this (and the other ACL methods) only work for servers with the ACL-related capability listed though :+1:

src/types/acls.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub(crate) mailbox: Cow<'a, str>,
+    /// The list of identifier/rights pairs for the mailbox
+    pub(crate) acls: Vec<AclEntry<'a>>,
+}
+
+/// From [section 3.6 of RFC 4313](https://datatracker.ietf.org/doc/html/rfc4314#section-3.6).
+///
+/// The list of identifiers and rights for the [`Acl`] response
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub struct AclEntry<'a> {
```
jonhoo, line 165: Let's set this to be `#[non_exhaustive]` in case other extensions we add support for in the future might add more fields.

src/types/acls.rs
```diff
@@ -0,0 +1,350 @@
...
+                    })),
+                    resp => Ok(MapOrNot::Not(resp)),
+                })
+            },
+        }
+        .try_build()
+    }
+
+    /// Returns the mailbox for the rights
+    pub fn mailbox(&self) -> &str {
```
jonhoo, line 208: Given the change we made around `Quota`, how do you feel about having just one accessor here that gives out `&InnerListRights` (and giving it a friendlier name)?

src/types/acls.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    /// Returns the set of rights that can be granted to the identifier
+    pub fn optional(&self) -> &AclRights {
+        &self.borrow_rights().optional
+    }
+}
+
+/// Inner struct to manage storing the references for ouroboros
+#[derive(Debug, Eq, PartialEq)]
+pub(crate) struct InnerListRights<'a> {
```
jonhoo, line 230: If we do make this `pub`, it should also be `#[non_exhaustive]`.

.github/workflows/test.yml
```diff
@@ -32,3 +32,28 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.toolchain }}
+      - uses: actions/checkout@v2
+      - name: cargo test
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --features test-full-imap -- --ignored
```
jonhoo, line 51: With the new `--features` I don't think you also need `--ignored` any more?

jonhoo (commented):
I think you'll need to rebase this onto `master` so that GitHub will run CI again.

src/types/acls.rs
urkle: If we went that route then it might make more sense to simply name the structs returned by the Session methods to be *Response (ListRightsResponse, MyRightsResponse) etc.. And then the "inner" structs will be renamed to ListRights, MyRights, etc..

Let me know what you think.

urkle: @jonhoo I'll push a fixup commit that performs this change (*Response structs) so you can see it and then we can decide to keep or go another direction

urkle:
@jonhoo I rebased + squashed.  then I added 2 fixups for the last set of changes you requested.

## 173add1
- remove test  --ignored
- add docs about server ACL capability is required
- non_exhaustive on AclRights
## be2c6a1
- rework Session method repones types to be *Response (MyRightsResponse AclResponse) and make public the now renamed "Inner*"

urkle:
@jonhoo all green! ✅

src/types/acls.rs
jonhoo: That seems reasonable to me!

src/types/acls.rs
```diff
@@ -0,0 +1,387 @@
...
+    Add,
+    /// Remove the given ACLs from the identifier for the mailbox
+    Remove,
+}
+
+/// A set of [`imap_proto::AclRight`]s.
+///
+/// Used as input for [`Session::set_acl`] as output in [`ListRights`], [`MyRights`], and [`AclEntry`]
+#[derive(Debug, Eq, PartialEq, Clone)]
+#[non_exhaustive]
```
jonhoo, line 27: I don't think this needs `#[non_exhaustive]` since it has private fields already anyway.

src/types/acls.rs
jonhoo: This still needs to be done.

src/types/acls.rs
```diff
@@ -0,0 +1,387 @@
...
+    pub fn list_rights(&self) -> &ListRights<'_> {
+        self.borrow_rights()
+    }
+}
+
+/// From [section 3.7 of RFC 4313](https://datatracker.ietf.org/doc/html/rfc4314#section-3.7).
+///
+/// Used by [`ListRightsResponse`].
+#[derive(Debug, Eq, PartialEq)]
+#[non_exhaustive]
```
jonhoo, line 232: This doesn't need to be `#[non_exhaustive]` since it has private fields. It would need `#[non_exhaustive]` if we made all those fields `pub` (and removed the accessor methods) though.

src/types/acls.rs
```diff
@@ -0,0 +1,387 @@
...
+                    })),
+                    resp => Ok(MapOrNot::Not(resp)),
+                })
+            },
+        }
+        .try_build()
+    }
+
+    /// Access to the wrapped [`ListRights`] struct
+    pub fn acl(&self) -> &Acl<'_> {
```
jonhoo, line 145: Instead of having this method be called something different for each type, how about a method by the same name in each one to access the "inner" type. Maybe it could be called `parsed`, `inner`, or `as_ref`?

src/types/acls.rs
```diff
@@ -0,0 +1,387 @@
...
+    pub fn my_rights(&self) -> &MyRights<'_> {
+        self.borrow_rights()
+    }
+}
+
+/// From [section 3.8 of RFC 4313](https://datatracker.ietf.org/doc/html/rfc4314#section-3.8).
+///
+/// Used by [`MyRightsResponse`].
+#[derive(Debug, Eq, PartialEq)]
+#[non_exhaustive]
```
jonhoo, line 311: This also does not need to be `#[non_exhaustive]` if it's only going to have private fields.

src/client.rs
```diff
@@ -443,9 +443,9 @@ impl<T: Read + Write> Client<T> {
     }
 
     /// This func does the handshake process once the authenticate command is made.
-    fn do_auth_handshake<A: Authenticator>(
+    fn do_auth_handshake(
         mut self,
-        authenticator: &A,
+        authenticator: &impl Authenticator,
```
jonhoo, line 448: This one too we should keep as `<A>`.

src/client.rs
jonhoo: Leaving this unresolved to remember that we'll probably still want to make some change to how extensions are handled so that it's harder to misuse them and so that we don't pollute `Client` with tons and tons of methods.

src/client.rs
```diff
@@ -1282,14 +1282,112 @@ impl<T: Read + Write> Session<T> {
         .and_then(|lines| parse_id_seq(&lines, &mut self.unsolicited_responses_tx))
     }
 
+    /// The [`SETACL` command](https://datatracker.ietf.org/doc/html/rfc4314#section-3.1)
+    ///
+    /// Modifies the ACLs on the given mailbox for the specified identifier.
+    /// Return [`Error::No`] if the logged in user does not have `a` rights on the mailbox.
+    ///
+    /// This method only works against a server with the ACL capability. Otherwise [`Error::Bad`]
```
jonhoo, line 1290: Is "ACL" the actual name for the capability one should check for on the server? Or does it have some other name that'll be what has to be checked for?

src/types/acls.rs
```diff
@@ -0,0 +1,387 @@
...
+use crate::types::UnsolicitedResponse;
+use imap_proto::types::AclRight;
+use imap_proto::Response;
+use ouroboros::self_referencing;
+use std::borrow::Cow;
+use std::collections::HashSet;
+use std::fmt::{Display, Formatter};
+use std::sync::mpsc;
+
+/// Specifies how [`Session::set_acl`] should modify an existing permission set.
```
jonhoo, line 12: Clippy seems to think that this link doesn't work. Almost certainly because `Session` isn't in scope here. You can fix that by adding this to the `use` list:

```rust
#[cfg(doc)]
use crate::Session;
```

I think.

src/types/acls.rs
```diff
@@ -0,0 +1,387 @@
...
+
+/// Error from parsing AclRight strings
+#[derive(Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum AclRightError {
+    /// Returned when a non-lower-case alpha numeric is provided in the rights list string.
+    InvalidRight,
+}
+
+impl Display for AclRightError {
```
jonhoo, line 90: Let's add a test for this `impl` if we're indeed going to ship it. I don't _think_ we actually rely on this `impl` any more, in which case it could arguably just be removed.

jonhoo (changes requested):
Nice! Took another pass over the latest changes.

src/types/acls.rs
urkle: I added it to the wrong structs :)

src/client.rs
urkle: ACL is the actual name in the server.n (So `has_str("ACL")`)

There are other "sub-pieces" for determining additional functionality. e.g.
- checking `RIGHTS=kx`
   - which the imap-proto crate does not easily allow you to query as `RIGHTS=` lists extra rights characters supported by the server.  I'll have to do a new PR so support that one. (and make the Capability enum non_exhaustive)
- also the ANNOTATE-EXPERIMENT-1 capability to detect the annotation extension (adds the `n` ACL)

src/types/acls.rs
urkle: Display (and Debug) are required in order to implement Error.

I can definitely add tests

src/types/acls.rs
urkle: So, `inner` or `parsed` maybe as it is the object inside the wrap
`as_ref()` somewhat, but it is changing types which is not has as_ref() behaves on other rust structs with that contract.

Though once we get to the quota PR there is one of the Responses there that has two types it returns (a QuotaRoot and an array of Quotas)

In the end, your decision on this.

urkle:
@jonhoo there is an error in test caused by a new release of ouroboros joshua-maros/ouroboros#66  There is a pending PR to fix it, but has not yet been released.

src/types/acls.rs
jonhoo: I'm 50/50 on `inner` vs `parsed`. `inner` is more "standard", but `parsed` more accurately reflects what it is. I think either will be fine, so I'll leave it to you :)

For a type that has two "inners", I think it's totally fine for the accessors to be named.

jonhoo:
Sorry for the delay in review!

And yeah, I saw https://github.com/joshua-maros/ouroboros/issues/66

src/types/acls.rs
urkle: I pushed a fixup that uses parsed

urkle:
@jonhoo pushed a fixup that renamed the "inner" methods to `parsed`.  Let me know if there is anything else.  If there is nothing let me know and I'll rebase + squash the commits.

jonhoo (approved):
Looks good to me!

jonhoo:
Released in alpha 9 :tada:

PR #228, opened 2022-05-17
update integration tests to work against cyrus imapd

urkle:
Adds cyrus-imapd support for testing.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/228)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ name: cargo test
 jobs:
   test:
     runs-on: ubuntu-latest
-    name: ${{ matrix.toolchain }}
+    name: Run tests with greenmail ${{ matrix.toolchain }}
     strategy:
       matrix:
         toolchain: [stable, beta, nightly]
@@ -32,3 +32,27 @@ jobs:
           - 3995:3995
         env:
           GREENMAIL_OPTS: "-Dgreenmail.setup.test.all -Dgreenmail.hostname=0.0.0.0 -Dgreenmail.auth.disabled -Dgreenmail.verbose"
+  test_cyrus:
+    runs-on: ubuntu-latest
+    name: Run tests with cyrus ${{ matrix.toolchain }}
+    strategy:
+      matrix:
+        toolchain: [stable, beta, nightly]
+    steps:
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.toolchain }}
+      - uses: actions/checkout@v2
+      - name: cargo test
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+    services:
+      cyrus_imapd:
+        image: outoforder/cyrus-imapd-tester:latest
+        ports:
+          - 3025:25
+          - 3143:143
+          - 3465:465
+          - 3993:993
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,7 @@ jobs:
       - name: cargo test
         uses: actions-rs/cargo@v1
         with:
-          command: test
+          command: test -- --ignored
     services:
       cyrus_imapd:
         image: outoforder/cyrus-imapd-tester:latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,8 @@ jobs:
       - name: cargo test
         uses: actions-rs/cargo@v1
         with:
-          command: test -- --ignored
+          command: test
+          args: -- --ignored
     services:
       cyrus_imapd:
         image: outoforder/cyrus-imapd-tester:latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ name: cargo test
 jobs:
   test:
     runs-on: ubuntu-latest
-    name: Run tests with greenmail ${{ matrix.toolchain }}
+    name: greenmail/${{ matrix.toolchain }}
     strategy:
       matrix:
         toolchain: [stable, beta, nightly]
@@ -34,7 +34,7 @@ jobs:
           GREENMAIL_OPTS: "-Dgreenmail.setup.test.all -Dgreenmail.hostname=0.0.0.0 -Dgreenmail.auth.disabled -Dgreenmail.verbose"
   test_cyrus:
     runs-on: ubuntu-latest
-    name: Run tests with cyrus ${{ matrix.toolchain }}
+    name: cyrus/${{ matrix.toolchain }}
     strategy:
       matrix:
         toolchain: [stable, beta, nightly]
--- a/tests/imap_integration.rs
+++ b/tests/imap_integration.rs
@@ -101,6 +101,7 @@ fn smtp(user: &str) -> lettre::SmtpTransport {
 }
 
 #[test]
+#[ignore]
 fn connect_insecure_then_secure() {
     let host = test_host();
     // ignored because of https://github.com/greenmail-mail-test/greenmail/issues/135
@@ -342,7 +343,7 @@ fn list() {
     let mut s = session("readonly-test@localhost");
     s.select("INBOX").unwrap();
     let subdirs = s.list(None, Some("%")).unwrap();
-    assert_eq!(subdirs.len(), 0);
+    assert_eq!(subdirs.len(), 1);
 
     // TODO: make a subdir
 }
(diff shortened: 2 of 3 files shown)


jonhoo:
Thanks! Two initial observations:

1. This seems to _replace_ Greenmail with Cyrus, which I don't think is what we should do. Instead, I'd prefer to test against _both_. I'm fine if we only run a subset of CI against both though. Coverage for example we can just run against one.
2. Is it feasible for us to run Cyrus through `services` the same way we do with Greenmail rather than constructing the Docker image inline with the build steps?

urkle:
> Thanks! Two initial observations:
> 
>     1. This seems to _replace_ Greenmail with Cyrus, which I don't think is what we should do. Instead, I'd prefer to test against _both_. I'm fine if we only run a subset of CI against both though. Coverage for example we can just run against one.

Yeah, I can do that. (and sounds very reasonable)   What I'll do to make things easier for controlling what tests are run is move the ACL tests that I'll write to a separate integration test file entirely. 

>     2. Is it feasible for us to run Cyrus through `services` the same way we do with Greenmail rather than constructing the Docker image inline with the build steps?

In order to do that I'd have to publish the pre-built containers to docker hub.  and once I get all the issues sorted (after writing the ACL tests) then I'll do that and that'll make this tests run much faster as well.

urkle:
@jonhoo I have reworked this PR to now only add a test-run for cyrus-imapd and pushed up an single image to my docker hub that spins up a service somewhat mimicking greenmail.

tests/imap_integration.rs
```diff
@@ -145,12 +196,11 @@ fn inbox() {
         .any(|m| m == &imap::types::UnsolicitedResponse::Recent(2)));
 
     // let's see that we can also fetch the e-mails
-    let fetch = c.fetch("1", "(ALL UID)").unwrap();
+    let fetch = c.fetch("1", "(ENVELOPE INTERNALDATE UID)").unwrap();
```
jonhoo, line 199: Why all these changes away from `ALL`?

.github/workflows/test.yml
```diff
@@ -32,3 +32,27 @@ jobs:
...
+          profile: minimal
+          toolchain: ${{ matrix.toolchain }}
+      - uses: actions/checkout@v2
+      - name: cargo test
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+    services:
+      cyrus_imapd:
+        image: outoforder/cyrus-imapd-tester:latest
```
jonhoo, line 53: One worry I have with this is that version of Cyrus is likely to get out of date over time unless you commit to updating it. Is there not a standard Cyrus Docker image that's regularly maintained we can use?

.github/workflows/test.yml
```diff
@@ -6,7 +6,7 @@ name: cargo test
 jobs:
   test:
     runs-on: ubuntu-latest
-    name: ${{ matrix.toolchain }}
+    name: Run tests with greenmail ${{ matrix.toolchain }}
```
jonhoo, line 9:
```suggestion
    name: greenmail/${{ matrix.toolchain }}
```

.github/workflows/test.yml
```diff
@@ -32,3 +32,27 @@ jobs:
           - 3995:3995
         env:
           GREENMAIL_OPTS: "-Dgreenmail.setup.test.all -Dgreenmail.hostname=0.0.0.0 -Dgreenmail.auth.disabled -Dgreenmail.verbose"
+  test_cyrus:
+    runs-on: ubuntu-latest
+    name: Run tests with cyrus ${{ matrix.toolchain }}
```
jonhoo, line 37:
```suggestion
    name: cyrus/${{ matrix.toolchain }}
```

jonhoo (changes requested):
Oh, yeah, this looks much nicer! Left another few notes.

tests/imap_integration.rs
urkle: Cyrus IMAP doesn't support ALL

.github/workflows/test.yml
urkle: No, there is no official cyrus docker image at all.  (I had looked for that first when setting up testing for the app I'm building).   But this docker image is based on Rocky8 (derived from RHEL8) and thus that base image will be maintained for quite a while.  And I'll setup a github repo here on the next few days with the docker file.   And in reality the supported features of cyrus for IMAP haven't changed much in years.  (e.g. this is version 3.0.7 and my other servers are running 2.4.x and the functionality is the same between them). 

But as I'm maintaining several mail servers (on Cent 6, and moving toward Rocky 8 or Rocky 9) and writing admin software that is utilizing your crate, I'll be maintaining this for a while.

What I'll do next is setup a github repo with the docker image(s) and set it up so it'll pop out a Rocky 8 and Rocky 9 base image with the same setup.

tests/imap_integration.rs
jonhoo: That's super weird — it's in the base IMAP RFC 3501 after all! For `FETCH` it is [defined as equivalent](https://datatracker.ietf.org/doc/html/rfc3501#page-55) to
```
FLAGS INTERNALDATE RFC822.SIZE ENVELOPE
```
so maybe we should use that?

tests/imap_integration.rs
```diff
@@ -19,9 +19,61 @@ fn tls() -> native_tls::TlsConnector {
...
+fn test_smtps_port() -> u16 {
+    std::env::var("TEST_SMTPS_PORT")
+        .unwrap_or("3465".to_string())
+        .parse()
+        .unwrap_or(3465)
+}
+
+fn clean_mailbox(session: &mut imap::Session<native_tls::TlsStream<TcpStream>>) {
+    session.select("INBOX").unwrap();
+    let inbox = session.search("ALL").unwrap();
```
jonhoo, line 54: Hah, so Cyrus supports `ALL` for `SEARCH`, just _not_ for `FETCH`. Very weird.

jonhoo (approved):

jonhoo:
Huh, it looks like `connect_insecure_then_secure` is suddenly failing with Greenmail?

tests/imap_integration.rs
urkle: I saw that as well that it was int he 3501 RFC, but not the earlier 2060 spec. both of which claim to  4rev1.

I changed it to be what was actually be checked in the tests vs everything.

tests/imap_integration.rs
urkle: yes.. very strange.

urkle:
> 

I just saw the failed tests come back.. I'll look into it and see what got broke and fix.

urkle:
@jonhoo ok.. so I had removed the `#[ignore]` on the starttls test as cyrus does support STARTTLS.  I pushed a fixup that re-added the `#[ignore]` and adjusted the cyrus test to run the ignored tests (for now).

I think the easier approach going forward would be to move the integration tests that cannot be run on greenmail to a separate integration test script.  Then we setup the test to run like this in the CI setup
```
cargo test --lib
cargo test --examples
cargo test --doc
cargo test --test imap_integration
cargo test --test extra_integration # for cyrus only.
```

Unfortunately the cargo test runner does not allow you to run all non-integration tests in one go. (for some reason --doc can't be used with other arguments)

tests/imap_integration.rs
```diff
@@ -288,7 +343,7 @@ fn list() {
     let mut s = session("readonly-test@localhost");
     s.select("INBOX").unwrap();
     let subdirs = s.list(None, Some("%")).unwrap();
-    assert_eq!(subdirs.len(), 0);
+    assert_eq!(subdirs.len(), 1);
```
jonhoo, line 346: This feels like a weird change. Why did the length suddenly change?

jonhoo (commented):
Another way to do this is to have test features. We could either have a feature like `test_on_cyrus`, or we could make it more general along the lines of `test_with_all_capabilities`.

tests/imap_integration.rs
jonhoo: Oh, right, I guess the test was ignored before so it was probably wrong.

urkle: :);.  I only changed it because I currently have cyrus running all ignored tests.

urkle:
> Another way to do this is to have test features. We could either have a feature like `test_on_cyrus`, or we could make it more general along the lines of `test_with_all_capabilities`.

Ahh.. using the cargo feature stuff:).  that is an option.   Once we land this we can play w/ that over on the other 2 branches.

jonhoo:
:tada:

Issue #230, opened 2022-07-03
Make imap panic free


qsdgy:
There's a lot of pain in dealing with panic from lib, especially working with background services.

At least I'd like to remove panics came across with tag asserts (#229 #211).

How about turn [`assert_eq!(tag.as_bytes(), match_tag.as_bytes());`](https://github.com/jonhoo/rust-imap/blob/db2911746358741a74f9dda8b2eecbf480506224/src/client.rs#L1396) to a new [imap](https://docs.rs/imap/latest/imap/index.html)::[error](https://docs.rs/imap/latest/imap/error/index.html)::[Error](https://docs.rs/imap/latest/imap/error/enum.Error.html) variant `Fatal(FatalError)` with some fatal error flag like `FatalErrorOccurred: bool`, and check `FatalErrorOccurred` flag before any `run_command()`?

`FatalError` could be like this:

```rust
enum FatalError{
    TagBroken,
    Other,
}
```

PR #231, opened 2022-07-04
Refactor tag assertion to FatalError check & Fix unexpected increase of tag

### LINKED ISSUE #229: Unexpected increase of  imap::client::Connection.tag
imap::client::Connection.tag has been [increased ](https://github.com/jonhoo/rust-imap/blob/db2911746358741a74f9dda8b2eecbf480506224/src/client.rs#L1469)no matter whether writing command to socket is okay.
```rust
fn run_command(&mut self, untagged_command: &str) -> Result<()> {
        let command = self.create_command(untagged_command);
        self.write_line(command.into_bytes().as_slice())
}
```
```rust
fn create_command(&mut self, command: &str) -> String {
        self.tag += 1;
        format!("{}{} {}", TAG_PREFIX, self.tag, command)
}
```

If `self.write_line()` fails, the next imap command call would violate [assertion](https://github.com/jonhoo/rust-imap/blob/db2911746358741a74f9dda8b2eecbf480506224/src/client.rs#L1396), and program panics.

---
background:
Using TcpStream with read/write timeout configured as underlying stream, ~~read()~~ [__read/write__](https://man7.org/linux/man-pages/man7/socket.7.html#:~:text=if%20no%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20data%20has%20been%20transferred%20and%20the%20timeout%20has%20been%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20reached%2C%20then%20%2D1%20is%20returned%20with%20errno%20set%20to%20EAGAIN%20or%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20EWOULDBLOCK%2C%20or%20EINPROGRESS%20(for%20connect(2))%20just%20as%20if%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20the%20socket%20was%20specified%20to%20be%20nonblocking.) may retrun WouldBlock/TimedOut error.

If some command run to failed, I try to logout() and create new session. and panic in logout().
```
[2022-07-01T14:53:07Z ERROR my_crate] fetch_insert_store failed: [Other(Io(Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }))]
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[97, 51, 53]`,
 right: `[97, 51, 54]`', /path/to/imap-3.0.0-alpha.7/src/client.rs:1396:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

qsdgy:
Try to turn tag assertion to an error check #230 
Also straight forward fix #229

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/231)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -8,7 +8,7 @@ use std::str;
 use std::sync::mpsc;
 
 use super::authenticator::Authenticator;
-use super::error::{Bad, Bye, Error, No, ParseError, Result, ValidateError};
+use super::error::{Bad, Bye, Error, FatalError, No, ParseError, Result, ValidateError};
 use super::extensions;
 use super::parse::*;
 use super::types::*;
@@ -167,6 +167,9 @@ pub struct Connection<T: Read + Write> {
 
     /// Tracks if we have read a greeting.
     pub greeting_read: bool,
+
+    /// Indicates if session failed to invalid state
+    pub fatal_error: Option<FatalError>,
 }
 
 /// A builder for the append command
@@ -335,6 +338,7 @@ impl<T: Read + Write> Client<T> {
                 tag: INITIAL_TAG,
                 debug: false,
                 greeting_read: false,
+                fatal_error: None,
             },
         }
     }
@@ -1340,6 +1344,8 @@ impl<T: Read + Write> Connection<T> {
     }
 
     fn run_command(&mut self, untagged_command: &str) -> Result<()> {
+        self.fatal_check()?;
+
         let command = self.create_command(untagged_command);
         let result = self.write_line(command.into_bytes().as_slice());
         if result.is_err() {
@@ -1397,7 +1403,19 @@ impl<T: Read + Write> Connection<T> {
                             ..
                         },
                     )) => {
-                        assert_eq!(tag.as_bytes(), match_tag.as_bytes());
+                        // check if tag matches
+                        if tag.as_bytes() != match_tag.as_bytes() {
+                            self.fatal_error =
+                                match tag.0.trim_start_matches(TAG_PREFIX).parse::<u32>() {
+                                    Ok(actual) => Some(FatalError::TagCorrupted {
+                                        expect: self.tag,
+                                        actual,
+                                    }),
+                                    Err(_) => Some(FatalError::Other),
+                                };
+                            break Err(Error::Fatal(self.fatal_error.unwrap()));
+                        }
+
                         Some(match status {
                             Status::Bad | Status::No | Status::Bye => Err((
                                 status,
@@ -1483,6 +1501,13 @@ impl<T: Read + Write> Connection<T> {
             eprintln!("C: {}", String::from_utf8(buf.to_vec()).unwrap());
         }
         Ok(())
+    }
+
+    fn fatal_check(&self) -> Result<()> {
+        match self.fatal_error {
+            None => Ok(()),
+            Some(err) => Err(Error::Fatal(err)),
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


jonhoo:
Thanks for taking this on! I don't think this should be a separate "special" error type — many of the other existing error variants are just as fatal as this one. I think I'd be fine with just adding an error variant for tag mismatch and return that as an error like any other instead of panicking.

qsdgy:
It's ok to flatten it.

Is it possible that we specify out what are fatal / resumable / determines errors ?
We could indicate it by doc / error type / fn on error / fn on session , providing more flexibility. Restart seesion on any error seems a bit wild.

src/client.rs
```diff
@@ -1341,7 +1341,11 @@ impl<T: Read + Write> Connection<T> {
 
     fn run_command(&mut self, untagged_command: &str) -> Result<()> {
         let command = self.create_command(untagged_command);
-        self.write_line(command.into_bytes().as_slice())
+        let result = self.write_line(command.into_bytes().as_slice());
+        if result.is_err() {
+            self.tag -= 1; // rollback tag increase in create_command()
```
jonhoo, line 1346: Following the discussion in https://github.com/jonhoo/rust-imap/issues/229, I'm okay with this change, though I think we should also add a `skip_tag` method to `Client` to allow manually incrementing the tag.

src/error.rs
```diff
@@ -104,6 +104,8 @@ pub enum Error {
     /// In response to a STATUS command, the server sent OK without actually sending any STATUS
     /// responses first.
     MissingStatusResponse,
+    /// Tag mismatch between client and server. New session must be created.
+    TagCorrupted(TagCorrupted),
```
jonhoo, line 108: I think the name `TagCorrupted` is misleading. A "corrupted" tag feels more like if a tag isn't a number for example, or has somehow otherwise been modified in-flight. What this represents (as the doc string indicates) is a `TagMismatch` or `Desynchronized`, so let's call it one of those.

src/error.rs
```diff
@@ -285,6 +290,36 @@ impl StdError for ValidateError {
...
+}
+
+impl StdError for TagCorrupted {
+    fn description(&self) -> &str {
+        "Tag is corrupted, session is in an inconsistent state"
+    }
+
+    fn cause(&self) -> Option<&dyn StdError> {
+        None
+    }
```
jonhoo, line 320: I think you can skip both of these `fn`s — they're not required and their default implementations will do the right thing.

jonhoo (changes requested):
Sorry it took me so long!

soywod:
Is there anything I can do to to help this PR to be merged? I also encounter the same issue when retrying a `.authenticate()` after an error (refreshing an expired OAuth 2.0 access token):

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[97, 49]`,
 right: `[97, 50]`', ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/imap-3.0.0-alpha.10/src/client.rs:1538:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

jonhoo:
I think my comments above still stand, so it's really just a matter of making those changes. If @qsdgy is still interested in picking that up, then that's probably easiest, but if they're otherwise preoccupied or no longer need this, then feel free to open a new PR that builds on this one so we can land it :+1:

soywod:
> then feel free to open a new PR that builds on this one so we can land it 👍

I opened a new PR and tried to apply the changes your were requesting: https://github.com/jonhoo/rust-imap/pull/284

jonhoo:
Closing this in favor of https://github.com/jonhoo/rust-imap/pull/284 :+1:

Issue #131, opened 2019-08-21
Add support for UIDPLUS repsonses


dignifiedquire:
When moving or copying mails, the `UIDPLUS` extension returns the destination ids. Currently this information is not exposed, it would be great to make this available. 

Reference: https://tools.ietf.org/html/rfc4315

Celti:
It looks like we need support for this in [`tokio-imap/imap-proto`](https://github.com/djc/tokio-imap/imap-proto) in order to properly extract that information. I've submitted djc/tokio-imap#66 to hopefully get that rolling.

mmirate:
[Well, that was easy.](djc/tokio-imap#109) Mostly.

PR #232, opened 2022-07-06
Add support for APPENDUID response data

bitfehler:
If the `UIDPLUS` extension is supported, the server will reply to
`APPEND` commands with the UID of the new message. This can even be a
list of UIDs if the `MULTIAPPEND` extension is also supported.

Make this information available to the user as the result of an
`AppendCmd`. The added doc strings have links to the relevant RFCs.

Related to #131.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/232)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -215,7 +215,7 @@ impl<'a, T: Read + Write> AppendCmd<'a, T> {
     ///
     /// Note: be sure to set flags and optional date before you
     /// finish the command.
-    pub fn finish(&mut self) -> Result<()> {
+    pub fn finish(&mut self) -> Result<Appended> {
         let flagstr = self
             .flags
             .clone()
@@ -246,7 +246,9 @@ impl<'a, T: Read + Write> AppendCmd<'a, T> {
         self.session.stream.write_all(self.content)?;
         self.session.stream.write_all(b"\r\n")?;
         self.session.stream.flush()?;
-        self.session.read_response().map(|_| ())
+        self.session
+            .read_response()
+            .and_then(|(lines, _)| parse_append(&lines, &mut self.session.unsolicited_responses_tx))
     }
 }
 
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -124,6 +124,40 @@ pub fn parse_expunge(
     }
 }
 
+pub fn parse_append(
+    mut lines: &[u8],
+    unsolicited: &mut mpsc::Sender<UnsolicitedResponse>,
+) -> Result<Appended> {
+    let mut appended = Appended::default();
+
+    loop {
+        match imap_proto::parser::parse_response(lines) {
+            Ok((rest, Response::Done { status, code, .. })) => {
+                lines = rest;
+                assert_eq!(status, imap_proto::Status::Ok);
+
+                if let Some(ResponseCode::AppendUid(validity, uids)) = code {
+                    appended.uid_validity = Some(validity);
+                    appended.uids = Some(uids);
+                }
+            }
+            Ok((rest, data)) => {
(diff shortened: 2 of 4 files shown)


bitfehler:
If this is of interest, I'd be happy to add `COPYUID` as well...

src/types/appended.rs
```diff
@@ -0,0 +1,40 @@
...
+impl Default for Appended {
+    fn default() -> Appended {
+        Appended {
+            uid_validity: None,
+            uids: None,
+        }
+    }
+}
+
+impl fmt::Display for Appended {
```
jonhoo, line 32: Why implement `Dispay` for this type? I don't think that's actually going to be all that useful compared to the `Debug` that's already derived.

src/client.rs
```diff
@@ -215,7 +215,7 @@ impl<'a, T: Read + Write> AppendCmd<'a, T> {
     ///
     /// Note: be sure to set flags and optional date before you
     /// finish the command.
-    pub fn finish(&mut self) -> Result<()> {
+    pub fn finish(&mut self) -> Result<Appended> {
```
jonhoo, line 218: Hmm, so, this is technically a breaking change. In practice it _should_ be very rare that someone relies on `()` specifically being the return type, but it is possible. Most likely if they have a call to `finish` somewhere without a trailing `;`, meaning its return value will end up being the evaluated value of the surrounding block as well.

How about we keep the return type of `finish` as-is, mark it as deprecated, and introduce a new method (bikeshed: `send` or `complete`) that returns `Result<Appended>`? `finish` can then just be a call to that method with a `.map(|_| ())` for the return value.

jonhoo (changes requested):
Nice, thank you! And yes, `COPYUID` would be good to add, though happy to take that in a separate PR if you'd prefer.

src/types/appended.rs
bitfehler: Right, I don't remember :innocent: . I think I used it for something at some point, but definitely no longer. Will remove. Do you prefer force pushed amendments or new commits?

jonhoo (approved):

src/types/appended.rs
jonhoo: New commits are generally better for changes after the first review since GitHub is _terrible_ at rendering diffs across squashes :sweat_smile:

jonhoo (changes requested):
Oh, can you please add a test for the new functionality too?

bitfehler:
Added a parse test much like the other one, I hope that's cool for now. I am happy to look into the integration tests, however I will be AFK for two weeks.

src/parse.rs
```diff
@@ -698,4 +698,26 @@ mod tests {
         assert_eq!(first.body(), None);
         assert_eq!(first.header(), None);
     }
+
+    #[test]
+    fn parse_append_uid() {
+        // If the user has enabled UIDPLUS (RFC 4315), the response contains an APPENDUID
+        // response code followed by the UIDVALIDITY of the destination mailbox and the
+        // UID assigned to the appended message in the destination mailbox.
+        // If the MULTIAPPEND extension is also used, there can be multiple UIDs.
```
jonhoo, line 707: Should we then also test the multi-append case?

jonhoo:
Looks like there's also now a conflict in `src/parse.rs`.

bitfehler:
Sorry for the wait! I rebased on current master and added another test for the MULTIAPPEND response. I hope this is fine for this PR. I'll start looking into the integration tests to see if I can add some for all the stuff I added now.

jonhoo (approved):
Looks great, thanks!

PR #233, opened 2022-07-14
Expose MODSEQ attribute in FETCH responses

bitfehler:
If the QRESYNC (RFC 7162) extension is being used, a FETCH response to a
STORE or UID STORE command with the UNCHANGEDSINCE query attribute will
return the mod sequence ID of the performed operation. This information
is crucial for building efficient caching clients.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/233)

### DIFF
--- a/src/types/fetch.rs
+++ b/src/types/fetch.rs
@@ -115,6 +115,17 @@ impl<'a> Fetch<'a> {
         &self.flags[..]
     }
 
+    /// Contains the mod sequence of the performed operation if in response to a `STORE` or
+    /// `UID STORE` with `UNCHANGEDSINCE` in the query arguments.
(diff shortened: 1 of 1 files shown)


jonhoo (changes requested):
This looks good, though I'd like for it to also mention `QRESYNC` if it only works with servers with that capability. Can we also link to the relevant part of the spec?

bitfehler:
> This looks good, though I'd like for it to also mention `QRESYNC` if it only works with servers with that capability. Can we also link to the relevant part of the spec?

I am assuming you mean in the doc-string? Added a reference with link to it.

jonhoo (approved):
Yes indeed!

PR #234, opened 2022-07-14
Expose HIGHESTMODSEQ value in EXPUNGE response

bitfehler:
If the `QRESYNC` extension (RFC 7162) is being used, `EXPUNGE` responses
will return the new highest mod sequence for the mailbox after the
expunge operation. Access to this value is quite valuable for caching
clients.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/234)

### DIFF
--- a/src/types/deleted.rs
+++ b/src/types/deleted.rs
@@ -1,13 +1,15 @@
 use super::{Seq, Uid};
 use std::ops::RangeInclusive;
 
-/// An enum representing message sequence numbers or UID sequence sets returned
-/// in response to a `EXPUNGE` command.
+/// A struct containing message sequence numbers or UID sequence sets and a mod
+/// sequence returned in response to a `EXPUNGE` command.
 ///
 /// The `EXPUNGE` command may return several `EXPUNGE` responses referencing
 /// message sequence numbers, or it may return a `VANISHED` response referencing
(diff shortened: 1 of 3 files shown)


bitfehler:
This one is a bit more involved. I tried to do it in such a way that the API stays essentially the same except for one added field. But I am open for suggestions if you think this would be better handled differently.

jonhoo (changes requested):
This looks great! Would you mind adding a test or two to make sure this actually works too?

src/types/deleted.rs
```diff
@@ -39,7 +41,16 @@ use std::ops::RangeInclusive;
 /// # }
 /// ```
 #[derive(Debug, Clone)]
-pub enum Deleted {
+pub struct Deleted {
```
jonhoo, line 44: Let's mark this as `#[non_exhaustive]` in case we realize there are other expunge details we want to add in the future.

jonhoo (approved):
No, that's perfect,thanks!

jonhoo:
I do wonder if your CI mail server (Greenmail) supports QRESYNC.. If so, we should be able to write an integration test too, not just a parsing test. But no need for that in this PR.

PR #235, opened 2022-07-16
add support for the imap quota extension (RFC 2087)

urkle:
Here is my first-stab at adding in support for the quota extension (already in imap-proto).

A few things I don't like with it right now.

1. it exposes some of the imap-proto structs (QuotaResource)
  - I can create local QuotaResource and map them.
2. the "getquotaroot" is "fun" as it returns 2 resources, (1) QUOTAROOT and (n) QUOTAs.
  - The issue here is that I have to run parse_many_into twice to get both resource sets. (unless I just don't use parse_many_into and just duplicate most of that in the method which I may do)
  - this also causes a difference in response types..  e.g. set_quota and get_quota return a local Quota struct, but I the get_quota_root returns the imap-proto Quota struct (I did not see any easy way to convert ouroboros structs).

Anyways, it's starting point.. Please provide feedback.  (this and #227)  are two pieces of functionality I'll need for the application I'm building.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/235)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -216,14 +216,7 @@ impl<'a, T: Read + Write> AppendCmd<'a, T> {
     /// Note: be sure to set flags and optional date before you
     /// finish the command.
     pub fn finish(&mut self) -> Result<Appended> {
-        let flagstr = self
-            .flags
-            .clone()
-            .into_iter()
-            .filter(|f| *f != Flag::Recent)
-            .map(|f| f.to_string())
-            .collect::<Vec<String>>()
-            .join(" ");
+        let flagstr = iter_join(self.flags.iter().filter(|f| **f != Flag::Recent), " ");
 
         let datestr = if let Some(date) = self.date {
             format!(" \"{}\"", date.format("%d-%h-%Y %T %z"))
--- a/src/client.rs
+++ b/src/client.rs
@@ -12,6 +12,7 @@ use super::error::{Bad, Bye, Error, No, ParseError, Result, ValidateError};
 use super::extensions;
 use super::parse::*;
 use super::types::*;
+use super::utils::*;
 
 static TAG_PREFIX: &str = "a";
 const INITIAL_TAG: u32 = 0;
@@ -1375,6 +1376,48 @@ impl<T: Read + Write> Session<T> {
         .and_then(|lines| MyRightsResponse::parse(lines, &mut self.unsolicited_responses_tx))
     }
 
+    /// The [`SETQUOTA` command](https://datatracker.ietf.org/doc/html/rfc2087#section-4.1)
+    ///
+    /// Updates the resource limits for a mailbox. Any previous limits for the named quota
+    /// are discarded.
+    ///
+    /// Returns the updated quota.
+    pub fn set_quota(
+        &mut self,
+        quota_root: impl AsRef<str>,
+        limits: &[QuotaResourceLimit<'_>],
+    ) -> Result<QuotaResponse> {
+        let limits = iter_join(limits.iter(), " ");
+        self.run_command_and_read_response(&format!(
+            "SETQUOTA {} ({})",
+            validate_str("SETQUOTA", "quota_root", quota_root.as_ref())?,
+            limits,
+        ))
+        .and_then(|lines| QuotaResponse::parse(lines, &mut self.unsolicited_responses_tx))
+    }
+
+    /// The [`GETQUOTA` command](https://datatracker.ietf.org/doc/html/rfc2087#section-4.2)
+    ///
+    /// Returns the quota information for the specified quota root
+    pub fn get_quota(&mut self, quota_root: impl AsRef<str>) -> Result<QuotaResponse> {
+        self.run_command_and_read_response(&format!(
+            "GETQUOTA {}",
+            validate_str("GETQUOTA", "quota_root", quota_root.as_ref())?
+        ))
+        .and_then(|lines| QuotaResponse::parse(lines, &mut self.unsolicited_responses_tx))
+    }
+
+    /// The [`GETQUOTAROOT` command](https://datatracker.ietf.org/doc/html/rfc2087#section-4.3)
+    ///
+    /// Returns the quota roots along with their quota information for the specified mailbox
+    pub fn get_quota_root(&mut self, mailbox_name: impl AsRef<str>) -> Result<QuotaRootResponse> {
+        self.run_command_and_read_response(&format!(
+            "GETQUOTAROOT {}",
+            validate_str("GETQUOTAROOT", "mailbox", mailbox_name.as_ref())?
+        ))
+        .and_then(|lines| QuotaRootResponse::parse(lines, &mut self.unsolicited_responses_tx))
+    }
+
     // these are only here because they are public interface, the rest is in `Connection`
     /// Runs a command and checks if it returns OK.
     pub fn run_command_and_check_ok(&mut self, command: impl AsRef<str>) -> Result<()> {
@@ -1633,6 +1676,17 @@ pub(crate) mod testutils {
             })
             .to_string()
         );
+    }
+
+    pub(crate) fn assert_quota_resource(
+        resource: &QuotaResource<'_>,
+        name: QuotaResourceName<'_>,
+        limit: u64,
+        usage: u64,
+    ) {
+        assert_eq!(resource.name, name);
+        assert_eq!(resource.usage, usage);
+        assert_eq!(resource.limit, limit);
     }
 }
 
@@ -2367,6 +2421,278 @@ mod tests {
             "Invalid myrights command"
         );
         assert!(matches!(acl, Err(Error::Parse(_))));
+    }
+
+    #[test]
+    fn set_quota() {
+        let response = b"* QUOTA my_root (STORAGE 10 500)\r\n\
+                a1 OK completed\r\n"
+            .to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota = session
+            .set_quota(
+                "my_root",
+                &[QuotaResourceLimit {
+                    name: QuotaResourceName::Storage,
+                    amount: 500,
+                }],
+            )
+            .unwrap();
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 SETQUOTA \"my_root\" (STORAGE 500)\r\n".to_vec(),
+            "Invalid setquota command"
+        );
+        let quota = quota.parsed().as_ref().unwrap();
+        assert_eq!(quota.root_name, "my_root");
+        assert_eq!(quota.resources.len(), 1);
+        assert_quota_resource(&quota.resources[0], QuotaResourceName::Storage, 500, 10);
+    }
+
+    #[test]
+    fn set_quota_no_such_quota_root() {
+        let response = b"a1 NO no such quota root\r\n".to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota = session.set_quota(
+            "my_root",
+            &[QuotaResourceLimit {
+                name: QuotaResourceName::Storage,
+                amount: 500,
+            }],
+        );
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 SETQUOTA \"my_root\" (STORAGE 500)\r\n".to_vec(),
+            "Invalid setquota command"
+        );
+        assert!(matches!(quota, Err(Error::No(_))));
+    }
+
+    #[test]
+    fn set_quota_invalid_no_quota_lines() {
+        let response = b"a1 OK completed\r\n".to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota = session
+            .set_quota(
+                "my_root",
+                &[QuotaResourceLimit {
+                    name: QuotaResourceName::Storage,
+                    amount: 500,
+                }],
+            )
+            .unwrap();
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 SETQUOTA \"my_root\" (STORAGE 500)\r\n".to_vec(),
+            "Invalid setquota command"
+        );
+        assert_eq!(quota.parsed(), &None);
+    }
+
+    #[test]
+    fn set_quota_invalid_too_many_quota_lines() {
+        let response = b"* QUOTA my_root (STORAGE 10 500)\r\n\
+                * QUOTA my_root2 (STORAGE 10 500)\r\n\
+                a1 OK completed\r\n"
+            .to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota = session.set_quota(
+            "my_root",
+            &[QuotaResourceLimit {
+                name: QuotaResourceName::Storage,
+                amount: 500,
+            }],
+        );
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 SETQUOTA \"my_root\" (STORAGE 500)\r\n".to_vec(),
+            "Invalid setquota command"
+        );
+        assert!(matches!(quota, Err(Error::Parse(_))));
+    }
+
+    #[test]
+    fn get_quota() {
+        let response = b"* QUOTA my_root ()\r\n\
+            a1 OK completed\r\n"
+            .to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota = session.get_quota("my_root").unwrap();
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 GETQUOTA \"my_root\"\r\n".to_vec(),
+            "Invalid getquota command"
+        );
+        let quota = quota.parsed().as_ref().unwrap();
+        assert_eq!(quota.root_name, "my_root");
+        assert_eq!(quota.resources, vec![]);
+    }
+
+    #[test]
+    fn get_quota_with_limits() {
+        let response = b"* QUOTA my_root (STORAGE 10 500)\r\n\
+                a1 OK completed\r\n"
+            .to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota = session.get_quota("my_root").unwrap();
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 GETQUOTA \"my_root\"\r\n".to_vec(),
+            "Invalid getquota command"
+        );
+        let quota = quota.parsed().as_ref().unwrap();
+        assert_eq!(quota.root_name, "my_root");
+        assert_eq!(quota.resources.len(), 1);
+        assert_quota_resource(&quota.resources[0], QuotaResourceName::Storage, 500, 10);
+    }
+
+    #[test]
+    fn get_quota_with_invalid_no_quota_lines() {
+        let response = b"a1 OK completed\r\n".to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota = session.get_quota("my_root").unwrap();
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 GETQUOTA \"my_root\"\r\n".to_vec(),
+            "Invalid getquota command"
+        );
+        assert_eq!(quota.parsed(), &None);
+    }
+
+    #[test]
+    fn get_quota_with_invalid_too_many_quota_lines() {
+        let response = b"* QUOTA my_root (STORAGE 10 500)\r\n\
+                * QUOTA my_root2 (STORAGE 10 500)\r\n\
+                a1 OK completed\r\n"
+            .to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota = session.get_quota("my_root");
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 GETQUOTA \"my_root\"\r\n".to_vec(),
+            "Invalid getquota command"
+        );
+        assert!(matches!(quota, Err(Error::Parse(_))));
+    }
+
+    #[test]
+    fn get_quota_with_no_such_quota_root() {
+        let response = b"a1 NO no such quota root\r\n".to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota = session.get_quota("my_root");
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 GETQUOTA \"my_root\"\r\n".to_vec(),
+            "Invalid getquota command"
+        );
+        assert!(matches!(quota, Err(Error::No(_))));
+    }
+
+    #[test]
+    fn get_quota_root_with_no_root() {
+        let response = b"* QUOTAROOT INBOX\r\n\
+                a1 OK completed\r\n"
+            .to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota_root = session.get_quota_root("INBOX").unwrap();
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 GETQUOTAROOT \"INBOX\"\r\n".to_vec(),
+            "Invalid getquotaroot command"
+        );
+        assert_eq!(quota_root.mailbox_name(), "INBOX");
+        assert_eq!(
+            quota_root.quota_root_names().collect::<Vec<_>>(),
+            Vec::<&str>::new()
+        );
+        assert_eq!(quota_root.quotas(), vec![]);
+    }
+
+    #[test]
+    fn get_quota_root_no_such_mailbox() {
+        let response = b"a1 NO no such mailboxd\r\n".to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota_root = session.get_quota_root("INBOX");
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 GETQUOTAROOT \"INBOX\"\r\n".to_vec(),
+            "Invalid getquotaroot command"
+        );
+        assert!(matches!(quota_root, Err(Error::No(_))));
+    }
+
+    #[test]
+    fn get_quota_root_invalid_no_quota_root_lines() {
+        let response = b"a1 OK completed\r\n".to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota_root = session.get_quota_root("INBOX");
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 GETQUOTAROOT \"INBOX\"\r\n".to_vec(),
+            "Invalid getquotaroot command"
+        );
+        assert!(matches!(quota_root, Err(Error::Parse(_))));
+    }
+
+    #[test]
+    fn get_quota_root_invalid_too_many_quota_root_lines() {
+        let response = b"* QUOTAROOT INBOX\r\n\
+                * QUOTAROOT INBOX\r\n\
+                a1 OK completed\r\n"
+            .to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota_root = session.get_quota_root("INBOX");
+        println!("Resp: {:?}", quota_root);
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 GETQUOTAROOT \"INBOX\"\r\n".to_vec(),
+            "Invalid getquotaroot command"
+        );
+        assert!(matches!(quota_root, Err(Error::Parse(_))));
+    }
+
+    #[test]
+    fn get_quota_root_with_root() {
+        let response = b"* QUOTAROOT INBOX my_root\r\n\
+                * QUOTA my_root (STORAGE 10 500)\r\n\
+                a1 OK completed\r\n"
+            .to_vec();
+        let mock_stream = MockStream::new(response);
+        let mut session = mock_session!(mock_stream);
+        let quota_root = session.get_quota_root("INBOX").unwrap();
+        assert_eq!(
+            session.stream.get_ref().written_buf,
+            b"a1 GETQUOTAROOT \"INBOX\"\r\n".to_vec(),
+            "Invalid getquotaroot command"
+        );
+        assert_eq!(quota_root.mailbox_name(), "INBOX");
+        assert_eq!(
+            quota_root.quota_root_names().collect::<Vec<_>>(),
+            vec!["my_root"]
+        );
+        assert_eq!(quota_root.quotas().len(), 1);
+        assert_eq!(quota_root.quotas().first().unwrap().root_name, "my_root");
+        assert_eq!(
+            quota_root.quotas().first().unwrap().resources,
+            vec![QuotaResource {
+                name: QuotaResourceName::Storage,
+                usage: 10,
+                limit: 500,
+            }]
+        );
     }
 
     #[test]
--- a/src/client.rs
+++ b/src/client.rs
@@ -2556,9 +2556,10 @@ mod tests {
 
     #[test]
     fn get_quota_with_limits() {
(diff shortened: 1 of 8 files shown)


src/client.rs
```diff
@@ -1280,6 +1280,48 @@ impl<T: Read + Write> Session<T> {
...
     }
 
+    /// The [`SETQUOTA` command](https://datatracker.ietf.org/doc/html/rfc2087#section-4.1)
+    ///
+    /// Updates the resource limits for a mailbox. Any previous limits for the named quota
+    /// are discarded.
+    /// Returns the updated quota
+    pub fn set_quota<S: AsRef<str>>(
+        &mut self,
+        quota_root: S,
```
jonhoo, line 1290: Let's use `impl AsRef<str>` here too.

src/client.rs
```diff
@@ -1280,6 +1280,48 @@ impl<T: Read + Write> Session<T> {
...
+        limits: Vec<QuotaLimit>,
```
jonhoo, line 1291: This should be `&[QuotaLimit]` — no need to require an owned `Vec`.

src/client.rs
```diff
@@ -1280,6 +1280,48 @@ impl<T: Read + Write> Session<T> {
...
+    ) -> Result<Quota> {
+        self.run_command_and_read_response(&format!(
+            "SETQUOTA {} ({})",
+            validate_str("SETQUOTA", "quota_root", quota_root.as_ref())?,
+            limits
+                .iter()
+                .map(|e| e.to_string())
+                .collect::<Vec<_>>()
```
jonhoo, line 1299: Feels unfortunate to have to allocate a `Vec` just to print this. There are a couple of other places I think we might be doing something like this, so maybe let's add a helper like this one from Cargo that can join straight from an iterator:
https://github.com/rust-lang/cargo/blob/576356f414bd663526b24e227a6afefe5ecf554b/src/cargo/util/mod.rs#L73-L97

src/client.rs
```diff
@@ -1280,6 +1280,48 @@ impl<T: Read + Write> Session<T> {
...
+        &mut self,
+        quota_root: S,
+        limits: Vec<QuotaLimit>,
+    ) -> Result<Quota> {
+        self.run_command_and_read_response(&format!(
+            "SETQUOTA {} ({})",
+            validate_str("SETQUOTA", "quota_root", quota_root.as_ref())?,
+            limits
+                .iter()
+                .map(|e| e.to_string())
```
jonhoo, line 1298: This, too, feels unfortunate. Technically we don't need `String`s, all we need is to `write!("{}", e)` into the string we already have. Which you can do with `write!(&mut somestring, "{}", e)` as long as `std::fmt::Write` is in scope.

src/client.rs
```diff
@@ -1280,6 +1280,48 @@ impl<T: Read + Write> Session<T> {
         .and_then(|lines| parse_id_seq(&lines, &mut self.unsolicited_responses_tx))
     }
 
+    /// The [`SETQUOTA` command](https://datatracker.ietf.org/doc/html/rfc2087#section-4.1)
+    ///
+    /// Updates the resource limits for a mailbox. Any previous limits for the named quota
+    /// are discarded.
+    /// Returns the updated quota
```
jonhoo, line 1287:
```suggestion
    ///
    /// Returns the updated quota.
```

src/client.rs
```diff
@@ -1280,6 +1280,48 @@ impl<T: Read + Write> Session<T> {
...
+                .iter()
+                .map(|e| e.to_string())
+                .collect::<Vec<_>>()
+                .join(" "),
+        ))
+        .and_then(|lines| Quota::parse(lines, &mut self.unsolicited_responses_tx))
+    }
+
+    /// The [`GETQUOTA` command](https://datatracker.ietf.org/doc/html/rfc2087#section-4.2)
+    ///
```
jonhoo, line 1306: This seems to be missing an actual description.

src/client.rs
```diff
@@ -1280,6 +1280,48 @@ impl<T: Read + Write> Session<T> {
...
+    pub fn get_quota<S: AsRef<str>>(&mut self, quota_root: S) -> Result<Quota> {
+        self.run_command_and_read_response(&format!(
+            "GETQUOTA {}",
+            validate_str("GETQUOTA", "quota_root", quota_root.as_ref())?
+        ))
+        .and_then(|lines| Quota::parse(lines, &mut self.unsolicited_responses_tx))
+    }
+
+    /// The [`GETQUOTAROOT` command](https://datatracker.ietf.org/doc/html/rfc2087#section-4.3)
+    ///
```
jonhoo, line 1316: Same here — missing description.

src/types/quota.rs
```diff
@@ -0,0 +1,160 @@
+use crate::error::{Error, ParseError};
+use crate::parse::{parse_many_into, try_handle_unilateral, MapOrNot};
+use crate::types::UnsolicitedResponse;
+use imap_proto::Response;
+use ouroboros::self_referencing;
+use std::fmt::{Display, Formatter};
+use std::sync::mpsc;
+
+pub struct QuotaLimit {
+    // The resource type
```
jonhoo, line 10: I'm guessing you intended these to be `///`?

src/types/quota.rs
```diff
@@ -0,0 +1,160 @@
...
+    /// Any other string (for future RFCs)
+    Atom(String),
+}
+
+impl Display for QuotaLimitName {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            QuotaLimitName::Storage => write!(f, "STORAGE"),
+            QuotaLimitName::Message => write!(f, "MESSAGE"),
+            QuotaLimitName::Atom(s) => write!(f, "{}", s),
```
jonhoo, line 37: Should this be escaping/quoting?

src/types/quota.rs
```diff
@@ -0,0 +1,160 @@
...
+        QuotaTryBuilder {
+            data: owned,
+            quota_builder: |input| {
+                let mut lines: &[u8] = input;
+
+                // There should only be one single QUOTA response
+                while !lines.is_empty() {
+                    match imap_proto::parser::parse_response(lines) {
+                        Ok((_rest, Response::Quota(a))) => {
+                            return Ok(a);
```
jonhoo, line 64: As we discussed in the other PR, this probably shouldn't be early-returning.

src/types/quota.rs
```diff
@@ -0,0 +1,160 @@
...
+                    input,
+                    &mut quota_roots,
+                    unsolicited,
+                    |response| match response {
+                        Response::QuotaRoot(q) => Ok(MapOrNot::Map(q)),
+                        Response::Quota(_) => Ok(MapOrNot::Ignore),
+                        resp => Ok(MapOrNot::Not(resp)),
+                    },
+                )?;
+                parse_many_into(input, &mut quotas, unsolicited, |response| match response {
```
jonhoo, line 126: I _think_ this will have the effect of sending any unsolicited items in `input` twice. We may need to make `parse_many_into` able to get a `None` for `unsolicited` so that for the second pass over the input ignores unsolicited things. Or I _think_ you could maybe do it by just mapping anything that's not `::Quota` to `::Ignore`.

jonhoo: Iterating over the input twice to parse it is somewhat unfortunate, though I think it's okay. It means we avoid duplicating the parsing loop, and quota replies are likely to be fairly short and not performance sensitive anyway.

src/types/quota.rs
```diff
@@ -0,0 +1,160 @@
...
+            },
+        }
+        .try_build()
+    }
+
+    pub fn mailbox_name(&self) -> &str {
+        &*self.borrow_inner().quota_root.mailbox_name
+    }
+
+    pub fn quota_root_names(&self) -> Vec<&str> {
```
jonhoo, line 148: Let's have this return `impl Iterator<Item = &str>` instead. No need for us to dictate the data structure to use.

jonhoo (changes requested):
I took a pass!

src/types/quota.rs
urkle: I do not believe so, as the spec calls this an "ATOM" which does not allow for any character that would require quoting.   (e.g. same as what is returned by CAPABILITIES or used in FETCH and SEARCH () blocks)

src/types/quota.rs
urkle: yeah.. already have this reworked in my next push

src/types/quota.rs
urkle: yeah..  I didn't like it..  I have reworked this now by creating a parse_many_into2 function that iterates once.

src/client.rs
urkle: I switched it to the cargo iter_into (as well as the other usage I found)

urkle:
@jonhoo a bunch of changes and I applied all your feedback.. the result is much cleaner IMHO that the first draft.

And once #228 is merged in I'll add more integration tests..

src/client.rs
wookietreiber: What is the difference between `<S: AsRef<str>>` and `impl AsRef<str>`? Is there some documentation on when to use which variant? (I don't know what to web-search for, I'm watching some repos to learn best practices and I haven't seen `impl AsRef` before, only the other variant.)

src/client.rs
```diff
@@ -124,6 +124,38 @@ fn validate_sequence_set(
     Ok(value)
 }
 
+/// Lovingly borrowed from the cargo crate
+///
+/// Joins an iterator of [std::fmt::Display]'ables into an output writable
+pub fn iter_join_onto<W, I, T>(mut w: W, iter: I, delim: &str) -> std::fmt::Result
```
jonhoo, line 130: These should definitely not be `pub` :sweat_smile: Can you change them to `pub(crate)`?

src/client.rs
```diff
@@ -216,14 +248,13 @@ impl<'a, T: Read + Write> AppendCmd<'a, T> {
...
-            .clone()
-            .into_iter()
-            .filter(|f| *f != Flag::Recent)
-            .map(|f| f.to_string())
-            .collect::<Vec<String>>()
-            .join(" ");
+        let flagstr = iter_join(
+            self.flags
+                .clone()
+                .into_iter()
```
jonhoo, line 254: Why the `clone` here? I think `.iter()` should be good enough, no?

src/client.rs
```diff
@@ -1282,6 +1313,47 @@ impl<T: Read + Write> Session<T> {
         .and_then(|lines| parse_id_seq(&lines, &mut self.unsolicited_responses_tx))
     }
 
+    /// The [`SETQUOTA` command](https://datatracker.ietf.org/doc/html/rfc2087#section-4.1)
+    ///
+    /// Updates the resource limits for a mailbox. Any previous limits for the named quota
+    /// are discarded.
+    /// Returns the updated quota.
```
jonhoo, line 1320:
```suggestion
    ///
    /// Returns the updated quota.
```
Note the added newline. Without the extra newline, the last line will be rendered as part of the previous paragraph.

src/parse.rs
```diff
@@ -71,6 +71,54 @@ where
...
+    Map1(T),
+    Map2(U),
+    Not(Response<'a>),
+    #[allow(dead_code)]
+    Ignore,
+}
+
+/// Parse many `T` or `U` Responses with `F` and extend `into1` or `into2` with them.
+/// Responses other than `T` or `U` go into the `unsolicited` channel.
+pub(crate) fn parse_many_into2<'input, T, U, F>(
```
jonhoo, line 84: Now that we have this, we can make `parse_many_into` just call this one with `()` (or even `!`) as `U`.

src/types/quota.rs
```diff
@@ -0,0 +1,233 @@
...
+use crate::types::UnsolicitedResponse;
+use imap_proto::Response;
+use ouroboros::self_referencing;
+use std::borrow::Cow;
+use std::fmt::{Display, Formatter};
+use std::sync::mpsc;
+
+/// From [SETQUOTA Resource limit](https://datatracker.ietf.org/doc/html/rfc2087#section-4.1)
+///
+/// Used by [Session::set_quota]
```
jonhoo, line 12:
```suggestion
/// Used by [`Session::set_quota`].
```

src/types/quota.rs
```diff
@@ -0,0 +1,233 @@
...
+
+impl Display for QuotaResourceLimit<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{} {}", self.name, self.amount)
+    }
+}
+
+/// From [Resources](https://datatracker.ietf.org/doc/html/rfc2087#section-3)
+///
+/// Used by [QuotaLimit], and [QuotaResource]
```
jonhoo, line 28:
```suggestion
/// Used by [`QuotaLimit`] and `[QuotaResource`].
```

src/types/quota.rs
```diff
@@ -0,0 +1,233 @@
...
+                // There should only be one single QUOTA response
+                parse_many_into(input, &mut quotas, unsolicited, |response| match response {
+                    Response::Quota(q) => Ok(MapOrNot::Map(Quota::from_imap_proto(q))),
+                    resp => Ok(MapOrNot::Not(resp)),
+                })?;
+
+                if quotas.is_empty() {
+                    Err(Error::Parse(ParseError::Invalid(input.to_vec())))
+                } else {
+                    Ok(quotas.remove(0))
```
jonhoo, line 83: This, too, would benefit from a `parse_until_done` like we've talked about in the other PR.

src/types/quota.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    /// The defined resources with their usage and limits (could be empty)
+    pub fn resources(&self) -> &[QuotaResource<'_>] {
+        self.borrow_quota().resources()
+    }
+
+    /// Access to the wrapped [Quota] struct
+    pub fn quota(&self) -> &Quota<'_> {
+        self.borrow_quota()
+    }
```
jonhoo, line 103: I feel like we should _either_ expose the `&Quota` _or_ accessors to the individual members (`root_name()` and `resources()`). Doing both feels a little weird. Given that this is already a `QuotaResponse`, it also feels weird to call `.quota()` on it. Thinking about it some more, how about we just `impl Deref<Target = Quota>` and then remove all three of these methods and _just_ rely on methods on `Quota` itself?

src/types/quota.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/// From [QUOTA Response](https://datatracker.ietf.org/doc/html/rfc2087#section-5.1)
+///
+/// Used by [QuotaResponse] and [QuotaRoot]
```
jonhoo, line 108:
```suggestion
/// Used by [`QuotaResponse`] and [`QuotaRoot`].
```

src/types/quota.rs
```diff
@@ -0,0 +1,233 @@
...
+                    limit: e.limit,
+                })
+                .collect(),
+        }
+    }
+}
+
+/// From [QUOTA Response](https://datatracker.ietf.org/doc/html/rfc2087#section-5.1)
+///
+/// The quota resource sub pieces int he Quota response. Used by [Quota]
```
jonhoo, line 150:
```suggestion
/// The quota resource sub-pieces in a [`Quota`].
```

src/types/quota.rs
```diff
@@ -0,0 +1,233 @@
...
+    pub name: QuotaResourceName<'a>,
+    /// current usage of the resource
+    pub usage: u64,
+    /// resource limit
+    pub limit: u64,
+}
+
+/// From [QUOTAROOT Response](https://datatracker.ietf.org/doc/html/rfc2087#section-5.2)
+///
+/// Used by [Session::get_quota_root]
```
jonhoo, line 163:
```suggestion
/// Used by [`Session::get_quota_root`].
```

src/types/quota.rs
```diff
@@ -0,0 +1,233 @@
...
+                        Response::Quota(q) => Ok(MapOrNot2::Map2(Quota::from_imap_proto(q))),
+                        resp => Ok(MapOrNot2::Not(resp)),
+                    },
+                )?;
+
+                if quota_roots.is_empty() {
+                    Err(Error::Parse(ParseError::Invalid(input.to_vec())))
+                } else {
+                    Ok(InnerQuotaRoot {
+                        quota_root: quota_roots.remove(0),
```
jonhoo, line 206: Should this error if there's >1?

jonhoo: And also do something other than panic if there are none (also a question for if we implement a helper)?

jonhoo (changes requested):
It does look a lot cleaner! Found another few tweaks we should make.

src/client.rs
urkle: oh.. good catch.

src/types/quota.rs
urkle: I reworked the handling here and in the Quota::parse to explicitly handle 1 vs not 1 and added tests to ensure (no panic) handling of those scenarios.

src/client.rs
jonhoo: `<>` generics are specifiable using the turbofish (e.g., `client.set_quota::<String>(...)`), whereas `impl Trait` is not. That may seem like a restriction, but in practice it allows you to ensure that "important" generics are made more explicit, whereas "simple" generics (like `impl AsRef<str>`) sort of fade into the background.

src/types/quota.rs
jonhoo: Did you mean to add a `Deref` impl?

src/types/quota.rs
```diff
@@ -0,0 +1,231 @@
...
+                // There should only be one single QUOTA response
+                parse_many_into(input, &mut quotas, unsolicited, |response| match response {
+                    Response::Quota(q) => Ok(MapOrNot::Map(Quota::from_imap_proto(q))),
+                    resp => Ok(MapOrNot::Not(resp)),
+                })?;
+
+                match quotas.len() {
+                    1 => Ok(quotas.remove(0)),
+                    _ => Err(Error::Parse(ParseError::Invalid(input.to_vec()))),
+                }
```
jonhoo, line 82: I'd still prefer to see this pattern encapsulated as a `parse_until_done` helper that calls `parse_many_into` but expects to only find a single elements. It could also probably use `Option` instead of `Vec` as the `IT` type.

src/types/quota.rs
urkle: @jonhoo I forgot to add a reply about that.  I TRIED to add a Deref impl, but it refused to work due to needing a lifetime specifier that couldn't resolve against the self referencing object. 

So, we have a few options
1. remove the 'quota' method but leave the other two
  - I don't like this approach as it means that the object isn't a "quota" and makes writing code that consumes a quota harder to build out.
  - We could build a trait to represent the two methods Quota has and impl that on both objects. but unsure how you feel about that.
2. fight more with the Deref impl..
  - but I have doubts that i'll work easily w/o going into unsafe {} due to the hackery in ouroboros
  - this should still allow consuming methods to work against a Quota if we can get it to work.
3. remove the two helper methods and leave JUST the quota() method.
  - This would more easily provide the access to the inner Quota object to make writing consuming methods nicer without having to add a trait

src/types/quota.rs
urkle: Once we get everything resolved over in the ACL PR, I'll update this one to reflect the final decisions there.

src/types/quota.rs
jonhoo: Ah, interesting, yeah, I can see how that might get awkward. Hmm hmm hmm. I think I'm probably in favor of option 3 even though it's not very ergonomic.

src/client.rs
wookietreiber: Thanks for the explanation @jonhoo. Will see to integrate this in https://rust-unofficial.github.io/patterns/

src/types/quota.rs
```diff
@@ -0,0 +1,226 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{} {}", self.name, self.amount)
+    }
+}
+
+/// From [Resources](https://datatracker.ietf.org/doc/html/rfc2087#section-3)
+///
+/// Used by [`QuotaLimit`], and [`QuotaResource`]
+#[derive(Debug, Eq, PartialEq)]
+pub enum QuotaResourceName<'a> {
```
jonhoo, line 30: Let's set this to be `#[non_exhaustive]` in case we want to promote other extensions to proper variants (rather than them having to be wrapped in `Atom` for all eternity).

src/types/quota.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// The quota root name
+    pub fn root_name(&self) -> &str {
+        &*self.root_name
+    }
+
+    /// The defined resources with their usage and limits (could be empty)
+    pub fn resources(&self) -> &[QuotaResource<'a>] {
+        &self.resources[..]
+    }
+}
```
jonhoo, line 113: I'm torn about whether to keep these accessors or whether we should just make the fields of `Quota` `pub` and mark it as `#[non_exhaustive]` the same way we've done with other such types (like [`Mailbox`](https://docs.rs/imap/3.0.0-alpha.7/imap/types/struct.Mailbox.html)). What do you think?

src/types/quota.rs
```diff
@@ -0,0 +1,226 @@
...
+                .collect(),
+        }
+    }
+}
+
+/// From [QUOTA Response](https://datatracker.ietf.org/doc/html/rfc2087#section-5.1)
+///
+/// The quota resource sub-pieces in a [`Quota`]
+#[derive(Debug, Eq, PartialEq)]
+pub struct QuotaResource<'a> {
```
jonhoo, line 140: Let's set this to be `#[non_exhaustive]` in case other extensions we add support for in the future might add more fields.

urkle:
@jonhoo once #227 is merged I'll rebase this and apply the same changes to this branch.

jonhoo:
@urkle #227 is now merged :tada:

src/types/quota.rs
```diff
@@ -0,0 +1,246 @@
...
+pub struct QuotaResourceLimit<'a> {
+    /// The resource type
+    pub name: QuotaResourceName<'a>,
+    /// The amount for that resource
+    pub amount: u64,
+}
+
+impl<'a> QuotaResourceLimit<'a> {
+    /// Creates a new [`QuotaResourceLimit`]
+    pub fn new(name: QuotaResourceName<'a>, amount: u64) -> Self {
```
jonhoo, line 24: How about taking `impl Into<QuotaResourceName<'a>>` here, and then implementing `From<str>` for `QuotaResourceName`?

src/types/quota.rs
```diff
@@ -0,0 +1,246 @@
...
+use imap_proto::Response;
+use ouroboros::self_referencing;
+use std::borrow::Cow;
+use std::fmt::{Debug, Display, Formatter};
+use std::sync::mpsc;
+
+/// From [SETQUOTA Resource limit](https://datatracker.ietf.org/doc/html/rfc2087#section-4.1)
+///
+/// Used by [`Session::set_quota`]
+#[derive(Debug, Eq, PartialEq)]
```
jonhoo, line 13: Is there a good reason not to derive `Clone` here?

src/types/quota.rs
```diff
@@ -0,0 +1,246 @@
...
+        write!(f, "{} {}", self.name, self.amount)
+    }
+}
+
+/// From [Resources](https://datatracker.ietf.org/doc/html/rfc2087#section-3)
+///
+/// Used by [`QuotaLimit`], and [`QuotaResource`]
+#[derive(Debug, Eq, PartialEq, Clone)]
+#[non_exhaustive]
+pub enum QuotaResourceName<'a> {
```
jonhoo, line 40: I wonder if we'll want to have all of these "raw" types be in their own module, at least in the public API — what do you think?

src/types/quota.rs
```diff
@@ -0,0 +1,246 @@
...
+                })
+                .collect(),
+        }
+    }
+}
+
+/// From [QUOTA Response](https://datatracker.ietf.org/doc/html/rfc2087#section-5.1)
+///
+/// The quota resource sub-pieces in a [`Quota`]
+#[derive(Debug, Eq, PartialEq)]
```
jonhoo, line 144: Here, too, I think we can derive `Clone`

src/types/quota.rs
```diff
@@ -0,0 +1,246 @@
...
+pub struct QuotaResponse {
+    data: Vec<u8>,
+    #[borrows(data)]
+    #[covariant]
+    pub(crate) quota: Option<Quota<'this>>,
+}
+
+impl QuotaResponse {
+    /// Parse the [`Quota`] response from a response buffer.
+    pub fn parse(
```
jonhoo, line 85: Does the parser need to be `pub`? Same for the other `fn parse` in here.

tests/imap_integration.rs
```diff
@@ -27,7 +27,6 @@ fn test_smtp_host() -> String {
         .unwrap_or_else(|_| std::env::var("TEST_HOST").unwrap_or("127.0.0.1".to_string()))
 }
 
-#[cfg(feature = "test-full-imap")]
```
jonhoo, line 30: Was this intentional?

tests/imap_integration.rs
```diff
@@ -82,10 +84,37 @@ fn session(user: &str) -> imap::Session<native_tls::TlsStream<TcpStream>> {
...
+fn get_greeting() -> String {
+    let host = test_host();
+    let tcp = TcpStream::connect((host.as_ref(), test_imap_port())).unwrap();
+    let mut client = imap::Client::new(tcp);
+    let greeting = client.read_greeting().unwrap();
+    String::from_utf8(greeting).unwrap()
+}
+
+fn delete_mailbox(s: &mut imap::Session<native_tls::TlsStream<TcpStream>>, mailbox: &str) {
+    // we are silently eating any error (e.g. mailbox does not exist
```
jonhoo, line 102: nit:
```suggestion
    // we are silently eating any error (e.g. mailbox does not exist)
```

tests/imap_integration.rs
```diff
@@ -670,3 +699,83 @@ fn qresync() {
...
+    if is_greenmail {
+        let mut c = session(to);
+
+        let quota_root = c.get_quota_root("INBOX").unwrap();
+
+        assert_eq!(quota_root.mailbox_name(), "INBOX");
+
+        let root_names = quota_root.quota_root_names().collect::<Vec<&str>>();
+        assert_eq!(root_names, Vec::<&str>::new());
+        // TODO build tests for greenmail
```
jonhoo, line 734: What do you want to do about this TODO?

tests/imap_integration.rs
```diff
@@ -670,3 +699,83 @@ fn qresync() {
...
+    } else {
+        // because we are cyrus we can "test" the admin account for checking the GET/SET commands
+        let mut admin = session_with_options("cyrus", false);
```
jonhoo, line 737: Could you add a comment about why we specifically need to _not_ `clean_mailbox` for the admin session?

tests/imap_integration.rs
```diff
@@ -670,3 +699,83 @@ fn qresync() {
...
+        let update = admin
+            .set_quota(
+                "user/inbox-quota@localhost",
+                &[QuotaResourceLimit::new(QuotaResourceName::Storage, 500)],
+            )
+            .unwrap();
+        // Cyrus does not return the quota root definition on modification
+        assert_eq!(update.parsed(), &None);
+
+        let quota_response = admin.get_quota("user/inbox-quota@localhost").unwrap();
```
jonhoo, line 770: Why is this get-quota command run through admin?

jonhoo (changes requested):
Thanks for rebasing! Left another batch of notes.

src/client.rs
```diff
@@ -1382,6 +1376,48 @@ impl<T: Read + Write> Session<T> {
...
+    pub fn set_quota(
+        &mut self,
+        quota_root: impl AsRef<str>,
+        limits: &[QuotaResourceLimit<'_>],
+    ) -> Result<QuotaResponse> {
+        let limits = iter_join(limits.iter(), " ");
+        self.run_command_and_read_response(&format!(
+            "SETQUOTA {} ({})",
+            validate_str("SETQUOTA", "quota_root", quota_root.as_ref())?,
+            limits,
```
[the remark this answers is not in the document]
urkle: the limits are a structured vec (with a Display impl). and each limit is a specific token atom with an unsigned.  So there really isn't really any validation needed over what the imap-proto provides.

src/parse.rs
```diff
@@ -42,6 +42,74 @@ pub(crate) fn parse_many_into<'input, T, F>(
...
+        MapOrNot::Ignore => Ok(MapOrNot2::Ignore),
+    })
+}
+
+/// Parse and return an expected single `T` Response with `F`.
+/// Responses other than `T` go into the `unsolicited` channel.
+///
+/// If more than one `T` are found then [`Error::Parse`] is returned
+/// If zero `T` are found and optional is false then [`Error::Parse`] is returned, otherwise None is
+pub(crate) fn parse_until_done<'input, T, F>(
```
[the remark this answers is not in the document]
urkle: not honeslty sure why.  I'll move it back down.

src/types/quota.rs
urkle: I don't think so.  I'll derive it.  the QuotaResourceName is Cow internally when needed so clone will be "Cheap" unless a full clone is needed.

src/types/quota.rs
urkle: I only left it "pub" as all existing parsers were put. 

Though having it "pub" was actually helpful in writing tests for my app that uses this crate as I could easily write mocks to build the responses of the imap methods. (if they were not pub there would be no way for me to do that).

I have a mock_double for the imap connection so I can test my code and stub the responses (e.g. having my tests not rely on having an actual IMAP server there).

IMHO we can address this in a separate PR e.g. building a "testing module" or something to allow easy building of those ouroboros response objects.

tests/imap_integration.rs
urkle: read more up on greenmail and how it does quotas :)  I'll see if I can write some meaningful tests here for greenmail.

urkle:
@jonhoo fixups pushed and your comments have been either fixed or replied to (For further discussion)

src/types/quota.rs
jonhoo: I'm happy with it happening in a separate PR — may be that it can be done for other type modules too.

src/types/quota.rs
jonhoo: A separate PR sounds like a good idea. I do think that in general the parsing methods should probably not be public, or at the very least be in their own module.

jonhoo (changes requested):
We're nearly there now!

tests/imap_integration.rs
```diff
@@ -725,15 +725,35 @@ fn quota() {
...
+        c.set_quota("INBOX", &[QuotaResourceLimit::new("STORAGE", 1000)])
+            .unwrap();
+
+        // Check it
         let quota_root = c.get_quota_root("INBOX").unwrap();
 
         assert_eq!(quota_root.mailbox_name(), "INBOX");
 
         let root_names = quota_root.quota_root_names().collect::<Vec<&str>>();
+        // not sure why, but greenmail returns no quota root names
```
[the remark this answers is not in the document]
urkle: Logged.  https://github.com/greenmail-mail-test/greenmail/issues/497

src/client.rs
```diff
@@ -2376,6 +2423,278 @@ mod tests {
...
+        let quota = quota.parsed().as_ref().unwrap();
+        assert_eq!(quota.root_name, "my_root");
+        assert_eq!(quota.resources, vec![]);
+    }
+
+    #[test]
+    fn get_quota_with_limits() {
+        let response = b"* QUOTA my_root (STORAGE 10 500)\r\n\
+                a1 OK completed\r\n"
+            .to_vec();
```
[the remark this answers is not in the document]
urkle: @jonhoo I thought I put a response to this.

So in attempting to do this change it looks kinda funny with rustfmt.

Without the \n\ it formats like this.

```
        let response = b"* QUOTA my_root (STORAGE 10 500)\r
a1 OK completed\r
"
        .to_vec();

```

Which I'm not sure is really anymore readable that what was there. (other than having less leading white-space). Also I was matching style with all the est of the tests you have which use "\r\n\r"

We can do this change globally in a followup PR if you wish.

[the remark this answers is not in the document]
jonhoo: Oh, interesting. I think if you put `"\` on the first line (and then a real newline before the `*`), it might look nicer, but yeah, sounds like all of these aren't great options :+1:

jonhoo (approved):
All right, let's land this thing! Thanks for all your follow-up work!

jonhoo:
Oops, yes, completely forgot about those! My bad. Oh well, not the end of the world 😅

PR #236, opened 2022-07-26
update dependencies

urkle:
- updated lettre
  - required some changes to the imap_integration to support
- update hostname
- update minimum regex to avoid potential security issue

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/236)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ default = ["native-tls"]
 [dependencies]
 native-tls = { version = "0.2.2", optional = true }
 rustls-connector = { version = "0.16.1", optional = true }
-regex = "1.0"
+regex = "1.5"
 bufstream = "0.1.3"
 imap-proto = "0.16.0"
 nom = { version = "7.1.0", default-features = false }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -30,14 +30,13 @@ lazy_static = "1.4"
...
-lettre = "0.9.2"
-lettre_email = "0.9.2"
+lettre = "0.10"
 rustls-connector = "0.16.1"
 structopt = "0.3"
 
 # to make -Zminimal-versions work
 encoding = "0.2.32"
-hostname = "0.1.3"
+hostname = "0.3.1"
```
jonhoo, line 39: This specifier was _only_ there to make `-Zminimal-versions`, so if bumping the major version here is now okay, it can probably be removed entirely!

Cargo.toml
```diff
@@ -20,7 +20,7 @@ default = ["native-tls"]
 [dependencies]
 native-tls = { version = "0.2.2", optional = true }
 rustls-connector = { version = "0.16.1", optional = true }
-regex = "1.0"
+regex = "1.5"
```
jonhoo, line 23: This feels unnecessary. We do not depend on anything from 1.5 (to my knowledge). We should leave bumping for security fixes and the like to anything that actually builds binaries and locks dependencies, which we do not.

urkle: fair enough..  This was just based on the dependency checker you have linked in the README ( https://deps.rs/repo/github/jonhoo/rust-imap ). So to make that happy we should do `>= 1.5.5` OR just do `1.6`. But if you do not care then I'll revert this line.

Cargo.toml
urkle: removed.

Cargo.toml
jonhoo: Yeah, I'm generally opposed to libraries dictating a higher minimum version than they technically require, and think deps.rs is over-zealous here.

jonhoo (approved):

PR #237, opened 2022-08-02
README: fix Greenmail version in example command

bitfehler:
The instructions on running the integration tests use an outdated
version of Greenmail with which the tests will not pass. Update it to
the latest version, like in the pull command above it.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/237)


jonhoo (approved):
Thanks!

PR #243, opened 2022-09-19
Native SASL support

dequbed:
Hi! ^^

This PR adds native SASL support using rsasl. I've been testing usability of the rsasl API by patching existing crates to use rsasl and this PR is basically this testing code for `rust-imap`.

So in essence this is an issue asking if this is someting that you would be interested in merging, but already coming with some code attached ^^  

## Why would we want this?

While with the current `Authenticator` system most of SASL can be implemented there are a few advantages in using an external crate.

For one, features like SASL security layers and [channel bindings](https://www.rfc-editor.org/rfc/rfc5056.html) aren't really doable with the current approach. The latter specifically is something that would want explicit support by this crate as the strongest channel binding — `tls-exporter` — needs access to the TLS state otherwise managed by `rust-imap`.
Mechanism selection is another thing that could be implemented somewhat easier upstream as an user of the crate doesn't have access to `capability` without authenticating first.

Lastly it would allow us to drop the direct dependency on `base64` (although rsasl itself depends on the very same again) ^^

## What's the cost?

Of course this will result in an additional dependency, but I designed rsasl to make heavy use of feature unification to make it as cheap as possible when not used — when I checked for lettre it added 2.2KiB in (pre-LTO) binary size and less than a second in compile time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/243)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ bufstream = "0.1.3"
 imap-proto = "0.16.1"
 nom = { version = "7.1.0", default-features = false }
 base64 = "0.21"
+rsasl = { version = "2.0.0", default-features = false, features = ["std", "provider_base64"], optional = true }
 chrono = { version = "0.4", default-features = false, features = ["std"]}
 lazy_static = "1.4"
 ouroboros = "0.15.0"
@@ -41,6 +42,8 @@ encoding = "0.2.32"
 failure = "0.1.8"
 mime = "0.3.4"
 openssl = "0.10.35"
+
+rsasl = { version = "2.0.0", default-features = false, features = ["config_builder"] }
 
 [[example]]
 name = "basic"
@@ -69,3 +72,11 @@ required-features = ["default"]
 [[test]]
 name = "imap_integration"
 required-features = ["default"]
+
+[[example]]
+name = "rustls_sasl"
+required-features = ["rustls-tls", "rsasl", "rsasl/registry_static"]
+
+[[example]]
+name = "rustls_sasl_gss"
+required-features = ["rustls-tls", "rsasl", "rsasl/registry_static", "rsasl/gssapi"]
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,4 +1,3 @@
-use base64::{engine::general_purpose, Engine as _};
 use bufstream::BufStream;
 use chrono::{DateTime, FixedOffset};
 use imap_proto::Response;
@@ -7,6 +6,7 @@ use std::io::{Read, Write};
 use std::ops::{Deref, DerefMut};
 use std::str;
 use std::sync::mpsc;
+use base64::Engine;
 
 use super::authenticator::Authenticator;
 use super::error::{Bad, Bye, Error, No, ParseError, Result, ValidateError};
@@ -396,6 +396,24 @@ impl<T: Read + Write> Client<T> {
             .ok_or_else(|| panic!("CAPABILITY call did not populate capability cache!"))
     }
 
+    /// The [`CAPABILITY` command](https://tools.ietf.org/html/rfc3501#section-6.1.1) requests a
+    /// listing of capabilities that the server supports.  The server will include "IMAP4rev1" as
+    /// one of the listed capabilities. See [`Capabilities`] for further details.
+    pub fn capabilities(&mut self) -> Result<&Capabilities> {
+        if self.capability_cache.is_none() {
+            // TODO: Figure out how to handle unsolicited messages here.
+            let (mut tx, _) = mpsc::channel();
+            let caps = self.run_command_and_read_response("CAPABILITY")
+                .and_then(|lines| Capabilities::parse(lines, &mut tx))?;
+            self.capability_cache = Some(caps);
+        }
+
+        self.capability_cache.as_ref()
+            // This path will not be hit; if the cache is not populated the above `if` will either
+            // populate it or return with an early error.
+            .ok_or_else(|| panic!("CAPABILITY call did not populate capability cache!"))
+    }
+
     /// Log in to the IMAP server. Upon success a [`Session`](struct.Session.html) instance is
     /// returned; on error the original `Client` instance is returned in addition to the error.
     /// This is because `login` takes ownership of `self`, so in order to try again (e.g. after
@@ -411,7 +429,7 @@ impl<T: Read + Write> Client<T> {
     /// match client.login("user", "pass") {
     ///     Ok(s) => {
     ///         // you are successfully authenticated!
-    ///     },
+    ///     }
     ///     Err((e, orig_client)) => {
     ///         eprintln!("error logging in: {}", e);
     ///         // prompt user and try again with orig_client here
@@ -469,7 +487,7 @@ impl<T: Read + Write> Client<T> {
     ///     match client.authenticate("XOAUTH2", &auth) {
     ///         Ok(session) => {
     ///             // you are successfully authenticated!
-    ///         },
+    ///         }
     ///         Err((e, orig_client)) => {
     ///             eprintln!("error authenticating: {}", e);
     ///             // prompt user and try again with orig_client here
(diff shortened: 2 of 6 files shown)


jonhoo:
Sorry for the slow reply!

This seems like a totally reasonable thing to add, though I'd want to add it behind a feature flag. I also think we could do it in an additive way. That is, keep the current `Authenticate`-based API, and then provide a SASL-authenticate method (e.g., `Client::sasl_auth`) when the SASL feature is enabled that does auth through SASL. `Authenticate` may still be useful on its own for users who are working with relatively simple (or experimental) mechanisms, such as during testing.

One thing that surprised me when glancing over the changes was the need for `Arc`-wrapping the config — can that be avoided? Why is that required by the `SASLClient` API?

jonhoo:
Oh, also, we'd want to make sure by the time this lands that we can take a regular crates.io dependency on `rsasl` rather than a git dependency, so that we can in turn continue to publish imap on crates.io.

dequbed:
> Sorry for the slow reply!

Hey no worries, we're all doing unpaid volunteer work here — I don't mind waiting a few days, your README already states you don't have much time for this project. :)

> This seems like a totally reasonable thing to add, though I'd want to add it behind a feature flag. I also think we could do it in an additive way. That is, keep the current `Authenticate`-based API, and then provide a SASL-authenticate method (e.g., `Client::sasl_auth`) when the SASL feature is enabled that does auth through SASL. `Authenticate` may still be useful on its own for users who are working with relatively simple (or experimental) mechanisms, such as during testing.

Sure thing! I'd like to note that I designed rsasl specifically to be extensible so this would still allow experimental / testing mechanisms to be used, just like with `Authenticate`, without further changes to rsasl or rust-imap being required. The main reason I removed that method was really ~~because I like deleting code~~ since it means less code this library has to worry about and maintain, but I'm very much not set on that.

> One thing that surprised me when glancing over the changes was the need for `Arc`-wrapping the config — can that be avoided? Why is that required by the `SASLClient` API?

The main reason for this is that `SASLConfig` is an unsized type, specifically a trait object. Since the config comes from user code and not from either rsasl or libraries using rsasl, having the config be a generic instead of a trait object would quickly lead to a generic explosion as every code path handling SASL would have to be generic over a type they never interact with directly. `Arc` specifically was chosen because I wanted to have one config for both server- and client-side and especially on the former a global config being used by multiple threads simultaniously is quite likely. So of the typical DST containers `Box`, `Rc` and `Arc` I considered Arc to be the best option, and it is not much more expensive than Box when used only by one thread.

That all being said, I am debating making SASLConfig (and related) generic types in a future version of rsasl, especially now that GATs become an option. However, that's somewhat further in the future and I do think it can be done in a backwards-compatible manner.


> Oh, also, we'd want to make sure by the time this lands that we can take a regular crates.io dependency on `rsasl` rather than a git dependency, so that we can in turn continue to publish imap on crates.io.

Oh yeah of course! I was only using the git dependency as I was trying out a changed API that I did not yet release, as I said I am doing this PR partly to make sure that my API makes sense — and I did in fact find an issue ^^

dequbed:
I pushed a new version of this PR with the changes you requested so far.
There are still a few to-dos I'd like to implement in rust-imap for the rsasl feature, but most of them can be added in a future PR too if we want to get this feature into a release sooner.

src/client.rs
```diff
@@ -499,6 +502,100 @@ impl<T: Read + Write> Client<T> {
...
+        // TODO: check for `SASL-IR` capability and send initial data if it's supported.
+        ok_or_unauth_client_err!(
+            self.run_command(&format!("AUTHENTICATE {}", mechanism.as_str())),
+            self
+        );
+        self.do_sasl_handshake(session)
+    }
+
+    /// This func does the SASL handshake process once the authenticate command is made.
+    fn do_sasl_handshake(
```
jonhoo, line 531: I wonder if we can pull out some of the code here into a helper function that can be used both by the existing `authenticate` and by the new SASL functionality. After all, they're doing _mostly_ the same thing.

src/client.rs
```diff
@@ -499,6 +502,100 @@ impl<T: Read + Write> Client<T> {
...
+#[cfg(feature = "rsasl")]
+impl<T: Read + Write> Client<T> {
+
+    /// Authenticate with the server using the given custom SASLConfig to handle the server's
+    /// challenge.
+    ///
+    pub fn sasl_auth(
+        mut self,
+        config: ::std::sync::Arc<SASLConfig>,
+        mechanism: &Mechname,
```
jonhoo, line 514: It's a little weird to me to take `&Mechname` here, for a couple of reasons.

First, it seems like `rsals` actually wants `&[Mechname]`, so shouldn't we take that?
Second, if it's possible to _not_ provide an initial mechanism, should we make that possible?
And third, if `Mechname` is `Copy` (which its name makes it sound like it is), why the `&`?

Cargo.toml
```diff
@@ -26,7 +26,8 @@ regex = "1.0"
 bufstream = "0.1.3"
 imap-proto = "0.16.1"
 nom = { version = "7.1.0", default-features = false }
-base64 = "0.13"
+rsasl = { version = "2.0.0-rc.2", default-features = false, features = ["std", "provider_base64"], optional = true }
+base64 = "0.13.0"
```
jonhoo, line 30: I don't think you need to modify `base64` here now?

Cargo.toml
```diff
@@ -41,6 +42,8 @@ encoding = "0.2.32"
 failure = "0.1.8"
 mime = "0.3.4"
 
+rsasl = { version = "2.0.0-rc.2", default-features = false, features = ["config_builder"] }
```
jonhoo, line 45: nit: Why the extra newline above this?

src/client.rs
```diff
@@ -499,6 +502,100 @@ impl<T: Read + Write> Client<T> {
     }
 }
 
+#[cfg(feature = "rsasl")]
+impl<T: Read + Write> Client<T> {
+
+    /// Authenticate with the server using the given custom SASLConfig to handle the server's
+    /// challenge.
+    ///
+    pub fn sasl_auth(
```
jonhoo, line 511: How do you feel about also adding `#[deprecated]` to `fn authenticate` and point people at this instead?

jonhoo (commented):
I generally like this a lot! Left another few notes.

dequbed:
Hi!

I was linked your stream VoD recently and I noticed that I could have explained myself a bit better, especially since most people won't have deep familarity with SASL, rsasl, and all that surrounds it ^^'

So to answer some questions you asked into the ether in that video:

A first thing; I'm currently working on an updated version of rsasl (hey, just like you with imap `3.0.0-alpha.*`! ;p) that is written entirely in Rust and doesn't depend on an external library like the previous version did. So the default docsrs link will not help you with most of the types as this is in essence a full rewrite. The version you'd want to look at is the `2.0.0-rc.*` (https://docs.rs/rsasl/2.0.0-rc.3/rsasl/index.html), and its API is exactly what I'm trying to test with these PRs ;)

"Channel Binding" in one sentence is taking some unmistakable information from an underlying secure transport (so e.g. TLS in the common IMAP case) and including it in the authentication exchange in a way that will make the authentication fail if the information doesn't match up. It only works for a limited number of SASL mechanisms but it can help detect some cases of active MitM that has intercepted the TLS stream using e.g. a faked certificate. It's a fringe feature as most servers don't offer the mechanisms required and even then Channel Binding support is somewhat brittle as it doesn't have good negotiation.

SASL authentication using e.g. client certificates or other contextual information is a specific SASL mechanism called `EXTERNAL`.

Also, the magic method of rsasl is [Session::step64](https://docs.rs/rsasl/2.0.0-rc.3/rsasl/prelude/struct.Session.html#method.step64), it takes base64 data, does the authentication magic part, and puts base64-encoded data into a provided writer ^^

Cargo.toml
dequbed: Whoops that was by mistake ^^'

Cargo.toml
dequbed: The dev-deps are somewhat grouped by functionality, I tried emulating that.

src/client.rs
dequbed: I think that's a good idea ^^

src/client.rs
jonhoo: I think it's actually entirely reasonable for `Client` to start auto-issuing a call to gather the server's capabilities at the start of every session and storing the result. I'd be happy to land that, though I think we should land it in a separate PR maybe? And then you could make use of those stored capabilities here and get rid of the `&[Mechname]` argument entirely! Or, I suppose, we could make it an `Option<&[Mechname]>` so that users _can_ override if they want, but can also say "no, just choose based on server capabilities".

jonhoo:
That's helpful, thank you!

dequbed:
EHLO!

With rsasl now being fully released as 2.0.0 I cleaned this PR up and squashed it down somewhat. It also depends on the commit in #255 which we said would potentially be better as an entirely separate PR.

jonhoo:
Thanks! I'll hold off on reviewing this until #255 lands :+1:

Issue #244, opened 2022-09-27
Add abstraction for connection to "hide away" that whether the connection is TCP, or TLS.


urkle:
Currently the way the data types are for rust-imap you must *know* the exact type of connection you will be using and which tls implementation you are going to use in order to define your types.

I would like to make a PR that abstracts that all away into an enum so that it is hidden from the user.

A user should be able to do something like this.

```rust
let mut builder = ClientBuilder::new(
            self.config.imap_host.as_ref().unwrap(),
            self.config.imap_port,
        );
if use_starttls {
   // tag that we want start tls
   builder.starttls();
} else if use_ssl {
   // tag that we are doing FULL ssl mode
   builder.ssl();
}
// otherwise we are doing RAW tcp (still valid as it is useful for internal only connections)

// connects and returns an opaque ImapClient that abstracts away the Tcp stream type)
let client : ImapClient = builder.connect()?;
// connects and returns an opaque ImapSession that abstracts away the Tcp stream type)
let session : ImapSession = client.login("user", "pass")?;
```

Other crates do this and it greatly simplifies client-code as I don't have to know how I compiled the code in order to connect. And I can choose to conditionally connect via SSL / TLS / raw.

One example is ldap3, which uses a simple ENUM to differentiate between Tcp, Tls, and Unix sockets (imap would only need the 2).

Let me know if you are open to this and I can build up a PR.

jonhoo:
I'll follow up in the PR :+1:

PR #245, opened 2022-10-06
add in a client builder that abstracts away connecting to TLS or non-TLS connections and what TLS provider is used.

urkle:
- this allows a more transparent and versatile usage of the library as one can simply compile it as-is and then use the builder to configure where we connect and how we connect without having to be concerned about what type is used for the imap::Client / imap::Session

This is inspired by how the ldap3 library hides-away the connection of ldap, ldaps, or ldapi (unix socket)

IMHO this should just be the main ClientBuilder, however, the methodology used in this PR requires that only either rustls OR nativetls be active, not both. So this PR adds it as a parallel client builder.

Implements issue described in #244 

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/245)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Below is a basic client example. See the `examples/` directory for more.
 ```rust
 fn fetch_inbox_top() -> imap::error::Result<Option<String>> {
 
-    let client = imap::ClientBuilder::new("imap.example.com", 993).native_tls()?;
+    let client = imap::ClientBuilder::new("imap.example.com", 993).connect()?;
 
     // the client we have here is unauthenticated.
     // to do anything useful with the e-mails, we need to log in
--- a/src/client.rs
+++ b/src/client.rs
@@ -340,9 +340,21 @@ impl<T: Read + Write> Client<T> {
     ///
     /// This consumes `self` since the Client is not much use without
     /// an underlying transport.
-    pub(crate) fn into_inner(self) -> Result<T> {
+    pub fn into_inner(self) -> Result<T> {
         let res = self.conn.stream.into_inner()?;
         Ok(res)
+    }
+
+    /// The [`CAPABILITY` command](https://tools.ietf.org/html/rfc3501#section-6.1.1) requests a
+    /// listing of capabilities that the server supports.  The server will include "IMAP4rev1" as
+    /// one of the listed capabilities. See [`Capabilities`] for further details.
+    ///
+    /// This allows reading capabilities before authentication.
+    pub fn capabilities(&mut self) -> Result<Capabilities> {
+        // Create a temporary channel as we do not care about out of band responses before login
+        let (mut tx, _rx) = mpsc::channel();
+        self.run_command_and_read_response("CAPABILITY")
+            .and_then(|lines| Capabilities::parse(lines, &mut tx))
     }
 
     /// Log in to the IMAP server. Upon success a [`Session`](struct.Session.html) instance is
@@ -355,7 +367,7 @@ impl<T: Read + Write> Client<T> {
     /// # {} #[cfg(feature = "native-tls")]
     /// # fn main() {
     /// let client = imap::ClientBuilder::new("imap.example.org", 993)
-    ///     .native_tls().unwrap();
+    ///     .connect().unwrap();
     ///
     /// match client.login("user", "pass") {
     ///     Ok(s) => {
@@ -412,7 +424,7 @@ impl<T: Read + Write> Client<T> {
     ///         user: String::from("me@example.com"),
     ///         access_token: String::from("<access_token>"),
     ///     };
-    ///     let client = imap::ClientBuilder::new("imap.example.com", 993).native_tls()
+    ///     let client = imap::ClientBuilder::new("imap.example.com", 993).connect()
     ///         .expect("Could not connect to server");
     ///
     ///     match client.authenticate("XOAUTH2", &auth) {
@@ -1819,6 +1831,31 @@ mod tests {
             session.stream.get_ref().written_buf == command.as_bytes().to_vec(),
             "Invalid authenticate command"
         );
+    }
+
+    #[test]
+    fn pre_login_capability() {
+        let response = b"* CAPABILITY IMAP4rev1 STARTTLS AUTH=GSSAPI LOGINDISABLED\r\n\
+            a1 OK CAPABILITY completed\r\n"
+            .to_vec();
+        let expected_capabilities = vec![
+            Capability::Imap4rev1,
+            Capability::Atom(Cow::Borrowed("STARTTLS")),
+            Capability::Auth(Cow::Borrowed("GSSAPI")),
+            Capability::Atom(Cow::Borrowed("LOGINDISABLED")),
+        ];
+        let mock_stream = MockStream::new(response);
+        let mut client = Client::new(mock_stream);
+        let capabilities = client.capabilities().unwrap();
+        assert_eq!(
+            client.stream.get_ref().written_buf,
+            b"a1 CAPABILITY\r\n".to_vec(),
+            "Invalid capability command"
+        );
+        assert_eq!(capabilities.len(), 4);
+        for e in expected_capabilities {
+            assert!(capabilities.has(&e));
+        }
     }
 
     #[test]
--- a/src/client_builder.rs
+++ b/src/client_builder.rs
@@ -1,55 +1,159 @@
-use crate::{Client, Result};
+use crate::{Client, Connection, Error, Result};
+
+use lazy_static::lazy_static;
 use std::io::{Read, Write};
 use std::net::TcpStream;
 
 #[cfg(feature = "native-tls")]
-use native_tls::{TlsConnector, TlsStream};
+use native_tls::TlsConnector as NativeTlsConnector;
+
+use crate::extensions::idle::SetReadTimeout;
 #[cfg(feature = "rustls-tls")]
-use rustls_connector::{RustlsConnector, TlsStream as RustlsStream};
+use rustls_connector::{
+    rustls,
+    rustls::{Certificate, ClientConfig, RootCertStore, ServerName},
+    rustls_native_certs::load_native_certs,
+    RustlsConnector,
+};
+#[cfg(feature = "rustls-tls")]
+use std::sync::Arc;
+
+#[cfg(feature = "rustls-tls")]
+struct NoCertVerification;
+
+#[cfg(feature = "rustls-tls")]
+impl rustls::client::ServerCertVerifier for NoCertVerification {
+    fn verify_server_cert(
+        &self,
+        _: &Certificate,
+        _: &[Certificate],
+        _: &ServerName,
+        _: &mut dyn Iterator<Item = &[u8]>,
+        _: &[u8],
+        _: std::time::SystemTime,
+    ) -> std::result::Result<rustls::client::ServerCertVerified, rustls::Error> {
+        Ok(rustls::client::ServerCertVerified::assertion())
+    }
+}
+
+#[cfg(feature = "rustls-tls")]
+lazy_static! {
+    static ref CACERTS: RootCertStore = {
+        let mut store = RootCertStore::empty();
+        for cert in load_native_certs().unwrap_or_else(|_| vec![]) {
+            if let Ok(_) = store.add(&Certificate(cert.0)) {}
+        }
+        store
+    };
+}
+
+lazy_static! {
+    static ref STARTLS_CHECK_REGEX: regex::bytes::Regex =
+        regex::bytes::Regex::new(r"\bSTARTTLS\b").unwrap();
+}
+
+/// The connection mode we are going to use
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum ConnectionMode {
+    /// Automatically detect what connection mode should be used.
+    ///
+    /// This will use TLS if the port is 993, and otherwise STARTTLS if available.
+    /// If no TLS communication mechanism is available, the connection will fail.
+    AutoTls,
+    /// Automatically detect what connection mode should be used.
+    ///
+    /// This will use TLS if the port is 993, and otherwise STARTTLS if available.
+    /// It will fallback to a plaintext connection if no TLS option can be used.
+    Auto,
+    /// A plain unencrypted TCP connection
+    Plaintext,
+    /// An encrypted TLS connection
+    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+    Tls,
+    /// An eventually-encrypted (i.e., STARTTLS) connection
+    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+    StartTls,
+}
+
+/// A selection for TLS implementation
+#[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum TlsKind {
+    /// Use the NativeTLS backend
+    #[cfg(feature = "native-tls")]
+    Native,
+    /// Use the Rustls backend
+    #[cfg(feature = "rustls-tls")]
+    Rust,
+    /// Use whatever backend is available (uses rustls if both are available)
+    Any,
+}
 
 /// A convenience builder for [`Client`] structs over various encrypted transports.
 ///
-/// Creating a [`Client`] using `native-tls` transport is straightforward:
+/// Creating a [`Client`] using TLS is straightforward.
+///
+/// This will make a TLS connection directly since the port is 993.
 /// ```no_run
 /// # use imap::ClientBuilder;
 /// # {} #[cfg(feature = "native-tls")]
 /// # fn main() -> Result<(), imap::Error> {
-/// let client = ClientBuilder::new("imap.example.com", 993).native_tls()?;
+/// let client = ClientBuilder::new("imap.example.com", 993).connect()?;
 /// # Ok(())
 /// # }
 /// ```
 ///
-/// Similarly, if using the `rustls-tls` feature you can create a [`Client`] using rustls:
+/// By default it will detect and use `STARTTLS` if available.
 /// ```no_run
 /// # use imap::ClientBuilder;
-/// # {} #[cfg(feature = "rustls-tls")]
+/// # {} #[cfg(feature = "native-tls")]
 /// # fn main() -> Result<(), imap::Error> {
-/// let client = ClientBuilder::new("imap.example.com", 993).rustls()?;
+/// let client = ClientBuilder::new("imap.example.com", 143).connect()?;
 /// # Ok(())
 /// # }
 /// ```
 ///
-/// To use `STARTTLS`, just call `starttls()` before one of the [`Client`]-yielding
-/// functions:
+/// To force a certain implementation you can call tls_kind():
 /// ```no_run
 /// # use imap::ClientBuilder;
 /// # {} #[cfg(feature = "rustls-tls")]
 /// # fn main() -> Result<(), imap::Error> {
 /// let client = ClientBuilder::new("imap.example.com", 993)
-///     .starttls()
-///     .rustls()?;
+///     .tls_kind(imap::TlsKind::Rust).connect()?;
+/// # Ok(())
+/// # }
+/// ```
+///
+/// To force the use `STARTTLS`, just call `mode()` before connect():
+///
+/// If the server does not provide STARTTLS this will error out.
+/// ```no_run
+/// # use imap::ClientBuilder;
+/// # {} #[cfg(feature = "rustls-tls")]
+/// # fn main() -> Result<(), imap::Error> {
+/// use imap::ConnectionMode;
+/// let client = ClientBuilder::new("imap.example.com", 993)
+///     .mode(ConnectionMode::StartTls)
+///     .connect()?;
 /// # Ok(())
 /// # }
 /// ```
 /// The returned [`Client`] is unauthenticated; to access session-related methods (through
 /// [`Session`](crate::Session)), use [`Client::login`] or [`Client::authenticate`].
+#[derive(Clone)]
 pub struct ClientBuilder<D>
 where
     D: AsRef<str>,
 {
     domain: D,
     port: u16,
-    starttls: bool,
+    mode: ConnectionMode,
+    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+    tls_kind: TlsKind,
+    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+    skip_tls_verify: bool,
 }
 
 impl<D> ClientBuilder<D>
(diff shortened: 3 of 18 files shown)


tjni:
Disclaimer: I'm a rust novice and new to this crate.

I haven't reviewed in detail, but I really like that enabling only one of `native-tls` or `rustls-tls` uses code that does not need to specify which one is used.

Personally, I think this is the overwhelming common case and would vote to make this part of the current builder even if it removes the option of using both TLS features at the same time, unless someone has a need for it.

If we want to support that situation, would it still be possible by requiring a TLS library enum as a parameter to `tls` when multiple TLS library features are enabled?

src/conn.rs
```diff
@@ -0,0 +1,59 @@
...
+#[cfg(feature = "native-tls")]
+use native_tls::TlsStream;
+#[cfg(all(not(doc), feature = "rustls-tls"))]
+use rustls_connector::TlsStream;
+use std::io::{Read, Write};
+use std::net::TcpStream;
+
+/// A generic stream wrapper around either a Tcp or a Tls connection
+#[derive(Debug)]
+pub enum ConnType {
```
jonhoo, line 12: Couldn't we just replace this whole thing with
```rust
pub trait ImapConnection: Read + Write {}
impl<T> ImapConnection for T where T: Read + Write {}

pub type ConnType = Box<dyn ImapConnection>;
```
I think that'd be much nicer, and it should also allow the two TLS implementations to co-exist rather than being mutually exclusive.

src/boxed_client_builder.rs
```diff
@@ -0,0 +1,216 @@
...
+    #[cfg(feature = "native-tls")]
+    fn build_tls_connection(&self, tcp: TcpStream) -> Result<ConnType> {
+        let mut builder = TlsConnector::builder();
+        if self.no_tls_verify {
+            builder.danger_accept_invalid_certs(true);
+            builder.danger_accept_invalid_hostnames(true);
+        }
+        let ssl_conn = builder.build()?;
+        Ok(TlsConnector::connect(&ssl_conn, self.domain.as_ref(), tcp)?.into())
+    }
```
jonhoo, line 215: The way I'd recommend we do this instead is to have something like:
```rust
#[non_exhaustive]
pub enum TlsKind {
  #[cfg(feature = "native-tls")]
  Native
  #[cfg(feature = "rustls-tls")]
  Rust
  #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
  Any
}
```
with
```rust
impl Builder {
  pub fn tls_kind(mut self, kind: TlsKind) -> Self { ... }
  fn build_tls_connection(&self, tcp: TcpStream) -> Result<ConnType> {
    match self.kind {
      #[cfg(feature = "native-tls")]
      TlsKind::Native => self.build_tls_native(),
      #[cfg(feature = "rustls-tls")]
      Rust => self.build_tls_rustls(),
      #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
      Any => self.build_tls_rustls(), // <- we get to choose the default here
    }
  }
}
```

src/conn.rs
jonhoo: Furthermore, by doing this, we should be able to make this just be part of the standard builder.

jonhoo (commented):
Two high-level comments about the design. Overall, I'm definitely a fan of landing something like this!

urkle:
@jonhoo I did the first part of the refactoring that uses your Boxed trait suggestion.  Now the code works with both native and rust enabled.

I'll look next at condensing the builders so we have a single builder

urkle:
@jonhoo I added another commit that combines the boxed client builder into the now single client builder.

Note a few contract changes happened.

1) the `native_tls()` and `rustls()` methods have been removed as tls_kind replaces that functionality.
2) the old `connect(FnOne)` method has been renamed `connect_tls_with(FnOnce)` I could not find a nice way to get it to behave and convert the `C` to a  `Box<dyn ImapConnection>` the compiler just yelled at me, so it still returns a `Client<C>`
   - if we can find a generic way to allow the user to set custom certs we may be able to just do away with this method completely :)

.github/workflows/check.yml
```diff
@@ -82,7 +82,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: hack
-          args: --feature-powerset check --all-targets
+          args: check --feature-powerset --all-targets
```
jonhoo, line 85: Why this change? I though hack arguments had to come before the cargo subcommand?

README.md
```diff
@@ -34,7 +34,7 @@ Below is a basic client example. See the `examples/` directory for more.
 ```rust
 fn fetch_inbox_top() -> imap::error::Result<Option<String>> {
 
-    let client = imap::ClientBuilder::new("imap.example.com", 993).native_tls()?;
+    let client = imap::ClientBuilder::new("imap.example.com", 993).tls().connect()?;
```
jonhoo, line 37: I think we should keep the `native_tls` and `rustls` constructors for convenience.

src/client_builder.rs
```diff
@@ -61,35 +129,88 @@ where
...
             port,
-            starttls: false,
+            mode: ConnectionMode::Tcp,
+            tls_kind: TlsKind::Any,
+            no_tls_verify: false,
         }
     }
 
     /// Use [`STARTTLS`](https://tools.ietf.org/html/rfc2595) for this connection.
-    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
```
jonhoo, line 69: We should still require one of these features to be set, because otherwise `ConnectionMode::StartTls` won't work.

src/client_builder.rs
```diff
@@ -1,55 +1,123 @@
...
+
+/// The connection mode we are going to use
+#[derive(Clone, Debug, PartialEq, Eq)]
+enum ConnectionMode {
+    /// A plain unencrypted Tcp connection
+    Tcp,
+    /// an encrypted TLS connection
+    Tls,
+    /// a start tls connection
+    StartTls,
```
jonhoo, line 58: These should both have `#[cfg]`s.

src/client_builder.rs
```diff
@@ -61,35 +129,88 @@ where
...
+    /// # {} #[cfg(feature = "rustls-tls")]
+    /// # fn main() -> Result<(), imap::Error> {
+    /// let client = ClientBuilder::new("imap.example.com", 993)
+    ///     .starttls().connect()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn connect(&self) -> Result<Client<ConnType>> {
+        let tcp = TcpStream::connect((self.domain.as_ref(), self.port))?;
+        #[cfg(any(feature = "rustls-tls", feature = "native-tls"))]
```
jonhoo, line 185: I think the body of this function would look a lot nicer if structured as:
```rust
match self.mode {
```
with each arm that is only available with TLS carrying a `#[cfg]`.

src/client_builder.rs
```diff
@@ -61,35 +129,88 @@ where
...
+            } else {
+                tcp
+            };
+            self.build_tls_connection(tcp)?
+        };
+        #[cfg(all(not(feature = "rustls-tls"), not(feature = "native-tls")))]
+        let stream: ConnType = if self.mode == ConnectionMode::Tcp {
+            Box::new(tcp)
+        } else {
+            return Err(crate::Error::NoTlsSupport);
```
jonhoo, line 204: We should make it impossible for this error to occur.

src/client_builder.rs
```diff
@@ -1,55 +1,123 @@
...
+#[non_exhaustive]
+pub enum TlsKind {
+    /// Use the NativeTLS backend
+    #[cfg(feature = "native-tls")]
+    Native,
+    /// Use the Rustls backend
+    #[cfg(feature = "rustls-tls")]
+    Rust,
+    /// Use whatever backend is available (native over rustls)
+    Any,
```
jonhoo, line 72: This should have `#[cfg(any(...))]`.

README.md
jonhoo: Also, I think those constructors could utilize `connect_tls_with` to avoid the `Box<dyn>` (and thus get a slight performance boost).

src/client_builder.rs
```diff
@@ -116,19 +237,23 @@ where
...
-    ///     .connect(|domain, tcp| {
+    ///     .connect_tls_with(|domain, tcp| {
     ///         let ssl_conn = RustlsConnector::new_with_native_certs()?;
     ///         Ok(ssl_conn.connect(domain, tcp)?)
     ///     })?;
     /// # Ok(())
     /// # }
     /// ```
-    pub fn connect<F, C>(&mut self, handshake: F) -> Result<Client<C>>
+    pub fn connect_tls_with<F, C>(&mut self, handshake: F) -> Result<Client<C>>
```
jonhoo, line 247: Note that this isn't _necessarily_ for TLS. This can be used for _any_ customization of the underlying I/O channel. I think it should just be called `connect_with`.

src/client_builder.rs
```diff
@@ -116,19 +237,23 @@ where
...
     where
         F: FnOnce(&str, TcpStream) -> Result<C>,
         C: Read + Write,
     {
-        let tcp = if self.starttls {
+        if self.mode == ConnectionMode::Tcp {
+            panic!("Can only be called with a TLS or StartTLS connection");
+        }
```
jonhoo, line 254: This doesn't seem right to me. My expectation would actually be that if someone calls `connect_with` and hasn't called `tls()`, then this should work just fine and just use exactly what comes out of the `FnOnce` as the channel. This suggests, perhaps, that `Tcp` isn't the right name for this mode, and that it should instead be called something like `Direct` or `Plain` or `Unmodified`.

src/client_builder.rs
```diff
@@ -141,10 +266,62 @@ where
         let tls = handshake(self.domain.as_ref(), tcp)?;
 
         let mut client = Client::new(tls);
-        if !self.starttls {
+        if self.mode != ConnectionMode::StartTls {
```
jonhoo, line 269: I think now that we have `self.mode`, I would expect this to _also_ respect `ConnectionMode::Tls`. That is, if someone did:
```rust
builder.tls().connect_with(|_host, tcp| tcp)
```
I would expect them to end up with a TLS connection. This comes back to the fact that `connect_with` is for wrapping the channel _in any way_, and so does not in and of itself imply that TLS has already been set up. Which also then means that it'd be a mistake to call `.tls().connect_with(|_, tcp| Rustls::new(tcp))`, because you'd end up running TLS-over-TLS.

jonhoo: We should also document this behavior.

src/conn.rs
```diff
@@ -0,0 +1,18 @@
+use crate::extensions::idle::SetReadTimeout;
+
+use std::fmt::{Debug, Formatter};
+use std::io::{Read, Write};
+
+/// Imap connection trait of a read/write stream
+pub trait ImapConnection: Read + Write + SetReadTimeout {}
```
jonhoo, line 7: We should probably [seal](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed) this trait.

src/error.rs
```diff
@@ -104,6 +105,8 @@ pub enum Error {
     /// In response to a STATUS command, the server sent OK without actually sending any STATUS
     /// responses first.
     MissingStatusResponse,
+    /// No TLS Backend was compiled in and a TLS connection was attempted
+    NoTlsSupport,
```
jonhoo, line 109: I think we should make this impossible.

src/conn.rs
```diff
@@ -0,0 +1,18 @@
...
+impl<T> ImapConnection for T where T: Read + Write + SetReadTimeout {}
+
+impl Debug for dyn ImapConnection {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Imap connection")
+    }
+}
+
+/// A boxed connection type
+pub type ConnType = Box<dyn ImapConnection>;
```
jonhoo, line 18: I think a better name for this would be `Connection`.

jonhoo (changes requested):

.github/workflows/check.yml
urkle: All the docs for the cargo hack command show them being called in this order.. "check --feature-powerset etc...).

README.md
urkle: @jonhoo which is why I kinda like the "enum" approach that the Ldap crate used as it removed the need for the box.

And yes, I can add back in the nativeTls / rusttls methods.

src/client_builder.rs
urkle: Which I handled in the connection method by having it return an error staging that no TLS was configured.

Any the reason I removed this #[cfg] was that if I didn't this enumeration would have zero values when no TLS implementation was compiled in.   Leaving this in felt the simplest way to handle that nicely.

the other way would be to completely CFG out this entire enum and tlskind struct member and other functions. (which I can do instead if you prefer that rough)

src/client_builder.rs
urkle: Per your comment about the function name, I'm just removing this check completely.

src/client_builder.rs
urkle: Which that now greatly complicates this method's use.  As if you go with that logic how would StartTLS behave?  Would the user of this method have to initialize starttls? or would the function still perform that behavior?

As your original intent for this method was to "customize" the TLS setup, which one cannot do AFTER the TLS object has been created and wrapped around the raw TCP connection.  Thus this is intended to be done before wrapping.

So, perhaps we should just scrap this method entirely until we come up with a better way to customize the connections?  e.g. a customize_native / customize_rustls ?  that provides the config/builder struct allowing the user to tweak it ?

e.g.
```rust
client_builder.tls().customize_nativetls(|&mut builder| .... modify builder).connect()?;
```

and in our build_native_tls() internal method we do this.

```rust
fn build_tls_native(&self, tcp: TcpStream) -> Result<ConnType> {
        let mut builder = NativeTlsConnector::builder();
        if self.no_tls_verify {
            builder.danger_accept_invalid_certs(true);
            builder.danger_accept_invalid_hostnames(true);
        }
        if Some(customizer) = self.customize_nativetls_callback {
           customizer(builder);
        }
        let ssl_conn = builder.build()?;
        Ok(Box::new(NativeTlsConnector::connect(
            &ssl_conn,
            self.domain.as_ref(),
            tcp,
        )?))
    }
```

We can either have two separate customization hooks OR have a unified with that takes an enum of either the nativetls builder or the rusttls config.   Also if a consumer is customizing the TLS connection do we want to process the defaults always? (loading stock certs?) or add a config option to disable that functionality?

src/error.rs
urkle: it's removed now

src/client_builder.rs
jonhoo: Why is it a problem for the enum to have no variants?

I'd rather have things fail at compile-time (i.e., make it impossible to write code that tries to `STARTTLS` when there's no TLS available), rather than have it return a runtime error.

README.md
jonhoo: When you do, I think we can also avoid changing all these examples

src/client_builder.rs
```diff
@@ -61,38 +134,91 @@ where
...
-    #[cfg(feature = "rustls-tls")]
-    #[cfg_attr(docsrs, doc(cfg(feature = "rustls-tls")))]
-    pub fn rustls(&mut self) -> Result<Client<RustlsStream<TcpStream>>> {
-        self.connect(|domain, tcp| {
-            let ssl_conn = RustlsConnector::new_with_native_certs()?;
-            Ok(ssl_conn.connect(domain, tcp)?)
-        })
+    /// Sets the TLS backend to use for this connection.
+    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+    pub fn tls_kind(&mut self, kind: TlsKind) -> &mut Self {
```
jonhoo, line 160: I wonder if this should auto-set `self.mode = Tls` if it's currently set to `Tcp`. At least, it feels weird to have to call `tls_kind` and _also_ `tls`.

src/client_builder.rs
```diff
@@ -61,38 +134,91 @@ where
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn connect(&self) -> Result<Client<Connection>> {
+        let tcp = TcpStream::connect((self.domain.as_ref(), self.port))?;
+        let stream = match self.mode {
+            ConnectionMode::Tcp => Box::new(tcp),
+            #[cfg(any(feature = "rustls-tls", feature = "native-tls"))]
+            _ => {
+                let tcp = if self.mode == ConnectionMode::StartTls {
```
jonhoo, line 199: I think we should just accept a little bit of repetition here and avoid the `_` match followed by an `if` on the same variable. it'll make this easier to find should we ever add a new `mode` in the future, and it'll make the code a little nicer to read I think.

src/client_builder.rs
jonhoo: Hmm, I do see your point, but at the same I simultaneously think this method is useful as-is _and_ I think it should do something reasonable in the `client.tls().connect_with` situation. I suppose we could return a runtime error? An alternative would be to do a type-state kind of thing where instead of keeping `mode` as a variable we keep it as a type parameter, but that feels excessive. Maybe just document that it'll return an error if `tls()` is set?

src/client_builder.rs
urkle: I removed the entire type when no TLS backend is compiled in, and the tls/starttls methods are also removed (based on your later comments).  So the runtime error code is gone now.

src/client_builder.rs
urkle: What error should it return?  ( a new variant?)

jonhoo: Yeah, a new variant seems fine

README.md
urkle: well, so, I would think we'd want the examples to be in the preferable "non-surprising" syntax. (e.g. it doesn't break if they switch backends)  (which is why I opted for dropping those methods to begin with)

Also, if you are concerned about performance of the Box we can use an enum of 
```rust
enum Connection {
    Tcp(TcpStream),
    NativeTls(NativeTlsStream),
    RustTls(RusTlsStream),
    Custom(Box<ImapConnection>),
}
```

jonhoo: Hmm, maybe. I'm torn, because it's _also_ useful to show folks how they can choose a particular TLS backend to use. I'll go whichever way you prefer here :+1: 

No, `Box<dyn>` is generally going to be better performance than dispatch-through-enum I think.

urkle: We can add examples showing the variations, and those two methods (rustls, and native_tls) that I added back in have doc-comment examples showing usage.

src/client_builder.rs
urkle: Well, hopefully any "reasonable" IMAP server is properly configured to disallow non-TLS connections in that case :)  ( I know my public facing IMAP servers are configured this way)

So a few things, 
1) if a user did connect to port 993 without calling TLS() it wouldn't connect.
2) The issue with defaulting TLS on is you are assuming they are connecting to port 993?   (thus port 143 wouldn't work).
3) Should we perhaps maybe introspect the port # passed in and default to TLS if 993 is passed? and starttls if port 143 is passed?
4) Or maybe have it default to an "auto mode" where by
   a) if port = 993 then TLS on
   b) otherwise use starttls *IF* the server supports it (e.g. we would introspect the initially advertised capabilities in the greeting, which a server support STARTTLS should be advertising there)

Honestly #4 would provide the nicest use-case as it would be security by default (if possible). and provide the ease of use you are wanting to give in the crate API.  And uses who want to "go beyond" can simply call the explicit `tls()`, `starttls()`, and (new) `raw()` methods to change the connection mode explicitly.


For "connect_with", we'd still have to figure that out.  Maybe changing the contract to something like this.
`connect_with(starttls: bool, handshake: FnOnce)`
Thus we have to explicitly pass whether we are expecting to do starttls or not.  (false if we are doing RAW or full TLS). Thus allowing that the full flexibility, and it would essentially "ignore" the connection mode enum in the connection object.

jonhoo: I like doing number 4!

For `connect_with`, I guess I wonder why we'd want to have a separate `bool` that overrides whatever `starttls` is set to, rather than just respecting what it's set to directly?

src/client_builder.rs
```diff
@@ -61,38 +134,131 @@ where
...
+    /// ```
     #[cfg(feature = "native-tls")]
-    #[cfg_attr(docsrs, doc(cfg(feature = "native-tls")))]
-    pub fn native_tls(&mut self) -> Result<Client<TlsStream<TcpStream>>> {
-        self.connect(|domain, tcp| {
-            let ssl_conn = TlsConnector::builder().build()?;
-            Ok(TlsConnector::connect(&ssl_conn, domain, tcp)?)
-        })
+    pub fn native_tls(&mut self) -> Result<Client<Connection>> {
+        if self.mode == ConnectionMode::Tcp {
```
jonhoo, line 195: It still feels so weird to me that we force `.tls()` here, but _not_ for `.tls_kind()`.

src/client_builder.rs
```diff
@@ -141,10 +307,62 @@ where
...
+            #[cfg(feature = "rustls-tls")]
+            TlsKind::Rust => self.build_tls_rustls(tcp),
+            TlsKind::Any => self.build_tls_any(tcp),
+        }
+    }
+
+    #[cfg(feature = "native-tls")]
+    fn build_tls_any(&self, tcp: TcpStream) -> Result<Connection> {
+        self.build_tls_native(tcp)
+    }
```
jonhoo, line 331: How do you feel about making rustls the default instead of native-tls?

urkle: I have no real opinion (other than I HAD to use native_tls for the project I am on as rustls wouldn't connect to 'older' machines)

src/client_builder.rs
urkle: Not sure what you mean here.

src/client_builder.rs
urkle: So, I have "auto" mode implemented locally for the main builder.

Now onto connect_with

So, my thinking of why connect_with would take the starttls param is that it's a "manual" connection mode anyway thus it isn't using any of the "knowledge" from the builder other than the port and host.  (e.g. its even ignoring "tls_kind").

if I have it rely on the ConnectionMode instead then the question is what happens in each scenario.

1. Auto
    - if 993 then TLS (leave alone) otherwise connect and "detect" starttls. Though we really need to "tell" the FnOnce what it is receiving in this case (raw or TLS connection)
2. Tcp/Raw
    - just pass the connection through to the callback
3. StartTls
    - upgrade the TLS always with starttls and pass the upgraded connection to the callback
4. Tls
    - just pass the connection through the callback

Thus to make this really work cleanly we'd really need to change the contract of the callback to accept the socket and the mode so that it knows if it has a Tcp/Raw, or TLS-ready connection has been passed.  or do we ONLY call the callback when we have a TLS connnection?

This goes back to what are the use-cases of this method?

Is it to provide TLS customization? (certs, etc..). If so we can provide better methods to handle that (as I explained elsewhere).

Is it to provide full customization of the TCP connection in other ways?  If so, do we really need to provide that in the ClientBuilder and can't the developer just create a TCP connection themselves?

Is the handling of StartTls in this method making it more difficult?

Anyways,

I'm pushing up a commit that

1) adds an "auto"mode as discussed in point 4 above
2) reworks the connect_with to utilize the "Auto" mode.

I'm going to play around a little more later tonight and possible rewrite the `connect` method to use connect_with and make the code smaller.

src/client_builder.rs
jonhoo: `Client::tls_kind` does not set `self.mode = Tls`. But `Client::native_tls` (which really just changes the kind) _does_ set `self.mode = Tls`. That feels inconsistent.

src/client_builder.rs
```diff
@@ -61,41 +142,125 @@ where
...
+    /// [`rustls::ClientConfig::dangerous`]
+    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+    pub fn danger_no_tls_verify(&mut self, no_tls_verify: bool) -> &mut Self {
+        self.no_tls_verify = no_tls_verify;
         self
     }
 
-    /// Return a new [`Client`] using a `native-tls` transport.
+    /// Makes a [`Client`] connection using the native TLS backend.
+    /// Forces TLS unless [`ClientBuilder::starttls()`] was called
```
jonhoo, line 198: I _think_ rustdoc prefers the first line to have a paragraph break after it. At least iirc, the entire "first paragraph" is considered the "short description", which would include the "Forces TLS" bit as currently written. Same for the method below.

src/client_builder.rs
jonhoo: I think I'm sold, including probably just removing this kind of "provide a handler" method. As you say, for customizing TCP users can just do that before passing the connection to `Client`, and for TLS configuration we can do better anyway.

Part of what sold me here is also that the current release version of `imap` doesn't even have this kind of a method, so it can't be something a lot of people depend on :p

urkle:
@jonhoo I rebased, squashed and cleaned things up a bit based on the latest comments.

Can you re-review things?

src/client_builder.rs
```diff
@@ -61,41 +155,132 @@ where
...
+    /// You should only use this as a last resort as it allows another server to impersonate the
+    /// server you think you're talking to, which would include being able to receive your
+    /// credentials.
+    ///
+    /// See [`native_tls::TlsConnectorBuilder::danger_accept_invalid_certs`],
+    /// [`native_tls::TlsConnectorBuilder::danger_accept_invalid_hostnames`],
+    /// [`rustls::ClientConfig::dangerous`]
+    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+    pub fn danger_no_tls_verify(&mut self, no_tls_verify: bool) -> &mut Self {
+        self.no_tls_verify = no_tls_verify;
```
jonhoo, line 208: The double negation here strikes me as unfortunate. Can we name the field and the method just `danger_tls_verify` and `tls_verify`, and have the latter default to `true` instead?

src/client_builder.rs
```diff
@@ -1,55 +1,149 @@
...
+lazy_static! {
+    static ref STARTLS_CHECK_REGEX: regex::bytes::Regex =
+        regex::bytes::Regex::new(r"\bSTARTTLS\b").unwrap();
+}
+
+/// The connection mode we are going to use
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ConnectionMode {
+    /// Automatically detect what connection mode should be used.
+    /// This will use TLS if the port is 993 and StartTLls if available.
```
jonhoo, line 60:
```suggestion
    /// This will use TLS if the port is 993, and StartTLS if the server says it's available.
```

src/client_builder.rs
jonhoo: Alternatively, `danger_skip_tls_verify`.

src/client_builder.rs
jonhoo: I agree — the helpers also end up confusing because it's hard to tell that `tls` sets the _mode_, but `native_tls` sets the _kind_. Having just mode and kind is clearer I think.

src/client_builder.rs
```diff
@@ -61,41 +155,132 @@ where
...
+    /// Sets the Connection mode to use for this connection
+    pub fn mode(&mut self, mode: ConnectionMode) -> &mut Self {
+        self.mode = mode;
         self
     }
 
-    /// Return a new [`Client`] using a `native-tls` transport.
+    /// Sets the TLS backend to use for this connection.
+    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+    pub fn tls_kind(&mut self, kind: TlsKind) -> &mut Self {
```
jonhoo, line 188: If we're now getting rid of the helpers, I think we should have this also specifically set `mode` to be `Tls` _if_ it is currently set to `Tcp`.

src/client_builder.rs
```diff
@@ -114,37 +299,115 @@ where
...
+                    client.read_greeting()?;
+                    greeting_read = true;
 
-        let mut client = Client::new(tls);
-        if !self.starttls {
+                    let capabilities = client.capabilities()?;
+                    if capabilities.has(&Capability::Atom("STARTTLS".into())) {
+                        client.run_command_and_check_ok("STARTTLS")?;
+                        let tcp = client.into_inner()?;
+                        Box::new(handshake(self.domain.as_ref(), tcp)?)
```
jonhoo, line 332: I think this is wrong if the user hasn't enabled any of the TLS-providing features. It'll call through to `handshake`, but `handshake` then just returns `Ok(tcp)`, even though we've told the server to switch to an encrypted channel. You probably need a `cfg!()` here to not send `STARTTLS` if we don't have TLS capability.

src/client_builder.rs
```diff
@@ -61,41 +155,132 @@ where
...
+    ///     .starttls().connect()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn connect(&self) -> Result<Client<Connection>> {
+        self.connect_with(|_domain, tcp| {
+            #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+            return self.build_tls_connection(tcp);
+            #[cfg(all(not(feature = "native-tls"), not(feature = "rustls-tls")))]
+            return Ok(tcp);
```
jonhoo, line 273: This is a nit, but I'd kind of like for `handshake` to be called something like `tls_handshake`, and then for us to ensure in `connect_with` that we never call it if we don't have any TLS capability features enabled. Which would then make this be a `unreachable!` rather than `return Ok(tcp)`. What do you think?

src/client_builder.rs
```diff
@@ -61,41 +155,132 @@ where
...
         })
     }
 
-    /// Make a [`Client`] using a custom TLS initialization. This function is intended
+    /// Make a [`Client`] using a custom initialization. This function is intended
     /// to be used if your TLS setup requires custom work such as adding private CAs
     /// or other specific TLS parameters.
     ///
+    /// Note, if the connection does not end up as TLS, handshake will not be called.
```
jonhoo, line 281: This isn't _quite_ true because we may call `handshake` (currently at least) even if no TLS-enabling features are on, which would mean it's called despite the connection _not_ ending up as TLS.

src/client_builder.rs
```diff
@@ -1,55 +1,149 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum TlsKind {
+    /// Use the NativeTLS backend
+    #[cfg(feature = "native-tls")]
+    Native,
+    /// Use the Rustls backend
+    #[cfg(feature = "rustls-tls")]
+    Rust,
+    /// Use whatever backend is available (rustls used it both are compiled in)
```
jonhoo, line 83:
```suggestion
    /// Use whatever backend is available (uses rustls if both are available)
```

src/client_builder.rs
```diff
@@ -61,41 +155,132 @@ where
         ClientBuilder {
             domain,
             port,
-            starttls: false,
+            mode: ConnectionMode::Auto,
```
jonhoo, line 158: I'm tempted to have `mode` default to `ConnectionMode::Tls`, just because these days it really is what most sane clients should be using (even `STARTTLS` has its problems). Users can then always override that behavior if they know they want to connect to a less-secure server. Maybe we have a helper method called `.compatible()` or something that will set `Auto + Any`?

src/client_builder.rs
```diff
@@ -1,55 +1,149 @@
...
+}
+
+/// The connection mode we are going to use
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ConnectionMode {
+    /// Automatically detect what connection mode should be used.
+    /// This will use TLS if the port is 993 and StartTLls if available.
+    Auto,
+    /// A plain unencrypted Tcp connection
+    Tcp,
```
jonhoo, line 63: I think I prefer the name `Plaintext` for this. What do you think?

jonhoo (changes requested):
We're so so close! Thanks for not giving up on this :sweat_smile: 
A few more nits, but they're small.

I also think we might want to add `= Connection` to the `T` generic type parameter on both `Connection` and `Session`, just so that in the (common) case where people use the builder, they don't have to specify `<Connection>` everywhere.

src/client_builder.rs
urkle: Skip tls verify sounds better. as the default should be to verify w/o needing to call anything.

src/client_builder.rs
urkle: Which in another comment you suggested getting rid of connect_with entirely.  Do you still want to keep connect_with? (which I agree with the unreachable! comment before and will remove that code-path)

src/client_builder.rs
urkle: Most of my clients do starttls by default vs using the SSL only port.   And anyways if they connect to port 993 it'll use Tls full anyway in Auto mode.

Maybe add a new mode that is AutoTls that will connect to any Tls (starttls or tls) by default only (and not non-tls).. and if the user wants to opt-in to anything (even insecure) they can just set it to auto explicitly?

src/client_builder.rs
urkle: I disagree with that one as they could intent mode to be StartTls, thus we really don't know.  And with my proposal of having the default AutoTls (don't allow plain text by default) this becomes even more so unexpected behavior.

urkle:
@jonhoo I rebased + squashed and then added a fixup addressing the latest comments

- added new Error for TlsNotConfigured  (when no tls is compiled in) and StartTlsNotAvailable (server does not provide but was requested)
- added AutoTls mode (default when tls available) which will refuse to connect via Plaintext
  - when tls is NOT available should we default to Plaintext vs Auto?
- renamed Tcp to Plaintext
- more cfg guards for when tls is disabled to return TlsNotConfigured (even in your unreachable suggestion)
- danger_no_tls_verify -> danger_skip_tls_verify
- removed public "helpers" starttls, tls, native_tls, and rusttls

Other items I am wondering about
1) remove connect_with completely you had made one comment about this that if the user want's to set it it up themselves they can simply wrap it in the Client.  Though if we do this, should we expose into_inner() publicly?
2) change mode and tls_kind, and danger_skip_tls_verify (the 3 config methods) not be &mut self but instead be simply mut self (consuming setters).   (`fn mode(mut self, mode: ConnectionMode) -> Self`)
This would make it so we don't need the configuration to be mut in the common case.

bsodmike:
Hello! @urkle Running tests locally for your wip branch I am seeing:

```rust
---- auto_tls stdout ----
thread 'auto_tls' panicked at tests/builder_integration.rs:103:31:
called `Result::unwrap()` on an `Err` value: StartTlsNotAvailable
```

Do we expect to enforce [needing StartTLS even in the AutoTLS scenario?](https://github.com/jonhoo/rust-imap/pull/245/files#diff-c0004691670dfac759abb1f0d5b3f13981f62e6065a35c3ff69ff6bc0a5b5a19R101-R103)

urkle:
> ```rust
> ---- auto_tls stdout ----
> thread 'auto_tls' panicked at tests/builder_integration.rs:103:31:
> called `Result::unwrap()` on an `Err` value: StartTlsNotAvailable
> ```

@bsodmike the intention here is that if you use AutoTls, then it will ONLY connect via a secured connection.  If you use port 993 then strait Tls, if you use any other port than it'll use StartTls, and if no starttls is supported it'll yield that StartTlsNotAvailable error.

I just pushed up a force commit to correct an issue in the AutoTls branch where if no TLS provider is compiled in it returns the wrong error message, it should have yielded TlsNotConfigured in that scenario.

The intention of the "Auto" mode is that it'll allow connecting to an Imap server via plain text if no starttls is available. Thus that, in general, is not preferred as it is less secure.

bsodmike:
> @bsodmike the intention here is that if you use AutoTls, then it will ONLY connect via a secured connection.

So it seems (by default) the `auto_tls` is meant to fail but the test asserts it should pass instead? If that's the case, the test needs patching up :)

src/client_builder.rs
```diff
@@ -61,49 +160,85 @@ where
         ClientBuilder {
             domain,
             port,
-            starttls: false,
+            #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
+            mode: ConnectionMode::AutoTls,
+            #[cfg(all(not(feature = "native-tls"), not(feature = "rustls-tls")))]
+            mode: ConnectionMode::Auto,
```
jonhoo, line 166: One thing I worry about here is that if someone pulls in this dependency with `default-features = false`, they might not ever realize that they're doing plaintext connections. I wonder if the default mode if not TLS feature is enabled should be one that will _always_ error saying "No TLS implementation available". And we should require an _explicit_ `.mode(ConnectionMode::Plaintext)`.

src/client_builder.rs
```diff
@@ -61,49 +160,85 @@ where
...
+        });
     }
 
-    /// Make a [`Client`] using a custom TLS initialization. This function is intended
+    /// Make a [`Client`] using a custom initialization. This function is intended
     /// to be used if your TLS setup requires custom work such as adding private CAs
     /// or other specific TLS parameters.
     ///
+    /// Note, if the connection does not end up as TLS, handshake will not be called.
+    /// This can happen if connection mode is set to Tcp, or the server does not support starttls.
```
jonhoo, line 231:
```suggestion
    /// This can happen if connection mode is set to "plaintext", or if the server does not support STARTTLS (and the connection mode is `Auto`).
```

src/client_builder.rs
jonhoo: Which, I suppose, is really just saying that I think the default should _always_ be `AutoTls`.

jonhoo (changes requested):
This is looking really good! Mostly smaller nits now.

As for your questions:

1. I think exposing `.into_inner` publicly is probably a good idea. Now that `connect_with` specifically only takes a thing that initiates a TLS handshake, I don't think it's very useful publicly any more. Let's make it non-`pub`, and then we can open it up if someone specifically asks for it in the future :+1: 
2. I like that idea :+1:

src/client.rs
```diff
@@ -340,11 +340,24 @@ impl<T: Read + Write> Client<T> {
     ///
     /// This consumes `self` since the Client is not much use without
     /// an underlying transport.
+    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
```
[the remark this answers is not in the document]
urkle: I think it was because it is only used in that scenario..  So it was to silence warnings:)  but since we are now making this pub I'll remove the cfg.

src/client.rs
```diff
@@ -340,11 +340,24 @@ impl<T: Read + Write> Client<T> {
...
         let res = self.conn.stream.into_inner()?;
         Ok(res)
     }
 
+    /// The [`CAPABILITY` command](https://tools.ietf.org/html/rfc3501#section-6.1.1) requests a
+    /// listing of capabilities that the server supports.  The server will include "IMAP4rev1" as
+    /// one of the listed capabilities. See [`Capabilities`] for further details.
+    ///
+    /// This allows reading capabilities before authentication.
+    pub fn capabilities(&mut self) -> Result<Capabilities> {
```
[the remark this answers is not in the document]
urkle: Yes, this was so I could query the capabilities BEFORE logging in. (to verify that STARTTLS is available)

src/client_builder.rs
```diff
@@ -1,55 +1,159 @@
...
+}
+
+lazy_static! {
+    static ref STARTLS_CHECK_REGEX: regex::bytes::Regex =
+        regex::bytes::Regex::new(r"\bSTARTTLS\b").unwrap();
+}
+
+/// The connection mode we are going to use
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ConnectionMode {
```
jonhoo, line 57: This needs to be marked with `#[non_exhaustive]` since the variants vary (and we may add others in the future).

src/client_builder.rs
```diff
@@ -1,55 +1,159 @@
...
+    /// Automatically detect what connection mode should be used.
+    ///
+    /// This will use TLS if the port is 993, and otherwise STARTTLS if available.
+    /// If no TLS communication mechanism is available, the connection will fail.
+    #[cfg(any(feature = "native-tls", feature = "rustls-tls"))]
```
jonhoo, line 62: I think this needs to not be `cfg`-guarded since we now always set it in `ClientBuilder::new` (in fact the code probably won't compile as-is).

jonhoo (approved):
Excellent, I think that does it! Just two trivial changes left to make so the code will build, and then I think we're good to go with a rebase!

src/client_builder.rs
urkle: Should we just make the otherones also not Cfg guarded since we have the proper error code response now when no Tls is compiled in?

jonhoo: No, let's keep those guarded (we can always un-guard them now that it's `#[non_exhaustive]`). It'll still then serve as a signal to folks to remember to enable the TLS features if they specifically try to enforce TLS.

jonhoo (approved):
:tada: :tada: :tada: 
Thank you for sticking with it!

jonhoo:
Ah, looks like we also now specifically have to set `Auto` (not `AutoTls`) for the greenmail tests, and [fix some doctests](https://github.com/jonhoo/rust-imap/actions/runs/6397840604/job/17366686378?pr=245#step:7:40).

urkle:
@jonhoo I rebased + squashed and fixed up the test errors.

for connect_with I simply removed the comments & doc example entirely since it's now an internal method.

urkle:
@jonhoo we are green except for the persisting code-coverage one.  I am reviewing those to at least add some coverage to my code changes where possible.

urkle:
@jonhoo 100% green!

jonhoo (approved):
Amazing, thank you! Finally, we can get this in :tada: :heart:

jonhoo:
Released in alpha.12 — thanks again for sticking with this for _so_ long!

PR #249, opened 2022-12-13
Add support for RFC 5819: LIST-STATUS extension

bitfehler:
See https://tools.ietf.org/html/rfc5819

Hi there. I would love to add support for RFC 5819. It is extremely valuable for clients to perform a full mailbox sync with as little requests as possible. This patch is not yet well-documented, but I wanted to get some feedback early on, specifically on the following:

- I implemented it as an extension, in the `extensions` folder. I think it makes sense there, but it could of course be done differently, potentially even merged with the regular `LIST` command, though I think that would make the API a little awkward.
- I took some cues from the `Names` data structure. I hope the name `ExtendedNames` makes sense, as this version of `LIST` is usually referred to as "extended LIST".
- Instead of creating a struct for each single response I opted for `(Name, Option<Mailbox>)`, because I think that most accurately reflects what is happening. A case could be made of course for abstracting that away, so let me know what you think.
- Implementing this made me realize that the `Mailbox` type as response to a `STATUS` command is actually a little confusing. It has non-`Option<>`-members that might not be returned by a `STATUS` command. However, I felt that for now just mirroring `STATUS` responses is the right thing to do. If you think this might be something worth changing while we are about to break the API anyways, let me know - I'm happy to take a stab.

And of course whatever else catches your eye :slightly_smiling_face:

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/249)

### DIFF
--- a/src/extensions/list_status.rs
+++ b/src/extensions/list_status.rs
@@ -1,0 +1,197 @@
+//! Adds support for the IMAP LIST-STATUS extension specificed in [RFC
+//! 5819](https://tools.ietf.org/html/rfc5819).
+
+use crate::client::{validate_str, Session};
+use crate::error::{Error, ParseError, Result};
+use crate::parse::try_handle_unilateral;
+use crate::types::{Mailbox, Name, UnsolicitedResponse};
+use imap_proto::types::{MailboxDatum, Response, StatusAttribute};
+use ouroboros::self_referencing;
+use std::io::{Read, Write};
+use std::slice::Iter;
+use std::sync::mpsc;
+
+/// A wrapper for one or more [`Name`] responses paired with optional [`Mailbox`] responses.
+///
+/// This structure represents responses to a LIST-STATUS command, as implemented in
+/// [`Session::list_status`]. See [RFC 5819, section 2](https://tools.ietf.org/html/rfc5819.html#section-2).
+#[self_referencing]
+pub struct ExtendedNames {
+    data: Vec<u8>,
+    #[borrows(data)]
+    #[covariant]
+    pub(crate) extended_names: Vec<(Name<'this>, Option<Mailbox>)>,
+}
+
+impl ExtendedNames {
+    /// Parse one or more LIST-STATUS responses from a response buffer
+    pub(crate) fn parse(
+        owned: Vec<u8>,
+        unsolicited: &mut mpsc::Sender<UnsolicitedResponse>,
+    ) -> core::result::Result<Self, Error> {
+        ExtendedNamesTryBuilder {
+            data: owned,
+            extended_names_builder: |input| {
+                let mut lines: &[u8] = &input;
+                let mut names = Vec::new();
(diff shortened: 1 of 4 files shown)


bitfehler:
Oh, and I can of course add tests once we agree on the rest :wink:

bitfehler:
@jonhoo gentle ping :innocent: would be great to know if you have any opinions?

jonhoo:
This looks entirely reasonable to me (subject to tests being added)! I'm a little surprised to see `Option<Mailbox>`, but I guess the server can just choose to not include any mailbox info in there?

I forget why `Mailbox` has some non-`Option` fields. I thought I made it accurately mirror what the spec said were required fields, but may have mis-parsed? I think we should leave it as-is for this PR though!

bitfehler:
> I'm a little surprised to see `Option<Mailbox>`, but I guess the server can just choose to not include any mailbox info in there?

The RFC is a bit weird. It says the server MUST include that information, but goes on to explicitly specify that in certain error situations the server may omit it. I've never seen that happen, but you never know...

I updated the documentation a bit and added some tests, including an integration test. I think the coverage should be quite good, but for some reason the coverage runner seems to have problems uploading the report? :thinking: 

Note: I also added a commit that changes the annotation for the qresync integration test from `#[ignore]` to `#[cfg(feature = "test-full-imap")]`. I wrote that test, and it works fine against e.g. Cyrus. But I think that feature did not exist back then? Anyways, seemed like the right thing to do while I was touching that file.

src/extensions/list_status.rs
```diff
@@ -0,0 +1,197 @@
...
+pub struct ExtendedNames {
+    data: Vec<u8>,
+    #[borrows(data)]
+    #[covariant]
+    pub(crate) extended_names: Vec<(Name<'this>, Option<Mailbox>)>,
+}
+
+impl ExtendedNames {
+    /// Parse one or more LIST-STATUS responses from a response buffer
+    pub fn parse(
```
jonhoo, line 28: I don't think this needs to be `pub`?

jonhoo (approved):
I'm okay with this modulo the one `fn` that I _think_ should be `pub(crate)`?

jonhoo:
Not sure why the CI step is being sad. I wonder if it just needs a re-run..

src/extensions/list_status.rs
bitfehler: Agree, made it `pub(crate)`

bitfehler:
Sorry for the double update, just now saw that you fixed the coverage upload on main.

jonhoo:
Hmm, it looks like now none of the CI steps ran :thinking: 
Can you try rebasing onto `main` and force-pushing?

jonhoo (approved):
All right, let's land this thing! Thanks :)

jonhoo:
Released in alpha.10 :tada:

PR #251, opened 2023-01-16
Un-export some exposed `parse()` functions

bitfehler:
As @jonhoo pointed out [here](https://github.com/jonhoo/rust-imap/pull/249#pullrequestreview-1249037461), the various types' `parse()` functions should not be exported. Some are, though. The 3.0 release is probably a good time to fix that.

Note: some types (e.g. `Mailbox`, `Deleted`) didn't have their `parse()` function exposed, so I think it's pretty safe to assume that nobody built anything seriously using these.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/251)

### DIFF
--- a/src/types/acls.rs
+++ b/src/types/acls.rs
@@ -115,7 +115,7 @@ pub struct AclResponse {
 
 impl AclResponse {
     /// Parse the given input into a [`Acl`] response.
-    pub fn parse(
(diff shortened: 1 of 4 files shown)


jonhoo (approved):
Thanks!

PR #255, opened 2023-02-06
Implementing capability caching

dequbed:
Based on [this comment by @jonhoo on #243](https://github.com/jonhoo/rust-imap/pull/243#discussion_r1008771894) this PR adds the ability for `Client` and `Session` to cache returned `Capabilities`.

Future work should try to parse (the optionally sent) Capabilities from untagged responses to `EHLO`, `STARTTLS`, and `AUTHENTICATE` automatically; if none are sent they should be automatically queried.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/255)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -139,7 +139,7 @@ fn validate_sequence_set(
 pub struct Session<T: Read + Write> {
     conn: Connection<T>,
 
-    // Capabilities are almost guaranteed to chance if encryption state or authentication state
+    // Capabilities are almost guaranteed to change if encryption state or authentication state
     // changes, so caching them in `Connection` is inappropiate.
     capability_cache: Option<Capabilities>,
 
@@ -159,7 +159,7 @@ pub struct Session<T: Read + Write> {
 pub struct Client<T: Read + Write> {
     conn: Connection<T>,
 
-    // Capabilities are almost guaranteed to chance if encryption state or authentication state
+    // Capabilities are almost guaranteed to change if encryption state or authentication state
     // changes, so caching them in `Connection` is inappropiate.
     capability_cache: Option<Capabilities>,
 }
@@ -362,16 +362,16 @@ impl<T: Read + Write> Client<T> {
     /// This method will always bypass the local capabilities cache and send a `CAPABILITY` command
     /// to the server. The [`Self::capabilities()`] method can be used when returning a cached
     /// response is acceptable.
-    pub fn capabilities_refresh(&mut self) -> Result<&Capabilities> {
+    pub fn current_capabilities(&mut self) -> Result<&Capabilities> {
         let (mut tx, _rx) = mpsc::channel();
         let caps = self.run_command_and_read_response("CAPABILITY")
             .and_then(|lines| Capabilities::parse(lines, &mut tx))?;
         self.capability_cache = Some(caps);
 
-        self.capability_cache.as_ref()
+        Ok(self.capability_cache.as_ref()
             // This path will not be hit; if the cache is not populated the above calls will either
             // populate it or return with an early error.
-            .ok_or_else(|| panic!("CAPABILITY call did not populate capability cache!"))
+            .expect("CAPABILITY call did not populate capability cache!"))
     }
 
     /// The [`CAPABILITY` command](https://tools.ietf.org/html/rfc3501#section-6.1.1) requests a
@@ -381,7 +381,7 @@ impl<T: Read + Write> Client<T> {
     /// This function will not query the server if a set of capabilities was cached, but request
     /// and cache capabilities from the server otherwise. The [`Self::capabilities_refresh`] method
     /// can be used to refresh the cache by forcing a `CAPABILITY` command to be send.
-    pub fn capabilities_ref(&mut self) -> Result<&Capabilities> {
+    pub fn capabilities(&mut self) -> Result<&Capabilities> {
         let (mut tx, _rx) = mpsc::channel();
         if self.capability_cache.is_none() {
             let caps = self.run_command_and_read_response("CAPABILITY")
@@ -390,10 +390,10 @@ impl<T: Read + Write> Client<T> {
             self.capability_cache = Some(caps);
         }
 
-        self.capability_cache.as_ref()
+        Ok(self.capability_cache.as_ref()
             // This path will not be hit; if the cache is not populated the above `if` will either
             // populate it or return with an early error.
-            .ok_or_else(|| panic!("CAPABILITY call did not populate capability cache!"))
+            .expect("CAPABILITY call did not populate capability cache!"))
     }
 
     /// Log in to the IMAP server. Upon success a [`Session`](struct.Session.html) instance is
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -345,6 +355,47 @@ impl<T: Read + Write> Client<T> {
...
+    pub fn capabilities_refresh(&mut self) -> Result<&Capabilities> {
+        let (mut tx, _rx) = mpsc::channel();
+        let caps = self.run_command_and_read_response("CAPABILITY")
+            .and_then(|lines| Capabilities::parse(lines, &mut tx))?;
+        self.capability_cache = Some(caps);
+
+        self.capability_cache.as_ref()
+            // This path will not be hit; if the cache is not populated the above calls will either
+            // populate it or return with an early error.
+            .ok_or_else(|| panic!("CAPABILITY call did not populate capability cache!"))
```
jonhoo, line 374: Let's use `.expect` here — that's what this is really saying. Then just wrap the output of that in `Ok()`.

src/client.rs
```diff
@@ -138,6 +138,11 @@ fn validate_sequence_set(
 #[derive(Debug)]
 pub struct Session<T: Read + Write> {
     conn: Connection<T>,
+
+    // Capabilities are almost guaranteed to chance if encryption state or authentication state
```
jonhoo, line 142:
```suggestion
    // Capabilities are almost guaranteed to change if encryption state or authentication state
```

src/client.rs
```diff
@@ -153,6 +158,10 @@ pub struct Session<T: Read + Write> {
 #[derive(Debug)]
 pub struct Client<T: Read + Write> {
     conn: Connection<T>,
+
+    // Capabilities are almost guaranteed to chance if encryption state or authentication state
```
jonhoo, line 162:
```suggestion
    // Capabilities are almost guaranteed to change if encryption state or authentication state
```

src/types/capabilities.rs
```diff
@@ -43,6 +44,17 @@ pub struct Capabilities {
     pub(crate) capabilities: HashSet<Capability<'this>>,
 }
 
+impl Clone for Capabilities {
+    fn clone(&self) -> Self {
+        // Give _rx a name so it's not immediately dropped. Otherwise any unsolicited responses
+        // that would be send there will return a SendError instead of the parsed response simply
+        // being dropped later.
```
jonhoo, line 51: Let's also add a note here that any unsolicited items in here can safely be dropped because they must already have been communicated the first time the capabilities were parsed.

src/client.rs
```diff
@@ -771,12 +823,52 @@ impl<T: Read + Write> Session<T> {
...
+
     /// The [`CAPABILITY` command](https://tools.ietf.org/html/rfc3501#section-6.1.1) requests a
     /// listing of capabilities that the server supports.  The server will include "IMAP4rev1" as
     /// one of the listed capabilities. See [`Capabilities`] for further details.
     pub fn capabilities(&mut self) -> Result<Capabilities> {
-        self.run_command_and_read_response("CAPABILITY")
-            .and_then(|lines| Capabilities::parse(lines, &mut self.unsolicited_responses_tx))
+        // TODO: This emulated the same behaviour as before, with each call issuing a command.
+        //       It may be sensible to allow hitting the cache here.
+        self.capabilities_refresh().map(|caps| caps.clone())
```
jonhoo, line 871: Given we're still in `beta` for the next `imap` release, I think we should make the change the "right way" — have `capabilities_ref` just be `capabilities`.

src/client.rs
```diff
@@ -771,12 +823,52 @@ impl<T: Read + Write> Session<T> {
...
     }
 
+    /// The [`CAPABILITY` command](https://tools.ietf.org/html/rfc3501#section-6.1.1) requests a
+    /// listing of capabilities that the server supports.  The server will include "IMAP4rev1" as
+    /// one of the listed capabilities. See [`Capabilities`] for further details.
+    ///
+    /// This method will always bypass the local capabilities cache and send a `CAPABILITY` command
+    /// to the server. The [`Self::capabilities()`] method can be used when returning a cached
+    /// response is acceptable.
+    pub fn capabilities_refresh(&mut self) -> Result<&Capabilities> {
```
jonhoo, line 833: I propose the name `current_capabilities`.

src/client.rs
```diff
@@ -138,6 +138,11 @@ fn validate_sequence_set(
 #[derive(Debug)]
 pub struct Session<T: Read + Write> {
     conn: Connection<T>,
+
+    // Capabilities are almost guaranteed to chance if encryption state or authentication state
+    // changes, so caching them in `Connection` is inappropiate.
+    capability_cache: Option<Capabilities>,
```
jonhoo, line 144: I don't know that I believe that we shouldn't have the cache be in `Connection`. We can always just forcibly-reset it whenever the relevant underlying state changes (e.g., in `login`). What do you think?

jonhoo (changes requested):

src/client.rs
```diff
@@ -345,6 +355,47 @@ impl<T: Read + Write> Client<T> {
...
+            // populate it or return with an early error.
+            .expect("CAPABILITY call did not populate capability cache!"))
+    }
+
+    /// The [`CAPABILITY` command](https://tools.ietf.org/html/rfc3501#section-6.1.1) requests a
+    /// listing of capabilities that the server supports.  The server will include "IMAP4rev1" as
+    /// one of the listed capabilities. See [`Capabilities`] for further details.
+    ///
+    /// This function will not query the server if a set of capabilities was cached, but request
+    /// and cache capabilities from the server otherwise. The [`Self::capabilities_refresh`] method
```
jonhoo, line 382: This reference needs to be updated, and same with the one in `Session`.

src/client.rs
dequbed: I had two arguments for not caching the capabilities in the `Connection`: Firstly and definitely the weaker of the two: Separation of concerns, namely doing so makes `Connection` more than just a very thin shim over the underlying `io::{Read + Write}`. Secondly and my main argument is that caching in `Connection` has a smaller fuck-up-protection factor for the maintainers, i.e. "*us, three years older*". Clearing the cache becomes an active action that needs to be thought of, and *forgetting* to do so is still syntactically and semantically valid code, just one

src/client.rs
dequbed: Goddammit, how did I miss that -.-'

src/client.rs
jonhoo: While I think that's true, I also think the current split is a foot-gun because it means we have to replicate the cache state and logic in two places. If it was on `Connection`, it would be in one place, which means one place to make improvements / bugfixes over time. I think I'd err on the side of simplicity here, as it'll be quite rare that we add a new way to wrap `Connection`. And if we do, we're likely to do so by copy-pasting an existing wrapper first.

jonhoo:
I think the last bit here now is https://github.com/jonhoo/rust-imap/pull/255#discussion_r1125573266

bogthe:
Hi @dequbed, I really liked that you were adding support for SASL in #243. Do you have any plans to update this PR? If you're short on time, I would be happy to help getting this (and that one) merged!

PR #260, opened 2023-05-11
Replace mpsc::{Sender,Receiver} by VecDeque

### LINKED ISSUE #254: UnsolicitedResponse isn't Send
As reported in https://github.com/jonhoo/rust-imap/issues/250, the `UnsolicitedResponse` type isn't `Send`, but it should be. We should fix that.

soywod:
Fixes #254.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/260)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ bufstream = "0.1.3"
 imap-proto = "0.16.1"
 nom = { version = "7.1.0", default-features = false }
 base64 = "0.22"
-chrono = { version = "0.4", default-features = false, features = ["std"]}
+chrono = { version = "0.4.37", default-features = false, features = ["std"]}
 lazy_static = "1.4"
 ouroboros = "0.18.0"
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -3,10 +3,10 @@ use bufstream::BufStream;
 use chrono::{DateTime, FixedOffset};
 use imap_proto::Response;
 use std::collections::HashSet;
+use std::collections::VecDeque;
 use std::io::{Read, Write};
 use std::ops::{Deref, DerefMut};
 use std::str;
-use std::sync::mpsc;
 
 use super::authenticator::Authenticator;
 use super::error::{Bad, Bye, Error, No, ParseError, Result, TagMismatch, ValidateError};
@@ -141,11 +141,9 @@ fn validate_sequence_set(
 #[derive(Debug)]
 pub struct Session<T: Read + Write> {
     conn: Connection<T>,
-    pub(crate) unsolicited_responses_tx: mpsc::Sender<UnsolicitedResponse>,
-
     /// Server responses that are not related to the current command. See also the note on
     /// [unilateral server responses in RFC 3501](https://tools.ietf.org/html/rfc3501#section-7).
-    pub unsolicited_responses: mpsc::Receiver<UnsolicitedResponse>,
+    pub unsolicited_responses: VecDeque<UnsolicitedResponse>,
 }
 
 /// An (unauthenticated) handle to talk to an IMAP server. This is what you get when first
@@ -262,7 +260,7 @@ impl<'a, T: Read + Write> AppendCmd<'a, T> {
         self.session.stream.flush()?;
         self.session
             .read_response()
-            .and_then(|(lines, _)| parse_append(&lines, &mut self.session.unsolicited_responses_tx))
+            .and_then(|(lines, _)| parse_append(&lines, &mut self.session.unsolicited_responses))
     }
 }
 
@@ -530,11 +528,9 @@ impl<T: Read + Write> Client<T> {
 impl<T: Read + Write> Session<T> {
     // not public, just to avoid duplicating the channel creation code
     fn new(conn: Connection<T>) -> Self {
-        let (tx, rx) = mpsc::channel();
         Session {
             conn,
-            unsolicited_responses: rx,
-            unsolicited_responses_tx: tx,
+            unsolicited_responses: VecDeque::new(),
         }
     }
 
@@ -561,7 +557,7 @@ impl<T: Read + Write> Session<T> {
             "SELECT {}",
             validate_str("SELECT", "mailbox", mailbox_name.as_ref())?
         ))
-        .and_then(|(lines, _)| parse_mailbox(&lines[..], &mut self.unsolicited_responses_tx))
+        .and_then(|(lines, _)| parse_mailbox(&lines[..], &mut self.unsolicited_responses))
     }
 
     /// The `EXAMINE` command is identical to [`Session::select`] and returns the same output;
@@ -573,7 +569,7 @@ impl<T: Read + Write> Session<T> {
             "EXAMINE {}",
             validate_str("EXAMINE", "mailbox", mailbox_name.as_ref())?
         ))
-        .and_then(|(lines, _)| parse_mailbox(&lines[..], &mut self.unsolicited_responses_tx))
+        .and_then(|(lines, _)| parse_mailbox(&lines[..], &mut self.unsolicited_responses))
     }
 
     /// Fetch retrieves data associated with a set of messages in the mailbox.
@@ -640,7 +636,7 @@ impl<T: Read + Write> Session<T> {
         query: impl AsRef<str>,
     ) -> Result<Fetches> {
         if sequence_set.as_ref().is_empty() {
-            Fetches::parse(vec![], &mut self.unsolicited_responses_tx)
+            Fetches::parse(vec![], &mut self.unsolicited_responses)
         } else {
             let synopsis = "FETCH";
             self.run_command_and_read_response(&format!(
@@ -648,7 +644,7 @@ impl<T: Read + Write> Session<T> {
                 validate_sequence_set(synopsis, "seq", sequence_set.as_ref())?,
                 validate_str_noquote(synopsis, "query", query.as_ref())?
             ))
-            .and_then(|lines| Fetches::parse(lines, &mut self.unsolicited_responses_tx))
+            .and_then(|lines| Fetches::parse(lines, &mut self.unsolicited_responses))
         }
     }
 
@@ -660,7 +656,7 @@ impl<T: Read + Write> Session<T> {
         query: impl AsRef<str>,
     ) -> Result<Fetches> {
         if uid_set.as_ref().is_empty() {
-            Fetches::parse(vec![], &mut self.unsolicited_responses_tx)
+            Fetches::parse(vec![], &mut self.unsolicited_responses)
         } else {
             let synopsis = "UID FETCH";
             self.run_command_and_read_response(&format!(
@@ -668,14 +664,14 @@ impl<T: Read + Write> Session<T> {
                 validate_sequence_set(synopsis, "seq", uid_set.as_ref())?,
                 validate_str_noquote(synopsis, "query", query.as_ref())?
             ))
-            .and_then(|lines| Fetches::parse(lines, &mut self.unsolicited_responses_tx))
+            .and_then(|lines| Fetches::parse(lines, &mut self.unsolicited_responses))
         }
     }
 
     /// Noop always succeeds, and it does nothing.
     pub fn noop(&mut self) -> Result<()> {
         self.run_command_and_read_response("NOOP")
-            .and_then(|lines| parse_noop(lines, &mut self.unsolicited_responses_tx))
+            .and_then(|lines| parse_noop(lines, &mut self.unsolicited_responses))
     }
 
     /// Logout informs the server that the client is done with the connection.
@@ -807,7 +803,7 @@ impl<T: Read + Write> Session<T> {
     /// one of the listed capabilities. See [`Capabilities`] for further details.
     pub fn capabilities(&mut self) -> Result<Capabilities> {
         self.run_command_and_read_response("CAPABILITY")
-            .and_then(|lines| Capabilities::parse(lines, &mut self.unsolicited_responses_tx))
+            .and_then(|lines| Capabilities::parse(lines, &mut self.unsolicited_responses))
     }
 
     /// The [`EXPUNGE` command](https://tools.ietf.org/html/rfc3501#section-6.4.3) permanently
@@ -816,7 +812,7 @@ impl<T: Read + Write> Session<T> {
     pub fn expunge(&mut self) -> Result<Deleted> {
         self.run_command("EXPUNGE")?;
         self.read_response()
-            .and_then(|(lines, _)| parse_expunge(lines, &mut self.unsolicited_responses_tx))
+            .and_then(|(lines, _)| parse_expunge(lines, &mut self.unsolicited_responses))
     }
 
     /// The [`UID EXPUNGE` command](https://tools.ietf.org/html/rfc4315#section-2.1) permanently
@@ -844,7 +840,7 @@ impl<T: Read + Write> Session<T> {
     pub fn uid_expunge(&mut self, uid_set: impl AsRef<str>) -> Result<Deleted> {
         self.run_command(&format!("UID EXPUNGE {}", uid_set.as_ref()))?;
         self.read_response()
-            .and_then(|(lines, _)| parse_expunge(lines, &mut self.unsolicited_responses_tx))
+            .and_then(|(lines, _)| parse_expunge(lines, &mut self.unsolicited_responses))
     }
 
     /// The [`CHECK` command](https://tools.ietf.org/html/rfc3501#section-6.4.1) requests a
@@ -934,7 +930,7 @@ impl<T: Read + Write> Session<T> {
             sequence_set.as_ref(),
             query.as_ref()
         ))
-        .and_then(|lines| Fetches::parse(lines, &mut self.unsolicited_responses_tx))
+        .and_then(|lines| Fetches::parse(lines, &mut self.unsolicited_responses))
     }
 
     /// Equivalent to [`Session::store`], except that all identifiers in `sequence_set` are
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -27,7 +27,7 @@ bufstream = "0.1.3"
 imap-proto = "0.16.1"
 nom = { version = "7.1.0", default-features = false }
 base64 = "0.21"
-chrono = { version = "0.4", default-features = false, features = ["std"]}
+chrono = { version = "0.4.24", default-features = false, features = ["std"]}
```
soywod, line 30: Not sure if it is the best way to fix the CI. Look like there is deprecated stuff between `0.4.22` (actual version) and `0.4.24`.

src/client.rs
```diff
@@ -138,11 +138,9 @@ fn validate_sequence_set(
 #[derive(Debug)]
 pub struct Session<T: Read + Write> {
     conn: Connection<T>,
-    pub(crate) unsolicited_responses_tx: mpsc::Sender<UnsolicitedResponse>,
-
     /// Server responses that are not related to the current command. See also the note on
     /// [unilateral server responses in RFC 3501](https://tools.ietf.org/html/rfc3501#section-7).
-    pub unsolicited_responses: mpsc::Receiver<UnsolicitedResponse>,
+    pub unsolicited_responses: VecDeque<UnsolicitedResponse>,
```
jonhoo, line 143: In case we want to change this again in the future, I wonder if we should stop having this be a `pub` field, and instead have a `take_unsolicited()` method, or something like it? Maybe:
```rust
impl Session {
    pub fn next_unsolicited(&mut self) -> Option<UnsolicitedResponse>;
    pub fn all_unsolicited(&mut self) -> Vec<UnsolicitedResponse>;
}
```
That way the exact type we use here isn't exposed in the API, and thus can be changed. What do you think?

src/parse.rs
```diff
@@ -354,7 +349,7 @@ pub fn parse_mailbox(
             }
             Ok((rest, Response::Expunge(n))) => {
                 lines = rest;
-                unsolicited.send(UnsolicitedResponse::Expunge(n)).unwrap();
+                unsolicited.push_front(UnsolicitedResponse::Expunge(n))
```
jonhoo, line 352: I think this should be a `push_back`, no?

jonhoo (changes requested):
Thanks for taking the time to make this change! And sorry for the slow reply :sweat_smile: 

Just one bit in there that I think is wrong, otherwise looks good :+1:

src/client.rs
d2weber: Maybe even:
```rust
    pub fn all_unsolicited(&mut self) -> impl Iterator<Item = UnsolicitedResponse> {
        std::mem::take(&mut self.unsolicited_responses).into_iter()
    }
```
Moving out of the field safes us an additional allocation of a `Vec`, while still hiding the concrete queue-type from the public interface.

jonhoo: That works too, though maybe `-> impl ExactSizeIterator` to be even nicer to the caller :)

bogthe:
Hi @soywod I really like this PR and it also looks super close to being finished. Is there anyway I can help get this merged?

soywod:
I forgot about this PR as it did not fix my initial issue. I can take a look back during this week and see if I can complete it.

soywod:
Sorry for the very long delay. The PR is ready for another review!

src/client.rs
```diff
@@ -530,14 +528,17 @@ impl<T: Read + Write> Client<T> {
...
-        let (tx, rx) = mpsc::channel();
         Session {
             conn,
-            unsolicited_responses: rx,
-            unsolicited_responses_tx: tx,
+            unsolicited_responses: VecDeque::new(),
         }
     }
 
+    /// Takes all the unsolicited responses
```
jonhoo, line 537:
```suggestion
    /// Takes all the unsolicited responses received thus far.
```

src/extensions/list_status.rs
```diff
@@ -145,13 +145,14 @@ impl<T: Read + Write> Session<T> {
...
-        self.run_command_and_read_response(&format!(
+        let lines = self.run_command_and_read_response(format!(
             "LIST {} {} RETURN (STATUS {})",
             &reference,
             mailbox_pattern.unwrap_or("\"\""),
             data_items
-        ))
-        .and_then(|lines| ExtendedNames::parse(lines, &mut self.unsolicited_responses_tx))
+        ))?;
+        let mut unsolicited_responses = self.all_unsolicited().collect();
```
jonhoo, line 154: Why does this take + collect and then pass that to `ExtendedNames::parse` rather than pass `&mut self.unsolicited_responses` in like we do elsewhere?

src/extensions/list_status.rs
```diff
@@ -171,9 +172,9 @@ mod tests {
                     * LIST (\\UnMarked) \".\" feeds\r\n\
                     * LIST () \".\" feeds.test\r\n\
                     * STATUS feeds.test (HIGHESTMODSEQ 757)\r\n";
-        let (mut send, recv) = mpsc::channel();
-        let fetches = ExtendedNames::parse(lines.to_vec(), &mut send).unwrap();
-        assert!(recv.try_recv().is_err());
+        let mut queue = VecDeque::new();
+        let fetches = ExtendedNames::parse(lines.to_vec(), &mut queue).unwrap();
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 177:
```suggestion
        assert_eq!(queue.pop_front(), None);
```
So we see what's there if the test fails.

src/extensions/metadata.rs
```diff
@@ -197,7 +197,8 @@ impl<T: Read + Write> Session<T> {
             .as_str(),
         );
         let (lines, ok) = self.run(command)?;
-        let meta = parse_metadata(&lines[..ok], &mut self.unsolicited_responses_tx)?;
+        let mut unsolicited_responses = self.all_unsolicited().collect();
+        let meta = parse_metadata(&lines[..ok], &mut unsolicited_responses)?;
```
jonhoo, line 201: Here too — why not pass `&mut self.unsolicited_responses`?

src/parse.rs
```diff
@@ -508,10 +505,10 @@ mod tests {
             Capability::Atom(Cow::Borrowed("LOGINDISABLED")),
         ];
         let lines = b"* CAPABILITY IMAP4rev1 STARTTLS AUTH=GSSAPI LOGINDISABLED\r\n";
-        let (mut send, recv) = mpsc::channel();
-        let capabilities = Capabilities::parse(lines.to_vec(), &mut send).unwrap();
+        let mut queue = VecDeque::new();
+        let capabilities = Capabilities::parse(lines.to_vec(), &mut queue).unwrap();
         // shouldn't be any unexpected responses parsed
-        assert!(recv.try_recv().is_err());
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 511:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -526,10 +523,10 @@ mod tests {
             Capability::Atom(Cow::Borrowed("STARTTLS")),
         ];
         let lines = b"* CAPABILITY IMAP4REV1 STARTTLS\r\n";
-        let (mut send, recv) = mpsc::channel();
-        let capabilities = Capabilities::parse(lines.to_vec(), &mut send).unwrap();
+        let mut queue = VecDeque::new();
+        let capabilities = Capabilities::parse(lines.to_vec(), &mut queue).unwrap();
         // shouldn't be any unexpected responses parsed
-        assert!(recv.try_recv().is_err());
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 529:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -539,18 +536,18 @@ mod tests {
     #[test]
     #[should_panic]
     fn parse_capability_invalid_test() {
-        let (mut send, recv) = mpsc::channel();
+        let mut queue = VecDeque::new();
         let lines = b"* JUNK IMAP4rev1 STARTTLS AUTH=GSSAPI LOGINDISABLED\r\n";
-        Capabilities::parse(lines.to_vec(), &mut send).unwrap();
-        assert!(recv.try_recv().is_err());
+        Capabilities::parse(lines.to_vec(), &mut queue).unwrap();
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 542:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -539,18 +536,18 @@ mod tests {
...
 
     #[test]
     fn parse_names_test() {
         let lines = b"* LIST (\\HasNoChildren) \".\" \"INBOX\"\r\n";
-        let (mut send, recv) = mpsc::channel();
-        let names = Names::parse(lines.to_vec(), &mut send).unwrap();
-        assert!(recv.try_recv().is_err());
+        let mut queue = VecDeque::new();
+        let names = Names::parse(lines.to_vec(), &mut queue).unwrap();
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 550:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -564,9 +561,9 @@ mod tests {
     #[test]
     fn parse_fetches_empty() {
         let lines = b"";
-        let (mut send, recv) = mpsc::channel();
-        let fetches = Fetches::parse(lines.to_vec(), &mut send).unwrap();
-        assert!(recv.try_recv().is_err());
+        let mut queue = VecDeque::new();
+        let fetches = Fetches::parse(lines.to_vec(), &mut queue).unwrap();
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 566:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -575,9 +572,9 @@ mod tests {
         let lines = b"\
                     * 24 FETCH (FLAGS (\\Seen) UID 4827943)\r\n\
                     * 25 FETCH (FLAGS (\\Seen))\r\n";
-        let (mut send, recv) = mpsc::channel();
-        let fetches = Fetches::parse(lines.to_vec(), &mut send).unwrap();
-        assert!(recv.try_recv().is_err());
+        let mut queue = VecDeque::new();
+        let fetches = Fetches::parse(lines.to_vec(), &mut queue).unwrap();
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 577:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -716,9 +713,9 @@ mod tests {
     fn parse_ids_test() {
         let lines = b"* SEARCH 1600 1698 1739 1781 1795 1885 1891 1892 1893 1898 1899 1901 1911 1926 1932 1933 1993 1994 2007 2032 2033 2041 2053 2062 2063 2065 2066 2072 2078 2079 2082 2084 2095 2100 2101 2102 2103 2104 2107 2116 2120 2135 2138 2154 2163 2168 2172 2189 2193 2198 2199 2205 2212 2213 2221 2227 2267 2275 2276 2295 2300 2328 2330 2332 2333 2334\r\n\
             * SEARCH 2335 2336 2337 2338 2339 2341 2342 2347 2349 2350 2358 2359 2362 2369 2371 2372 2373 2374 2375 2376 2377 2378 2379 2380 2381 2382 2383 2384 2385 2386 2390 2392 2397 2400 2401 2403 2405 2409 2411 2414 2417 2419 2420 2424 2426 2428 2439 2454 2456 2467 2468 2469 2490 2515 2519 2520 2521\r\n";
-        let (mut send, recv) = mpsc::channel();
-        let ids = parse_id_set(lines, &mut send).unwrap();
-        assert!(recv.try_recv().is_err());
+        let mut queue = VecDeque::new();
+        let ids = parse_id_set(lines, &mut queue).unwrap();
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 718:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -739,16 +736,16 @@ mod tests {
         );
 
         let lines = b"* SEARCH\r\n";
-        let (mut send, recv) = mpsc::channel();
-        let ids = parse_id_set(lines, &mut send).unwrap();
-        assert!(recv.try_recv().is_err());
+        let mut queue = VecDeque::new();
+        let ids = parse_id_set(lines, &mut queue).unwrap();
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 741:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -739,16 +736,16 @@ mod tests {
...
         let ids: HashSet<u32> = ids.iter().cloned().collect();
         assert_eq!(ids, HashSet::<u32>::new());
 
         let lines = b"* SORT\r\n";
-        let (mut send, recv) = mpsc::channel();
-        let ids = parse_id_seq(lines, &mut send).unwrap();
-        assert!(recv.try_recv().is_err());
+        let mut queue = VecDeque::new();
+        let ids = parse_id_seq(lines, &mut queue).unwrap();
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 748:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -773,14 +770,14 @@ mod tests {
         assert_eq!(None, uids.next());
 
         // Should be nothing in the unsolicited responses channel
-        assert!(recv.try_recv().is_err());
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 773:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -793,7 +790,7 @@ mod tests {
             }
             what => panic!("Unexpected response in unsolicited responses: {:?}", what),
         }
-        assert!(recv.try_recv().is_err());
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 793:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -809,16 +806,16 @@ mod tests {
...
         // Test VANISHED mixed with FETCH
         let lines = b"* VANISHED 3:5,12\r\n\
                       B202 OK [HIGHESTMODSEQ 20010715194045319] expunged\r\n";
 
-        let deleted = parse_expunge(lines.to_vec(), &mut send).unwrap();
+        let deleted = parse_expunge(lines.to_vec(), &mut queue).unwrap();
 
         // No unsolicited responses, they are aggregated in the returned type
-        assert!(recv.try_recv().is_err());
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 818:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -836,10 +833,10 @@ mod tests {
         // UID assigned to the appended message in the destination mailbox.
         // If the MULTIAPPEND extension is also used, there can be multiple UIDs.
         let lines = b"A003 OK [APPENDUID 38505 3955] APPEND completed\r\n";
-        let (mut send, recv) = mpsc::channel();
-        let resp = parse_append(lines, &mut send).unwrap();
+        let mut queue = VecDeque::new();
+        let resp = parse_append(lines, &mut queue).unwrap();
 
-        assert!(recv.try_recv().is_err());
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 839:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/parse.rs
```diff
@@ -858,10 +855,10 @@ mod tests {
         // UID assigned to the appended message in the destination mailbox.
         // If the MULTIAPPEND extension is also used, there can be multiple UIDs.
         let lines = b"A003 OK [APPENDUID 38505 3955:3957] APPEND completed\r\n";
-        let (mut send, recv) = mpsc::channel();
-        let resp = parse_append(lines, &mut send).unwrap();
+        let mut queue = VecDeque::new();
+        let resp = parse_append(lines, &mut queue).unwrap();
 
-        assert!(recv.try_recv().is_err());
+        assert!(queue.pop_front().is_none());
```
jonhoo, line 861:
```suggestion
        assert_eq!(queue.pop_front(), None);
```

src/client.rs
```diff
@@ -530,14 +528,17 @@ impl<T: Read + Write> Client<T> {
...
+    pub fn all_unsolicited(&mut self) -> impl ExactSizeIterator<Item = UnsolicitedResponse> {
```
jonhoo, line 538: I wonder, should this be called `take_all_unsolicited`?

jonhoo (commented):
Excellent, thank you! A few instances of one change I'd like to see (I'll just commit those), and otherwise two questions.

src/extensions/list_status.rs
soywod: Because I get the following error:

```
rust-analyzer [E0616]: field `unsolicited_responses` of `Session` is private
```

which I do not understand because the fn is attached to `impl Session`. The work-around I found is to make the `unsolicited_responses` field `pub(crate)`, is that OK?

src/extensions/metadata.rs
soywod: Same as https://github.com/jonhoo/rust-imap/pull/260#discussion_r1906762635.

src/extensions/list_status.rs
jonhoo: Oh, yeah, totally fine to make it `pub(crate)` :+1:

jonhoo: I _think_ you get it because privacy is about where the code that tries to access the field is located, not about the type it's on. For example, I can do:
```rust
trait MyTrait { fn foo(&self); }
impl MyTrait for std::collections::HashSet<bool> {
  fn foo(&self) {
    // but I don't get to access private fields of `self` here!
  }
}

jonhoo (approved):
Hey, we managed to land it before two years passed :joy:
Mostly my fault — thanks for sticking with it!

PR #262, opened 2023-05-24
Create a testing module that includes helper methods to build mock responses for testing

### LINKED ISSUE #261: How to create mock response objects for testing
Previous to #251 I was able to call parse to create mock response objects to use in my application testing. With the latest update I am no longer able to do that.  

So the question is how can we provide that functionality to allow a user of the crate to mock and write tests for their code that calls the imap crate without having an imap server setup.

In my case I am using the mock_double crate to do things like this.
```rust
            imap.expect_get_acl()
                .withf(|email: &String| email == "user/bob@example.org")
                .returning(|_email| {
                    let (mut tx, _rx) = mpsc::channel();
                    let response = b"* ACL INBOX bob@example.com lrwpx\r\n".to_vec();

                    AclResponse::parse(response, &mut tx)
                });
```

I know we had discussed on the ACL/quota PRs about how to create a better way of exposing this.  But now that this has been removed we should discuss this again to figure out a clean solution to allow building response objects for the purpose of mocking and testing.

urkle:
Allows building response objects for testing applications that use this crate.

Implements a possibility to resolve #261 

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/262)

### DIFF
--- a/src/testing.rs
+++ b/src/testing.rs
@@ -1,0 +1,210 @@
+//! Enable the test_helpers feature to expose helper methods to build
+//! mock response structures for testing your code that uses the imap crate
+//!
+//! To use add a dev-dependency on the imap extension adding the feature "test_helpers"
+//! e.g.
+//!
+//! ```toml
+//! [dependencies]
+//! imap = { version = "3.0" }
+//!
+//! [dev-dependencies]
+//! # mirror the same configuration your dependencies and add test_helpers
+//! imap = { version = "3.0", features = ["test_helpers"] }
+//! ```
+//!
+
+/// Methods to build a [`Capabilities`] response object
+pub mod capabilities {
+    use crate::types::Capabilities;
+    use std::sync::mpsc;
+
+    /// Builds an [`Capabilities`] based on the provided input
+    ///
+    /// Example input.
+    ///
+    /// ```
+    /// let input = "* CAPABILITY IMAP4rev1 STARTTLS AUTH=GSSAPI LOGINDISABLED\r\n";
+    /// let response = imap::testing::capabilities::parse(input);
+    /// ```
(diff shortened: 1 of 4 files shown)


src/testing.rs
```diff
@@ -0,0 +1,148 @@
...
+use std::sync::mpsc;
+
+/// Builds an [`Capabilities`] based on the provided input
+///
+/// Example input.
+///
+/// let input = b"\
+/// * CAPABILITY IMAP4rev1 STARTTLS AUTH=GSSAPI LOGINDISABLED\r\n\
+/// ";
+pub fn build_capabilities(input: Vec<u8>) -> Result<Capabilities> {
```
jonhoo, line 33: I think it'd be nice if these were in submodules instead so that we have a place to add builders later. I'm thinking:
```rust
mod capabilities {
  pub fn parse(input: Vec<u8>) -> Result<Capabilities>;
}
```
I also think we should use `impl Into<Vec<u8>>` instead so that byte strings will just work and not require `.into()`.

And finally, since this is just for testing (and to help ensure that's the case), what do you think of always `.unwrap`ing inside of these rather than exposing the result. Makes it less attractive to use them for anything else.

jonhoo (changes requested):
Thanks for picking this up! Left some notes.

src/testing.rs
urkle: sounds good on all points.   I'll make the adjustments.

urkle:
@jonhoo all comments addressed.  It looks much nicer with the modules.  Also the "parse" methods now have more complete example comments to test that things actually compile and run correctly.

jonhoo (approved):
Looks great now, thanks!

PR #272, opened 2023-09-27
Implement From<String> for QuotaResourceName

urkle:
In using the quota feature in my application (finally) I noticed that I was running into life-time specifier issues trying to convert my input String into a QuotaResourceLimit through the Into<QuotaResourceName> argument.

I had to resort to re-implementing my own match line to convert into an Owned QuotaResourceName.

This PR adds an implementation of From<String> for QuotaResourceName<'_> so strings can be utilized as well as &str.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/272)

### DIFF
--- a/src/types/quota.rs
+++ b/src/types/quota.rs
@@ -57,6 +57,16 @@ impl<'a> From<&'a str> for QuotaResourceName<'a> {
     }
 }
 
+impl From<String> for QuotaResourceName<'_> {
+    fn from(input: String) -> Self {
+        match input.as_str() {
+            "STORAGE" => QuotaResourceName::Storage,
+            "MESSAGE" => QuotaResourceName::Message,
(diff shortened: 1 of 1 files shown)


jonhoo:
Hmm, this doesn't seem quite right — it means there's no way to _avoid_ turning it into an owned string. Wouldn't it be better to `impl From<String> for QuotaResourceName<'static>` in addition instead?

urkle:
> Hmm, this doesn't seem quite right — it means there's no way to _avoid_ turning it into an owned string. Wouldn't it be better to `impl From<String> for QuotaResourceName<'static>` in addition instead?

Yup.. that makes better sense.  I knew this was "fishy", just wasn't sure how to easily fix it.

urkle:
@jonhoo updated

jonhoo (approved):
Perfect, thank you!

PR #281, opened 2024-02-05
Update greenmail docker image version in README.

ath:
Hint to use the newer (stable) `1.6.15` version.
All tests still pass.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-imap/281)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
         run: cargo test --locked --all-targets
     services:
       greenmail:
-        image: greenmail/standalone:1.6.8
+        image: greenmail/standalone:1.6.15
         ports:
           - 3025:3025
           - 3110:3110
--- a/README.md
+++ b/README.md
@@ -91,8 +91,8 @@ running](http://www.icegreen.com/greenmail/#deploy_docker_standalone). The
 easiest way to do that is with Docker:
 
 ```console
-$ docker pull greenmail/standalone:1.6.8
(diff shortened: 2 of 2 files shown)


jonhoo:
Good idea! Would you mind also updating the images used in CI (under `.github/workflows/`)?

ath:
> Good idea! Would you mind also updating the images used in CI (under `.github/workflows/`)?

Absolutely, I’ve added another commit. The workflow ran okay and passed all tests, with the exception of coverage (codecov.io): https://github.com/ath/rust-imap/actions

jonhoo (approved):
Thanks!

jonhoo:
Hmm, looks like the greenmail CI job is failing?

ath:
> Hmm, looks like the greenmail CI job is failing?

Yes, that is a bit surprising.
This is my run on GH Actions: https://github.com/ath/rust-imap/actions/runs/7863986957
And here our PR run: https://github.com/jonhoo/rust-imap/actions/runs/7863987060

I understodd that my coverage test on codecov failed, as I have no account there and didn’t provide an API key. You otoh likely configured this, so the PR run was okay with it.

But the greenmail tests ran without any troubles, a few minutes before it was started here in this repo.

ath:
I looked at the two failed tests. In the first case all tests passed. Only at the end the output says that the "operation was cancelled".

The second step failed though in the test `tls_force` with a network problem, the connection was lost. Given that it ran on Github in my repo, and given that it runs without problems on my local machine I assume that this might be connected with GH Actions. Can you trigger the build & tests again and we see if this runs through?

Locally I did this:
`docker run -d -p 3025:3025 -p 3110:3110 -p 3143:3143 -p 3465:3465 -p 3993:3993 -p 3995:3995 -e GREENMAIL_OPTS="-Dgreenmail.setup.test.all -Dgreenmail.hostname=0.0.0.0 -Dgreenmail.auth.disabled -Dgreenmail.verbose" greenmail/standalone:1.6.15`

and then
`cargo test --locked --all-targets`

Works perfectly.

jonhoo:
Interesting — looks like they both passed after I retried them, so we're all good! Thanks :tada:

PR #303, opened 2025-01-28
Keep track of absolute IDLE time for timeout

### LINKED ISSUE #300: Wrongful IDLE timeout reset on response
I'm connecting to an IMAP server that boots IDLE connections after 30 minutes. I'm using imap-3.0.0-alpha.14, pretty much straight forward
```rust
    let imap = client.login(username, password)?;
    let mbox = imap.select(inbox)?;
    match imap.idle().wait_while(|response| { ... });
```
I works well, apart from getting booted periodically.

I believe the issue may be, that the time is not counted from the start of the IDLE loop, but for every read, since the timeout is implemented by calling [`TcpStream::set_read_timeout()`](https://github.com/jonhoo/rust-imap/blob/6fe22ed11a1ccffe1799a73f48e589366d8d100f/src/extensions/idle.rs#L257) [once before the loop](https://github.com/jonhoo/rust-imap/blob/6fe22ed11a1ccffe1799a73f48e589366d8d100f/src/extensions/idle.rs#L236). So when a response arrives on the client, the timeout is effectively reset.

z33ky:
Fixes #300.

### DIFF
--- a/src/extensions/idle.rs
+++ b/src/extensions/idle.rs
@@ -13,7 +13,7 @@ use rustls_connector::TlsStream as RustlsStream;
 use std::io::{self, Read, Write};
 use std::net::TcpStream;
 use std::ops::DerefMut;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 /// `Handle` allows a client to block waiting for changes to the remote mailbox.
 ///
@@ -58,7 +58,7 @@ pub struct Handle<'a, T: Read + Write> {
     session: &'a mut Session<T>,
     timeout: Duration,
     keepalive: bool,
-    done: bool,
+    last_idle: Option<Instant>,
 }
 
 /// The result of a wait on a [`Handle`]
@@ -91,11 +91,14 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
             session,
             timeout: Duration::from_secs(29 * 60),
             keepalive: true,
-            done: false,
+            last_idle: None,
         }
     }
 
-    fn init(&mut self) -> Result<()> {
+    fn init(&mut self) -> Result<Instant> {
+        let last_idle = Instant::now();
+        self.last_idle = Some(last_idle);
+
         // https://tools.ietf.org/html/rfc2177
         //
         // The IDLE command takes no arguments.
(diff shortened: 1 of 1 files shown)


src/extensions/idle.rs
```diff
@@ -108,39 +111,84 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
-            match self.session.readline(&mut v) {
+            match {
+                // The server MAY consider a client inactive if it has an IDLE command
+                // running, and if such a server has an inactivity timeout it MAY log
+                // the client off implicitly at the end of its timeout period.  Because
+                // of that, clients using IDLE are advised to terminate the IDLE and
+                // re-issue it at least every 29 minutes to avoid being logged off.
+                // This still allows a client to receive immediate mailbox updates even
+                // though it need only "poll" at half hour intervals.
+                let time_since_idle = Instant::now() - self.last_idle.map(Ok).unwrap_or_else(|| self.init())?;
```
jonhoo, line 170:
```suggestion
                let time_since_idle = self.last_idle.map(|t| Ok(t.elapsed())).unwrap_or_else(|| self.init())?;
```

src/extensions/idle.rs
```diff
@@ -108,39 +111,84 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+                if time_since_idle >= self.timeout {
+                    Err(Error::Io(io::ErrorKind::TimedOut.into()))
+                } else {
+                    self.session
+                        .stream
+                        .get_mut()
+                        .set_read_timeout(Some(self.timeout - time_since_idle))
+                        // set_read_timeout() can fail if the argument is Some(0), which can never
+                        // be the case here.
+                        .unwrap();
```
jonhoo, line 180:
```suggestion
                        .expect("cannot be Some(0) since time is monotonically increasing");
```

src/extensions/idle.rs
jonhoo, line 170: I _think_ this means that if `init` fails, we may fail to reset the read timeout to `None` if we've already gone around the loop at least one? Because the `?` here will return from the _function_, not into the match.

src/extensions/idle.rs
```diff
@@ -108,39 +111,84 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+                }
+            } {
                 Err(Error::Io(ref e))
                     if e.kind() == io::ErrorKind::TimedOut
                         || e.kind() == io::ErrorKind::WouldBlock =>
                 {
+                    if self.keepalive {
+                        self.terminate()?;
+                        continue;
+                    }
```
jonhoo, line 191: Nice — this makes the flow much better.

jonhoo (changes requested):
Thank you for this! Two nits, and then one thing that I _think_ is wrong (but also easily fixed).

src/extensions/idle.rs
z33ky: True. Also from `self.terminate()?` a couple of lines below. I'll change both to break on `Err` instead of using `?`.

src/extensions/idle.rs
qsdgy: Will it panic here on the first loop iteration when the `timeout` is set to `Duration::ZERO`? 
If so, would it be better to return an error instead?

src/extensions/idle.rs
```diff
@@ -183,60 +243,11 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
-    /// timeout duration set by [`Handle::timeout`]. If you do not want this behaviour, call
-    /// this function and the connection will simply IDLE until `wait_while` returns or
-    /// the timeout expires.
-    pub fn keepalive(&mut self, keepalive: bool) -> &mut Self {
-        self.keepalive = keepalive;
-        self
-    }
+        // set_read_timeout() can fail if the argument is Some(0), which can never be the
+        // case here.
+        self.session.stream.get_mut().set_read_timeout(None).unwrap();
```
qsdgy, line 245: Is it possible to restore the original [`read_timeout`](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.read_timeout) ? When building a `Client` from custom `stream`, its `timeout` may have been customized.

z33ky: I don't see a reason why it wouldn't be possible.
The previous code overwrote the timeout without attempting to restore the original value as well, so that is a separate change that I'm not planning to implement.

src/extensions/idle.rs
z33ky: An error is returned in that case
https://github.com/jonhoo/rust-imap/pull/303/files#diff-a11b127eccbbebaad79d0cfb2418302d979952456e70664a86161fc63b622de4R174

qsdgy: It seems that [`.unwrap_or_else(|| self.init().map(|()| Duration::ZERO))`](https://github.com/jonhoo/rust-imap/pull/303/files#diff-a11b127eccbbebaad79d0cfb2418302d979952456e70664a86161fc63b622de4R180) covers that `Err(Error::Io(io::ErrorKind::TimedOut.into()))` and leads to [`.set_read_timeout(Some(self.timeout - Duration::ZERO))`](https://github.com/jonhoo/rust-imap/pull/303/files#diff-a11b127eccbbebaad79d0cfb2418302d979952456e70664a86161fc63b622de4R186)

z33ky: Right, on the first iteration `last_idle` is `None` so the check is skipped. I'm moving it to fix this issue.

src/extensions/idle.rs
```diff
@@ -108,39 +110,94 @@ impl<'a, T: Read + Write + 'a> Handle<'a, T> {
...
+                            return Err(Error::Io(io::ErrorKind::TimedOut.into()));
+                        }
+                        self.session
+                            .stream
+                            .get_mut()
+                            .set_read_timeout(Some(self.timeout - time_since_idle))
+                            .expect("cannot be Some(0) since that is guarded against");
+                        self.session.readline(&mut v)
+                    })
+            } {
```
qsdgy, line 187: How about changing to this flavor? The nesting is too deep.
```
        let result = loop {
            // The server MAY consider a client inactive if it has an IDLE command
            // running, and if such a server has an inactivity timeout it MAY log
            // the client off implicitly at the end of its timeout period.  Because
            // of that, clients using IDLE are advised to terminate the IDLE and
            // re-issue it at least every 29 minutes to avoid being logged off.
            // This still allows a client to receive immediate mailbox updates even
            // though it need only "poll" at half hour intervals.

            // If there's no self.last_idle, initialize the connection.
            if self.last_idle.is_none() {
                self.init()?;
            }

            let time_since_idle = last_idle.elapsed();
            if self.timeout <= time_since_idle {
                return Err(Error::Io(io::ErrorKind::TimedOut.into()));
            }

            self.session
                .stream
                .get_mut()
                .set_read_timeout(Some(self.timeout - time_since_idle))
                .expect("cannot be Some(0) since that is guarded against");

            // If no error occurred, read from the stream.
            match self.session.readline(&mut v) {
```

